PR #18, opened 2016-12-13
Added an example for a line graph

yonghah:
In some traditions of street network research, street becomes a node and the edges are connected when these streets touch each other. In OSMnx, basically, the intersections of streets become nodes and streets become edges. Therefore, in order to build a network according to the previous tradition, you have to modify the network that OSMnx generates properly. Fortunately, NetworkX has line_graph() function that does exactly that operation. This example demonstrates the process of transforming a city's street network into a line graph using OSMnx and calculating and coloring the closeness centrality for this line graph.


gboeing:
Thanks for this. A demo of line graph based analysis of centralities is a great addition to OSMnx's network analysis capabilities.

PR #26, opened 2017-01-16
Update save_load.py: Add Feature: read regular  shp and convert to routable graph (osmnx object)

philippschw:
Hi,
my initial data is not in OSM format but in conventional shapefile or geojson format. Therefore, I implemented a way to transform an arbitrary file that consists of a linestring layer into a networkx graph, where then I can apply all osmnx operations. I am not an experienced developer but it works well for me. I hope you would be interested to make it part of your package (though the package name suggest something else..:)).

As minimal condition, the file needs to include at least: 1. geometry linestring and 2. the length of the line (maybe this condition could be checked and then  computed on the spot if needed) .
Because an osmid is naturally not present I map to each node in the graph a serial index.

### DIFF
--- a/osmnx/save_load.py
+++ b/osmnx/save_load.py
@@ -364,3 +364,86 @@ def save_gdf_shapefile(gdf, filename=None, folder=None):
     if not hasattr(gdf, 'name'):
         gdf.name = 'unnamed'
     log('Saved the GeoDataFrame "{}" as shapefile "{}"'.format(gdf.name, folder_path))        
+    
+
+def convert_shp2graph(p, make_G_bidi = True, name='unamed'):
+    """
+    Converts shapefile to routable networkx graph.
+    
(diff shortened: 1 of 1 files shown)


gboeing:
I like the idea of this PR in general, though like you mentioned it is a bit outside the scope of the OSMnx project (as it doesn't use OSM). However, it's not really workable in its current format. It looks like it only loads a single shapefile of edges and then constructs nodes from their endpoints, but OSMnx expects both nodes and edges with consistent sets of attributes. As users presumably have transportation network shapefiles in innumerable formats (e.g., separate shapefiles for nodes and edges? consistent attribute names?), it is hard to assume a single format and parse accordingly without making this a much larger project. That said, I like the general idea of loading shapefiles into OSMnx, but it would probably need another PR to address a clearer and common format of shapefiles.

PR #63, opened 2017-06-21
Option for download of other infrastructures

DavidBreuer:
A new keyword parameter infrastructure='way["highway"]' was added to enable download and analysis of other infrastructures:

    infrastructure : string
        download infrastructure of given type (default is streets (ie, 'way["highway"]') but other 
        infrastructures may be selected like power grids (ie, 'way["power"~"line"]'))

A neutral filter network_type='none' was added so that other infrastructures are not removed due to missing "highway" tags.

Thanks and best,
David

### DIFF
--- a/osmnx/__init__.py
+++ b/osmnx/__init__.py
@@ -15,4 +15,4 @@ from .simplify import *
 from .stats import *
 from .utils import *
 
-__version__ = '0.5'
+__version__ = '0.5.1'
--- a/osmnx/buildings.py
+++ b/osmnx/buildings.py
@@ -131,30 +131,32 @@ def create_buildings_gdf(polygon=None, north=None, south=None, east=None, west=N
     GeoDataFrame
     """
 
-    results = osm_bldg_download(polygon, north, south, east, west)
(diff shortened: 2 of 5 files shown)


gboeing:
Hi @DavidBreuer, thanks much for this PR! Looks like the Travis CI tests are currently failing -- can you give it a look and see what can be done to get this commit to pass the tests?

gboeing:
@DavidBreuer nevermind, this seems to be a GDAL problem: https://github.com/conda-forge/gdal-feedstock/issues/151#issuecomment-310257157

DavidBreuer:
Hi Geoff, ok, thanks!



Am 22.06.2017 um 04:22 schrieb Geoff Boeing:

ocefpaf:
@DavidBreuer PR #64 helps to understand the issue, but my guess is that it is already resolved. Can you try to restart the CIs here so we can check?

gboeing:
@ocefpaf re-ran the CIs, all working now, thanks!

ocefpaf:
Glad it is resolved. Deployment of new versions in conda-forge is always like

PR #64, opened 2017-06-22
help debug travis-ci

ocefpaf:
I will comment the changes in the diff.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,14 +12,15 @@ before_install:
   - bash miniconda.sh -b -p $HOME/miniconda
   - export PATH="$HOME/miniconda/bin:$PATH"
   - hash -r
-  - conda config --set always_yes yes
-  - conda config --add channels conda-forge
-  - conda update --all
-  - conda info -a
-  
-install:
-  - conda create -n TESTENV python=$TRAVIS_PYTHON_VERSION pytest coverage coveralls requests numpy matplotlib pandas networkx Shapely Rtree descartes geopandas folium
+  - conda config --set always_yes yes --set show_channel_urls true
+  - conda update conda
+  - conda config --add channels conda-forge --force
+  - conda create --quiet --name TESTENV python=$TRAVIS_PYTHON_VERSION --file requirements.txt --file requirements-dev.txt
   - source activate TESTENV
+  - conda info --all
+  - conda list
+
+install:
   - pip install .
   - pip show osmnx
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -12,14 +12,15 @@ before_install:
...
   - export PATH="$HOME/miniconda/bin:$PATH"
   - hash -r
-  - conda config --set always_yes yes
-  - conda config --add channels conda-forge
-  - conda update --all
-  - conda info -a
-  
-install:
-  - conda create -n TESTENV python=$TRAVIS_PYTHON_VERSION pytest coverage coveralls requests numpy matplotlib pandas networkx Shapely Rtree descartes geopandas folium
+  - conda config --set always_yes yes --set show_channel_urls true
```
ocefpaf: With this we can see the origin of the package in the with `conda list`.

.travis.yml
```diff
@@ -12,14 +12,15 @@ before_install:
...
+  - conda update conda
```
ocefpaf: Update only `conda` instead of everything. The packages will be up-to-date when we create the env.

.travis.yml
```diff
@@ -12,14 +12,15 @@ before_install:
...
+  - conda config --add channels conda-forge --force
```
ocefpaf: Adding the channel with the `--force` option to ensure it will be added.

.travis.yml
```diff
@@ -12,14 +12,15 @@ before_install:
...
+  - conda create --quiet --name TESTENV python=$TRAVIS_PYTHON_VERSION --file requirements.txt --file requirements-dev.txt
```
ocefpaf: This change allow us to keep the dependencies in one place and use it both on Travis-CI and `setup.py`.
Not sure if this changed is desired, I can revert if not.

.travis.yml
```diff
@@ -12,14 +12,15 @@ before_install:
...
   - source activate TESTENV
+  - conda info --all
+  - conda list
```
ocefpaf: This helps to debug the env packages.

.travis.yml
```diff
@@ -12,14 +12,15 @@ before_install:
...
+
+install:
```
ocefpaf: Note that I moved all the parts regarding the env creation to `before_install` and execute the `conda` debug commands like `list` and `info` inside the env.

PR #81, opened 2017-08-15
code clean up

gboeing:


osmnx/buildings.py
```diff
@@ -165,7 +165,7 @@ def create_buildings_gdf(polygon=None, north=None, south=None, east=None,
                 nodes = result['nodes']
                 try:
                     polygon = Polygon([(vertices[node]['lon'], vertices[node]['lat']) for node in nodes])
-                except:
+                except Exception as e:
```
kuanb: Hope not too much backseat driving but I think the two possible errors here would be `ValueError` and `AssertionError`, right? Any others. Just a thought if a more specific error catching was desired - explicitly handle those errors related to polygon parsing and raise any others (example would be if there was a type in line 167 that would be silent).

kuanb: Also, 94+% coverage is impressive :D

gboeing: Yeah. I'm thinking about adding explicit errors, but it'll be a chunk of work. For now I'm just considering going through to at least define a general Exception e. If you are looking for more specific error catching downstream, PRs always welcome!

kuanb: Fair enough! :)

PR #84, opened 2017-08-22
Migrating to networkx 2.0

gboeing:
Migrating OSMnx from networkx 1.x to networkx 2.0's new API.

### DIFF
--- a/osmnx/plot.py
+++ b/osmnx/plot.py
@@ -346,8 +346,8 @@ def plot_graph(G, bbox=None, fig_height=6, fig_width=None, margin=0.02,
     """
 
     log('Begin plotting the graph...')
-    node_Xs = [float(data['x']) for _, data in G.nodes(data=True)]
-    node_Ys = [float(data['y']) for _, data in G.nodes(data=True)]
+    node_Xs = [float(x) for _, x in G.nodes(data='x')]
+    node_Ys = [float(y) for _, y in G.nodes(data='y')]
 
     # get north, south, east, west values either from bbox parameter or from the
     # spatial extent of the edges' geometries
@@ -554,7 +554,7 @@ def plot_graph_route(G, route, bbox=None, fig_height=6, fig_width=None,
     lines = []
     for u, v in edge_nodes:
         # if there are parallel edges, select the shortest in length
-        data = min([data for data in G.edges[u, v].values()], key=lambda x: x['length'])
+        data = min(G.get_edge_data(u, v).values(), key=lambda x: x['length'])
 
         # if it has a geometry attribute (ie, a list of line segments)
         if 'geometry' in data and use_geom:
@@ -868,9 +868,11 @@ def plot_figure_ground(G=None, address=None, point=None, dist=805,
         # for each node, get a nodesize according to the narrowest incident edge
         node_widths = {}
         for node in G_undir.nodes():
-            edge_types = [list(edge.values())[0]['highway'] for edge in G_undir.edge[node].values()]
+            # first, identify all the highway types of this node's incident edges
+            incident_edges_data = [G_undir.get_edge_data(node, neighbor) for neighbor in G_undir.neighbors(node)]
+            edge_types = [data[0]['highway'] for data in incident_edges_data]
             if len(edge_types) < 1:
-                # if node has no edges, make size zero
+                # if node has no incident edges, make size zero
                 node_widths[node] = 0
             else:
                 # flatten the list of edge types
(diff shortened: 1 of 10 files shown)


gboeing:
More info:

- networkx 2.0 migration guide: https://networkx.readthedocs.io/en/latest/release/migration_guide_from_1.x_to_2.0.html
- MultiDiGraph class documentation: https://networkx.readthedocs.io/en/latest/reference/classes/multidigraph.html

osmnx/plot.py
```diff
@@ -554,7 +554,7 @@ def plot_graph_route(G, route, bbox=None, fig_height=6, fig_width=None,
     lines = []
     for u, v in edge_nodes:
         # if there are parallel edges, select the shortest in length
-        data = min([data for data in G.edge[u][v].values()], key=lambda x: x['length'])
+        data = min(G.get_edge_data(u, v).values(), key=lambda x: x['length'])
```
dschult: Could be ```G[u][v].values()``` or ```G.adj[u][v].values()``` or ```G.get_edge_data(u, v).values()```. They are equivalent.

osmnx/plot.py
```diff
@@ -868,9 +868,11 @@ def plot_figure_ground(G=None, address=None, point=None, dist=805,
         # for each node, get a nodesize according to the narrowest incident edge
         node_widths = {}
         for node in G_undir.nodes():
-            edge_types = [list(edge.values())[0]['highway'] for edge in G_undir.edge[node].values()]
+            # first, identify all the highway types of this node's incident edges
+            incident_edges_data = [G_undir.get_edge_data(node, neighbor) for neighbor in G_undir.neighbors(node)]
+            edge_types = [data[0]['highway'] for data in incident_edges_data]
```
dschult: You could stick with what you had replacing ```G_undir.edge[node]``` with ```G_undir.adj[node]``` or ```G_undir[node]```

osmnx/save_load.py
```diff
@@ -373,15 +373,16 @@ def get_undirected(G):
         # if we haven't already flagged this edge as a duplicate
         if not (u, v, key) in duplicate_edges:
 
-            # look at this first edge's parallel edges one at a time
-            for key_other in H.edge[u][v]:
+            # look at every other edge between u and v, one at a time
+            for key_other in dict(H[u][v]).keys():
```
dschult: The code ```H[u][v]``` creates a dict.  So you don't need the dict/keys parts.  Replace ```G.edge[u][v]``` with either ```G[u][v]``` or ```G.adj[u][v]```   (they are equivalent)

osmnx/simplify.py
```diff
@@ -71,13 +71,13 @@ def is_endpoint(G, node, strict=True):
 
         # add all the edge OSM IDs for incoming edges
         for u in G.predecessors(node):
-            for key in G.edge[u][node]:
-                osmids.append(G.edge[u][node][key]['osmid'])
+            for key in dict(G[u][node]).keys():
```
dschult: Same as above: could replace ```G.edge``` with ```G.adj``` everywhere in the old code. You can also use ```G[u][node]``` in place of ```G.edge[u][node]``` if you prefer.

osmnx/simplify.py
```diff
@@ -243,14 +243,13 @@ def simplify_graph(G_, strict=True):
         for u, v in zip(path[:-1], path[1:]):
 
             # there shouldn't be multiple edges between interstitial nodes
-            edges = G.edge[u][v]
-            if not len(edges) == 1:
+            if not G.number_of_edges(u=u, v=v) == 1:
```
dschult: Could be:   ```if len(G[u][v]) != 1:```   but maybe ```G.number_of_edges``` is clearer.

osmnx/utils.py
```diff
@@ -509,7 +509,7 @@ def get_route_edge_attributes(G, route, attribute, minimize_key='length'):
     for u, v in zip(route[:-1], route[1:]):
         # if there are parallel edges between two nodes, select the one with the
         # lowest value of minimize_key
-        data = min([data for data in G.edge[u][v].values()], key=lambda x: x[minimize_key])
+        data = min(G.get_edge_data(u, v).values(), key=lambda x: x[minimize_key])
```
dschult: same options as above: ```G.edge``` could be replaced by ```G```, ```G.adj``` or by using ```G.get_edge_data```. They are equivalent.

osmnx/simplify.py
gboeing: @dschult thanks Dan, very helpful.

PR #85, opened 2017-08-26
Possible bug in simplify.py:build_path

debsankha:
Currently simplify.build_path function looks for all successors of a node and tries to build a path ending in one of teh endpoint nodes specified. However, if it finds a path, it does not stop, but explores all remaining successors. I think this is a bug.

### DIFF
--- a/osmnx/simplify.py
+++ b/osmnx/simplify.py
@@ -116,6 +116,7 @@ def build_path(G, node, endpoints, path):
                 # if this successor is not an endpoint, recursively call
                 # build_path until you find an endpoint
                 path = build_path(G, successor, endpoints, path)
+                break
             else:
                 # if this successor is an endpoint, we've completed the path,
                 # so return it


gboeing:
I'm not sure this is a bug. For example, if we have an endpoint node that represents a 4-way intersection of bidirectional streets, this node will have 4 successors. If you break the for-loop after building a path through the first successor, you will be ignoring the other 3 successors and never building a path through them. To correctly simplify the graph, we want to build these paths to simplify from each endpoint node through *each* of their successors.

debsankha:
Thanks for the reply. I will illustrate what I mean by a MWE. 

```python
>>> H = nx.DiGraph()
>>> H.add_path([1,2,3,4])
>>> H.add_path([2,5,6])
>>> ox.build_path(H, 2, {1,4,6}, [1,2])
[1, 2, 3, 4, 5, 6]
```
Shouldn't the expected result be `[1,2,3,4]` or `[1,2,5,6]`?
Apologies if I have misunderstood the algorithm.

gboeing:
No. The first reason is because 1, 2, 4, and 6 are endpoints (aka, intersections and dead-ends), so you wouldn't have a path running through 2. The second reason is because the arguments you're passing `build_path` aren't quite right. Rather, the algorithm works like this:

```python
>>> ox.build_path(H, 3, {1,2,4,6}, [2,3])
[2, 3, 4]
```

For example:
```python
>>> ox.get_paths_to_simplify(H)
[[2, 3, 4], [2, 5, 6]]
```

Or in the full context, adapted from how the algorithm appears in the module:
```python
>>> H = nx.MultiDiGraph()
>>> H.add_path([1,2,3,4])
>>> H.add_path([2,5,6])
>>> endpoints = set([node for node in H.nodes() if ox.is_endpoint(H, node)])
>>> for node in endpoints:
>>>     for successor in H.successors(node):
>>>         if successor not in endpoints:
>>>             path = ox.build_path(H, successor, endpoints, path=[node, successor])
>>>             print(path)
[2, 3, 4]
[2, 5, 6]
```

This is the correct outcome. We do not need to simplify the (1,2) path, but we do need to simplify the (2,3,4) path and the (2,5,6) path to remove non-intersection pseudonodes.

gboeing:
@debsankha let me know if this is clear. If the functions' docstrings are confusing, it might be helpful to change this PR to add a bit more detail to the function descriptions and parameters documentation.

debsankha:
@gboeing : Thanks for your explanations. I think the docstrings are fine. 
Now I understand that the code is correct.

However, regarding my original question, I still have a doubt. 
I cannot see any scenario where after line 118 the build_path function will continue onto line 111. I think so, because if it did, that would mean `node` has more than one successor, and therefore `node` is an `endpoint`. Therefore no path would be built through it anyways. 

I would be glad to have your opinion on this.

gboeing:
@debsankha yes possibly. The original logic behind this function was to make it generalizable for other uses, but I think over time it ended up getting specialized to fit into OSMnx's precise workflow. Given  your point that `node` should only have one successor (because it is a directed graph, and it has already been identified as *not* being an intersection or dead-end node when `build_path()` was called by `get_paths_to_simplify()`), perhaps a good way to handle this would be to elminate the for loop on line 110, and replace it with an `assert len(list(G.successors(node))) == 1` to make sure it only has a single successor?

Issue #93, opened 2017-10-06
Graph from file?


migurski:
I’m looking at OSMnx and needing to create a graph from a local OSM file, such as the ones generated by [Metro Extracts](https://mapzen.com/metro-extracts). I would expect to see this in the group of `graph_from_*` functions (e.g. graph_from_address, graph_from_bbox, etc.). Does this exist someplace? If it doesn't, would it be a welcome addition to OSMnx?

kuanb:
I don't believe there exists, currently, the ability to do a graph from a local file.

All the `graph_from_*` (I believe) wrap just over `graph_from_bbox`, which feeds `osm_net_download` output into `create_graph`.

If it (a `from_local` were to be generated, I think there could be the option to read from a file instead of run `osm_net_download` and pass that to `create_graph`. Would just need to resolve the fact that the response is current a JSON and the extracts would be XML, PBF, etc...

migurski:
Thanks Kuan! The way I'm thinking of using this data, I need to have a known version of OSM data to coordinate between a few different pieces of software. I’ll root around in the code to see if any obvious implementations jump out at me.

migurski:
Yeah! What’s nice about starting from OSM files is that you can [run Osmosis on them](http://wiki.openstreetmap.org/wiki/Osmosis) to do your filtering. I don't specifically have an issue with the OO vs. functions thing, just looking for another way to get vetted data in.

gboeing:
@migurski @kuanb correct, this functionality does not currently exist. However it would be a very welcome addition. I know there has been interest in this previously from other quarters... I just haven't had the time or personal need to develop it myself. But a graph_from_osm_file() function would be really useful if you have any interest in looking into it. It wouldn't take much processing to pass the read data off to the create_graph() function.

gboeing:
Resolved by #95

zifeo:
@gboeing Is there any way of saving a graph as osm (or pbf)? Is this reasonable a feature?

gboeing:
@zifeo no, OSMnx saves graphs as GraphML, ESRI shapefile, or regular old node/edge lists. Converting data back to .osm format would be a specialized task and probably best to consult https://wiki.openstreetmap.org/wiki/Converting_map_data_between_formats or similar.

zifeo:
Thanks!

PR #95, opened 2017-10-10
Add graph from file

migurski:
Introduces new `graph_from_file()` function in osmnx.core. Parses OSM XML files and generates Overpass-like JSON data. This is a stub implementation, and I'm looking for feedback on these questions before I continue:

- Is there some deeper underlying object format I should use in preference to Overpass JSON, or does this XML-to-JSON flow seem good?
- Do OSM relations need to be supported?
- What optional parameters need to exist for `graph_from_file()`?

Addresses #93.


gboeing:
1. I think xml to json works well. That will make hand-off to [create_graph](https://github.com/gboeing/osmnx/blob/master/osmnx/core.py#L1262) easy enough.

2. As of now, no. But that's a future enhancement that'd be nice to make library-wide. For now, just adding a stub or comment or something pointing out it's not there would be useful.

3. I think `simplify`, `retain_all`, `network_type`, and `name`. Each is easy to handle in like a line or 2 of code such as [here](). The `network_type` argument is needed by create_graph to create bidirectional edges as appropriate. Presumably, the purpose of this new `graph_from_file` function is to turn the all the way contents of the .osm file into a graph, so I assume the function will not filter edge types based on the `network_type` argument. But I'm open to suggestions.

gboeing:
and, thanks much for digging into this!

migurski:
I think this is done. I’ve added a test for reading data from files, and I believe the unit failure test is something to do with geocoding.

gboeing:
Thanks @migurski! I'll take a look at this soon.

sevamoo:
Thanks a lot for this new function. It works very well.

awa5114:
Does the pbf need to contain any specific information or can it contain all default osm data? For instance I have tried to run the `graph_from_file` function on a pbf generated from GeoFabriks Europe extract. The subextract I am using is for a very small test area with poly as follows:

```
1
0
    5.07448295115847     52.01761020898715
    5.061074139766845     52.020260672951096
    5.05768817426537     52.04045565493337
    5.064169334843836     52.07093448682436
    5.076798284522554     52.07550514373811
    5.100512051914809     52.065312416966435
    5.095128489319279     52.03868762461073
    5.07448295115847     52.01761020898715
END
END
```

The osmconvert command I did on the GeoFabrik extract is as follows:

`osmconvert64 europe.osm.pbf -B=test.poly -o=test.osm`

With no additional filters. I then try to run 

```
import osmnx

G = osmnx.graph_from_file('test.osm')
```

But I get an error

>   File "C:\Users\aboufira\AppData\Local\Continuum\miniconda3\envs\netwerkanalyse\lib\site-packages\osmnx\core.py", line 1911, in graph_from_file
>     retain_all=retain_all, name=name)
>   File "C:\Users\aboufira\AppData\Local\Continuum\miniconda3\envs\netwerkanalyse\lib\site-packages\osmnx\core.py", line 1368, in create_graph
>     G = add_edge_lengths(G)
>   File "C:\Users\aboufira\AppData\Local\Continuum\miniconda3\envs\netwerkanalyse\lib\site-packages\osmnx\core.py", line 1201, in add_edge_lengths
>     coords = np.array([[u, v, k, G.nodes[u]['y'], G.nodes[u]['x'], G.nodes[v]['y'], G.nodes[v]['x']] for u, v, k in G.edges(keys=True)])
>   File "C:\Users\aboufira\AppData\Local\Continuum\miniconda3\envs\netwerkanalyse\lib\site-packages\osmnx\core.py", line 1201, in <listcomp>
>     coords = np.array([[u, v, k, G.nodes[u]['y'], G.nodes[u]['x'], G.nodes[v]['y'], G.nodes[v]['x']] for u, v, k in G.edges(keys=True)])
> KeyError: 'y'


Is it perhaps looping through some nodes/ways it does not actually need? Do I have to do more than simply extract all OSM data for a given extent? Do I have to filter for specific features before I use the `graph_from_file` function? If so it would be nice to include this in the docstring for this function....

celhannah:
Has this function been removed?

gboeing:
Yes, a year or two ago in the package reorganization. See the docs: https://osmnx.readthedocs.io/en/stable/osmnx.html#osmnx.graph.graph_from_xml

Issue #98, opened 2017-10-19
Example in README doesn't appear to work


bkj:
Hi -- I'm trying to run the example given in the README and get the following error:

```
>>> import osmnx as ox
>>> G = ox.graph_from_place('Manhattan, New York, USA', network_type='drive')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/bkj/.anaconda/lib/python2.7/site-packages/osmnx/core.py", line 1800, in graph_from_place
    clean_periphery=clean_periphery, infrastructure=infrastructure)
  File "/home/bkj/.anaconda/lib/python2.7/site-packages/osmnx/core.py", line 1664, in graph_from_polygon
    raise ValueError('Geometry must be a shapely Polygon or MultiPolygon')
ValueError: Geometry must be a shapely Polygon or MultiPolygon
```

Anyone have any ideas?

bkj:
Turning on logging reveals that that query string returns a single point.  Changing to
```
G = ox.graph_from_place('Manhattan Island, New York City, New York, USA', network_type='drive')
```
appears to get the network, but then fails like:
```
Got all network data within polygon from API in 1 request(s) and 0.23 seconds
Creating networkx graph from downloaded OSM data...
Created graph with 20,681 nodes and 29,511 edges in 1.25 seconds
Added edge lengths to graph in 0.76 seconds
Identifying all nodes that lie outside the polygon...
Created r-tree spatial index for 20,681 points in 0.00 seconds
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "osmnx/core.py", line 1800, in graph_from_place
    clean_periphery=clean_periphery, infrastructure=infrastructure)
  File "osmnx/core.py", line 1680, in graph_from_polygon
    G_buffered = truncate_graph_polygon(G_buffered, polygon_buffered, retain_all=True, truncate_by_edge=truncate_by_edge)
  File "osmnx/core.py", line 1125, in truncate_graph_polygon
    points_within_geometry = intersect_index_quadrats(gdf_nodes, polygon, quadrat_width=quadrat_width, min_num=min_num, buffer_amount=buffer_amount)
  File "osmnx/core.py", line 1064, in intersect_index_quadrats
    possible_matches_index = list(sindex.intersection(poly.bounds))
AttributeError: 'NoneType' object has no attribute 'intersection'
```

kuanb:
The error is being thrown on this line:

[...]

PR #99, opened 2017-10-20
Fix manhattan example

bkj:


gboeing:
thanks @bkj !

gboeing:
resolves #98

Issue #97, opened 2017-10-19
passing empty graph to get_nearest_node


Wario7:
Greetings, 

First, I would like to thank you for such wonderful and organized tool. It is very useful for my research.

To reproduce the issue, pass an empty graph to get_nearest_node function

`ox.get_nearest_node(nx.Graph(), point2, return_dist=False)`

it will output an error similar to the below

```
File "/home/ubuntu/.local/lib/python3.5/site-packages/osmnx/utils.py", line 356, in get_nearest_node
    df = pd.DataFrame(coords, columns=['node', 'x', 'y']).set_index('node')
  File "/home/ubuntu/.local/lib/python3.5/site-packages/pandas/core/frame.py", line 306, in __init__
    copy=copy)
  File "/home/ubuntu/.local/lib/python3.5/site-packages/pandas/core/frame.py", line 483, in _init_ndarray
    return create_block_manager_from_blocks([values], [columns, index])
  File "/home/ubuntu/.local/lib/python3.5/site-packages/pandas/core/internals.py", line 4303, in create_block_manager_from_blocks
    construction_error(tot_items, blocks[0].shape[1:], axes, e)
  File "/home/ubuntu/.local/lib/python3.5/site-packages/pandas/core/internals.py", line 4278, in construction_error
    raise ValueError("Empty data passed with indices specified.")

```

I would kindly suggest get_nearest_node to output a friendly error message such as 'Graph passed is empty' in my opinion.

In my case, I am trying to find a nearest neighbor to a given node, and a given location (point2 in example). So

```
import osmnx as ox
ox.config(log_console=True, use_cache=True)
G = ox.graph_from_place('Beijing, china', network_type='drive')

point1 = (40.021889999999999, 116.51600000000001)
source_node = ox.get_nearest_node(G, point1, return_dist=False)

point2 = (40.054729999999999, 116.53664999999999)
destination_node = ox.get_nearest_node(G.neighbors(source_node), point2, return_dist=False)
```
source_node unfortunately had no neighbors, thus, G.neighbors(source_node) produced an empty graph which produced the error.

Thank you for your time.

gboeing:
Sure. I think `get_nearest_node` could either have a quick check that the passed graph is not empty, or we could update the docstring to specify that the graph cannot be empty. I'm open to either. Would you like to contribute a PR as this should be a very simple fix?

Wario7:
created my first PR on github :). I hope I did not miss any step in the contributing guidelines. Thanks!

PR #101, opened 2017-10-20
check if graph G is empty or contains no nodes

Wario7:
test output:

```
python test_osmnx.py

Configured osmnx
test debug
test info
test warning
test error
```


gboeing:
resolves #97 
@Wario7 looks good, thanks!

PR #103, opened 2017-10-21
Add mock data for test requests

### LINKED ISSUE #96: Use mock requests in tests?
I noticed the recent test failures due to imprecise geocoding, which seem to be a product of changing responses from the live Nominatim server. Would you be interested in a change to the test suite that mocks responses from remote servers, so that tests are no longer dependent on 3rd party services or an internet connection? We use these all over the place in `openaddresses/machine`, [to implement responses from observed ArcGIS and other servers](https://github.com/openaddresses/machine/blob/master/openaddr/tests/__init__.py#L88).

migurski:
Stop talking to live Nominatim and Overpass API during testing, and instead use stored mock responses. Closes #96.


gboeing:
Thanks @migurski, looks like `httmock` isn't available on conda... I'll get it up on conda-forge shortly, then we can re-run these tests.

gboeing:
@migurski ok, now only py2.7 tests are failing because of how urllib parse is being imported. To make it python 2/3 compatible we'll need to do something like:

```python
try:
    from urllib.parse import urlparse
except ImportError:
     from urlparse import urlparse
```

migurski:
Passed! The coveralls failures seem unimportant, because this PR only edits test code.

gboeing:
Thanks @migurski! Could you leave a comment here describing how you create the txt and json.gz response files, for documentation in case someone add new tests later?

migurski:
Sure!

For each Nominatim and Overpass API interaction, I queried the API directly using `curl` and wrote the raw JSON response to a file.

PR #104, opened 2017-11-16
modified folium polyline to be compatible with folium-0.5

nsteins:
folium-0.5 does not appear to support the 'latlon=False' flag for Polyline anymore, so it's necessary to flip the coords from lonlat ourselves before building the polyline. This should be backward compatible with previous versions of folium as well.


gboeing:
@nsteins thanks!

PR #105, opened 2017-11-17
Switch from US-only NAD83 datum to global WGS84 datum

migurski:
In [osmnx/projection.py](https://github.com/gboeing/osmnx/blob/cddf8ac83b4e800b4059a1b74c6a4b6b8e135c23/osmnx/projection.py#L108-L115), output CRS always gets `NAD83` as its datum. Since this is only defined for North America, there’s not a complete set of corresponding EPSG codes: http://spatialreference.org/ref/?search=NAD83+UTM

Consider switching to WGS84, which is defined globally: http://spatialreference.org/ref/?search=WGS84+UTM


gboeing:
@migurski thanks Mike!

Issue #107, opened 2017-11-19
Detect API changes on a regular basis


gboeing:
Per conversations with @migurski, Nominatim and Overpass API changes need to be detected more regularly, and as a separate process from code changes. We want to start using cached server responses in the code tests, so we need some other process checking the API now and then to make sure it's returning what we expect.

CI is triggered only when commits and PRs happen, while API changes can happen at any time and affect users regardless of changes to the codebase. We might want to use a daily scheduled task to check for them to let us know that something's up even if no one is actively committing code changes for a lengthy period of time.

migurski:
Here are some [sample calls from the tests](https://github.com/gboeing/osmnx/blob/5008ad755/tests/test_osmnx.py) that a preliminary version should look for:

- `ox.gdf_from_place('Manhattan, New York City, New York, USA', buffer_dist=100)`
- `ox.graph_from_place('Piedmont, California, USA')`
- `ox.graph_from_bbox(37.79, 37.78, -122.41, -122.43, network_type='drive_service')`
- `ox.graph_from_point(ox.bbox_from_point((37.791427, -122.410018), project_utm=True), distance=750, distance_type='bbox', network_type='drive')`
- `ox.graph_from_address(address='350 5th Ave, New York, NY', distance=1000, distance_type='network', network_type='bike')`
- `ox.graph_from_place(['Los Altos, California, USA', {'city':'Los Altos Hills', 'state':'California'}, 'Loyola, California'], network_type='all', clean_periphery=False)`
- `ox.graph_from_place('Piedmont, California, USA', network_type='drive', simplify=False)`
- `ox.graph_from_address('398 N. Sicily Pl., Chandler, Arizona', distance=800, network_type='drive')`
- `ox.buildings_from_place(place='Piedmont, California, USA')`
- `ox.buildings_from_address(address='260 Stockton Street, San Francisco, California, USA', distance=300)`

migurski:
☝️ I’ll use that to make a shortlist of request types and some basic tests.

mmd-osm:
I’m a bit curious if you’ve encountered such an API change in the past. New Overpass releases should always be downward compatible unless there’s some regression which hasn’t been caught by the test suite. Do you plan to report any findings upstream, or what kind of process did you have in mind?

migurski:
[...]

PR #111, opened 2017-11-28
Add scheduled Overpass and Nominatim API tests

migurski:
Adds tests for OpenStreetMap APIs designed to run on a schedule under AWS Lambda.

- [x] Test simple request to Nominatim
- [x] Test [remaining sample calls](https://github.com/gboeing/osmnx/issues/107#issuecomment-345540929)

This is a stub and I’m looking for feedback on the approach. `osm_api_tests.py` is outside the normal project hierarchy, and gets deployed to Lambda as a standalone Python 3 script. It’s possible but slightly annoying to include 3rd party packages on Lambda, so I’m using only standard library functionality. If merged, the AWS deployment details will need to be updated.

Will close https://github.com/gboeing/osmnx/issues/107.


mmd-osm:
If I‘m not mistaken those API requests currently won’t provide a User-Agent in the http header, which would be helpful in case of issues. Also a bit more error handling for Overpass might be needed (think rate limiting).

migurski:
Good point, I'll add a User-Agent header.

gboeing:
@migurski thanks -- looking good. Could you tell me a little more about where and how this runs?

migurski:
Sure! In the bottom of the Travis config file you’ll notice a new `deploy` section. Travis (for `migurski/osmnx` only) is configured with an AWS key/secret for my account to update the named Lambda function with code from the configured Python file. I configured a Cloudfront scheduled task to run the function every 15 minutes, and then attached an alarm to the graph of failures.

If it fails more than a couple times in an hour, the alarm triggers an email to me. So far, there seem to be a few intermittent failures suggesting that I should raise the threshold somewhat. This will require some tuning over time, starting with too many alarms until the threshold feels right.

This setup is effectively free to run, so if you’re interested in using it we can just continue to use my AWS account and add an email to you. Alternatively, I can help you get it set up under your own AWS account.

mmd-osm:
To add a bit of confusion to this topic, you're actually testing the old server (136.243.42.136) only. osmnx  uses `overpass-api.de` in a few places, which would resolve to one of the two available servers via round robin. 

```
> www.overpass-api.de
Server:		127.0.1.1
Address:	127.0.1.1#53

Non-authoritative answer:
Name:	www.overpass-api.de      << old server only
Address: 136.243.42.136             

     
> overpass-api.de
Server:		127.0.1.1
Address:	127.0.1.1#53

Non-authoritative answer:
Name:	overpass-api.de
Address: 136.243.42.136                << old server
Name:	overpass-api.de
Address: 178.63.48.217                 << new server
> 
```

BTW:  testing once per hour should be sufficient as a test interval imho. Servers are not changing that often. In some cases you're already hitting rate limiting, i.e. queries get delayed a bit, or you even get an HTTP 429 response.

migurski:
Thanks @mmd-osm, I updated the PR to reflect your suggested Overpass domain. When running these tests, we can space them out more. It’s just convenient to have them run more frequently while developing.

gboeing:
@migurski looking good. Ready to merge to master? Should Travis config be updated to use master branch?

gboeing:
resolves #107

migurski:
I think it’s ready! It might require some coordination to get the secret updated for the current `gboeing/osmnx` repo instead of the `migurski/osmnx` fork. I will update the branch name here, and then after you merged you may need to make a followup edit for the AWS details. I’ll share those with you offline.

gboeing:
Thanks @migurski !

gboeing:
@migurski yep looks like we need to edit AWS details: https://travis-ci.org/gboeing/osmnx/jobs/318247643#L1116

Let me know how you'd like to proceed.

PR #124, opened 2018-01-25
Add junction roundabout as oneway

RomainQueraud:
Roundabout implies oneway=yes but is not tagged in OSM as it would be redundant.
It still needs to be taken as oneway=yes by not adding edges in both ways between nodes.
[OSM wiki about roundabouts](https://wiki.openstreetmap.org/wiki/Tag:junction%3Droundabout)

### DIFF
--- a/osmnx/core.py
+++ b/osmnx/core.py
@@ -1248,7 +1248,7 @@ def add_paths(G, paths, network_type):
             # add this path (in only one direction) to the graph
             add_path(G, data, one_way=True)
 
-        elif 'junction' in data and data['junction'] == 'roundabout':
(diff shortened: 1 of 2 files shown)


gboeing:
@RomainQueraud great catch, thanks! I think we need to add `and not network_type=='walk'` into the new `elif` condition though, because pedestrians can walk either direction around a roundabout.

RomainQueraud:
Added to a new commit ! I didn't though of this as I'm using osmnx to retrieve driving roads only.

gboeing:
@RomainQueraud thanks!

Issue #127, opened 2018-02-14
G.add_edge keyword args may need to be updated


kuanb:
I was running tests on master locally and found that `test_network_saving_loading` was failing.

It was failing in `gdfs_to_graph` on the following:
```
> G.add_edge(u=row['u'], v=row['v'], key=row['key'], **attrs)
E TypeError: add_edge() missing 2 required positional arguments: 'u_for_edge' and 'v_for_edge'
```

We can view that line in the OSMnx repo, here: https://github.com/gboeing/osmnx/blob/cddf8ac83b4e800b4059a1b74c6a4b6b8e135c23/osmnx/save_load.py#L513

Comparing this to the current docs for NetworkX, I see that the `u` and `v` parameters have different names. I clicked around the NetworkX repo to see how long ago this was set, but didn't dig anything up right away.

You can see both in the docs:
https://networkx.github.io/documentation/stable/reference/classes/generated/networkx.DiGraph.add_edge.html#networkx.DiGraph.add_edge

And in the source code for the method:
https://networkx.github.io/documentation/stable/_modules/networkx/classes/digraph.html#DiGraph.add_edge

That the arguments are, instead: `u_of_edge` and `v_of_edge`.

I assume this is not causing problems for others. I'm using NetworkX version 2.1:
```
>>> import networkx as nx
>>> nx.__version__
'2.1'
```

I assume this complies with the requirements of the library, which is set as `networkx>=2.0`.

Just wanted to see if I am making some obvious mistake or this is indeed something other's have dealt with as well...

gboeing:
Thanks @kuanb. Just checked it out... to add more detail, the following minimal working example succeeds with nx 2.0 but fails with nx 2.1, suggesting that they made an API-breaking change in the minor version release.

```python
import networkx as nx
import osmnx as ox
print(ox.__version__)
print(nx.__version__)
G = ox.graph_from_place('Piedmont, California', network_type='drive')
n, e = ox.graph_to_gdfs(G)
H = ox.gdfs_to_graph(n, e)
fig, ax = ox.plot_graph(H)
```

Looks like a simple fix one line fix. Would you like to do it and open a PR? Otherwise I'll take a look later.

kuanb:
Yep, can do that!

PR #128, opened 2018-02-14
Update argument names in add_edge to support NetworkX 2.1+

kuanb:
Fixes https://github.com/gboeing/osmnx/issues/127

Update `add_edge` to rely on the position of `u` and `v` arrays, as the names of these arguments have been updated/changed from `u` to `u_for_edge ` and `v` to `v_for_edge `.

PTAL and squash/merge if everything looks good!


gboeing:
thanks @kuanb

PR #131, opened 2018-02-19
use haversine formula for great-circle distance

IlyaOrson:
Haversine formula is commonly [prefered](https://gis.stackexchange.com/questions/4906/why-is-law-of-cosines-more-preferable-than-haversine-when-calculating-distance-b) due to smaller floating point errors.


gboeing:
Thanks @IlyaOrson. What are the trade-offs here? That is, can you quickly validate how much precision/accuracy is gained with the haversine method over the cosines method, and what is the difference in performance time?

PR #135, opened 2018-03-03
Add bearing analysis functions

Gbachant:
I added two functions:
One that gives a rounded version of edges' bearing to more easily allow you to group edges of the same and or perpendicular bearing mainly for the purpose of plotting.

And another that collects desired bearing(s) to allow one to chose which bearings to plot and how.

I also made a jupyter notebook example for them that I will submit the the examples repository.

I tried to follow commenting and docstring conventions as much as I could, but if you had any suggestions or anything I would be very grateful.
Thank you

### DIFF
--- a/osmnx/utils.py
+++ b/osmnx/utils.py
@@ -544,7 +544,7 @@ def add_search_bearings(search_bearing, perpendicular=True):
     search_list=[]
     if type(search_bearing) == list:
         for i in search_bearing:
-            search.append(i)
+            search_list.append(i)
             if perpendicular==True:
                 for t in range(3):
                     search_list.append((i+(90*(t+1)))%360)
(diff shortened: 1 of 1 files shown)


gboeing:
@Gbachant thanks for contributing to OSMnx! Can you tell me a little more about your additions? In particular, I want to make sure they are general enough for inclusion into the package itself, rather than just being demonstrated as a standalone example. It looks like the key feature:

```python
for u,v,a in G.edges(data=True):
        a['rounded_bearing']=int(round(a['bearing']))
        a['modulo_bearing']=a['rounded_bearing']%90
```

gives your graph integer bearings and modulo of these integers. This is useful for binning, but as it's only a couple lines of code, it seems easy to leave as standalone in an example notebook. I'm less clear what's happening with the `add_search_bearings` function. What are they being added to? It looks like it just returns a list of bearings. The code inside the function doesn't have any comments so it's hard to see what it's doing and why.

Issue #116, opened 2017-12-18
New feature: Retrieving OSM POIs with osmnx?


HTenkanen:
Hi, 

first of all, awesome work with the package! Highly useful and needed. 

I was wondering if there would be interest to include separate functionalities to retrieve OpenStreetMap POIs with osmnx? I am currently starting to write such functionalities for my own project, and I think I will be using osmnx as a basis for fetching Points of interest from OpenStreetMap (such as restaurants etc.). I know that it is possible to work with [other infrastructures](https://github.com/gboeing/osmnx-examples/blob/master/notebooks/16-work-with-other-infrastructures.ipynb) than roads, but if I have understood correctly that function works only with graphs, not with point data such as POIs. 

If you think this would be interesting, I would be happy to help implementing these functions. 

Cheers,
Henrikki

P.S. FYI: I introduce osmnx also with the fully open GIS course (Lesson 7) that can be found here [http://autogis.github.io](http://autogis.github.io).

gboeing:
@HTenkanen your course looks really cool. Nice work. Yes I think it might be interesting to work with POIs directly in OSMnx... perhaps with a separate new POI module.

HTenkanen:
@gboeing thanks! 👍 

Nice, I think having POIs included in OSMnx would be a nice addition! And yes, I think a separate module would be good, and it could be written e.g. into `pois.py` in a similar manner as you have now the `buildings.py`. Or what would you suggest?

Basically the idea could be that you can pass a list of amenities to a function `pois_from_poly()` that you want to retrieve from Overpass API, and then it would return a GeoDataFrame out of those amenities. This would follow the style that have, and there would be separate functions for `pois_from_bbox()`, `pois_from_place()`, `pois_from_address()` etc.   

I actually already wrote a first version of module with this style (only `pois_from_poly()` now) for my own purposes, and it works quite nicely. There are, however, still some things needs a bit of thinking:

**Should the module provide information about what amenities are available?**
    - e.g. `ox.pois_available` could return a list of possible amenity values
    - This would mean that the amenity list should be updated every now and then...

**The function searches now amenities from** `nodes`, `ways` and `relation`s: 
```
query_template = ('[out:json][timeout:{timeout}]{maxsize};((node["amenity"~"{amenities}"]({south:.8f},'
                  '{west:.8f},{north:.8f},{east:.8f});(._;>;););(way["amenity"~"{amenities}"]({south:.8f},'
[...]

PR #139, opened 2018-03-16
allow custom query filters

### LINKED ISSUE #90: Best practice - custom query parameters
Question: 2 parts:
1. Confirm `infrastructure` kwarg is appropriate way to pass custom OSM queries.
2. Propose conversion of `get_osm_filter` to a class method to allow for customization.

OSM Filter defaults are set as a dictionary assembled every time `get_osm_filter` is called.

1. In my workflow, I'd like to introduce a modification/new filter parameters (drive excluding highways). [This example](https://github.com/gboeing/osmnx-examples/blob/master/notebooks/16-work-with-other-infrastructures.ipynb) suggests that the best way to do this would be to use network type `'none'` and to then pass the custom query through the `infrastructure` kwarg. Is this correct?

2. If the above is correct, I'd like to suggest that, perhaps, a class `OSMSearchFilter` be created with a method `.get_filter(filter_name)`. When initialized, it has some defaults, but also the ability to `.edit_filter(filter_name, modification)` and `.add_filter(filter_name, filter_string)`. 

The 2 makes sense, I'd be happy to sketch it out and submit it.

Update: Upon further review I see that `infrastructure` actually subselects a particular way type from the network values so, in order to do a custom network (in my case prune out `highway=motorway`), I'd need to introduce a new network default into OSM. (Let me know if this is wrong, of course.)

gboeing:
resolves #90 
similar request in #116

### DIFF
--- a/osmnx/core.py
+++ b/osmnx/core.py
@@ -557,7 +557,8 @@ def get_osm_filter(network_type):
 
 def osm_net_download(polygon=None, north=None, south=None, east=None, west=None,
                      network_type='all_private', timeout=180, memory=None,
-                     max_query_area_size=50*1000*50*1000, infrastructure='way["highway"]'):
+                     max_query_area_size=50*1000*50*1000, infrastructure='way["highway"]',
+                     custom_filter=None):
     """
     Download OSM ways and nodes within some bounding box from the Overpass API.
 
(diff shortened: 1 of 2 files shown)


kuanb:
Awesome, thanks!

PR #150, opened 2018-04-04
Custom errors on core.py operations

kuanb:
Updated all generic `ValueError` errors in the `core.py` script to have custom errors (or be replaced by `TypeError` where appropriate). 

This helps with error when using OSMnx.

Example use case: I query for a location that does not have any OSM data. It catches the issue when checking the response JSON and noting the lack of any elements. But it raises a generic `ValueError`. This makes it hard for a method using OSMnx to know if the `ValueError` being thrown is the specific to this issue or something else that should not be handled/excepted in the same way.

Example script to create that error:
```
import osmnx as ox
from shapely.geometry import box
b = box(1.428005,30.787449,1.516929,30.828036)
G = ox.graph_from_polygon(b, network_type='walk', retain_all=True)
```

### DIFF
--- a/osmnx/core.py
+++ b/osmnx/core.py
@@ -49,6 +49,21 @@ class EmptyOverpassResponse(ValueError):
         Exception.__init__(self,*args,**kwargs)
 
 
+class InvalidDistanceType(ValueError):
+    def __init__(self,*args,**kwargs):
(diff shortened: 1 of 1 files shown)


kuanb:
Thanks for the review, @gboeing, spelled "network" correctly this time. :)

PR #206, opened 2018-10-12
Extending `request_nominatim` to support additional Nominatim queries (reverse and lookup)

PedrosWits:
Hi Geoff,

---

### Motivation
The `osmnx.core.nominatim_request` method builds a nominatim request from the following base url:
`https://nominatim.openstreetmap.org/search`. However, as described in the [`Nominatim wiki`](https://wiki.openstreetmap.org/wiki/Nominatim), two other types of queries, reverse geocoding and address lookup, are available using the **reverse** and **lookup** keywords, instead of the **search** keyword which is reserved for geocoding queries. The `nominatim_request` method can be easily extended two support these two other types of queries.

### Changes
Added the Enum class `NominatimService` which represents the type of Nominatim query to run: search, reverse, or lookup. The `nominatim_request` method now takes an additional parameter `service` which is of the `NominatimService` type.

All source changes were made to `osmnx/core.py`.

In hindsight, the Enum class could have been given a better name, something like `NominatimQuery` or `NominatimQueryType`.

### Example

A search nominatim request:
```python
address = "Newcastle A186 Westgate Rd"

params = OrderedDict()
params['format'] = "json"
params['address_details'] = 0
params['q'] = address

response_json = ox.nominatim_request(params = params, service = ox.NominatimService.SEARCH)
```

A lookup nominatim request:
```python
params = OrderedDict()
params['format'] = "json"
params['address_details'] = 1
params['osm_ids'] = "W68876073"

response_json = ox.nominatim_request(params = params, service = ox.NominatimService.LOOKUP)
```

### Tests
Added the test method `test_nominatim` to `test_osmnx.py`, which run successfully on python 3.6, along with the other test methods.

---

Cheers,
Pedro

---
Edited: I'm sorry about the typo in your name.

### DIFF
--- a/osmnx/core.py
+++ b/osmnx/core.py
@@ -65,21 +65,6 @@ class InsufficientNetworkQueryArguments(ValueError):
         Exception.__init__(self,*args,**kwargs)
 
 
-class NominatimService(Enum):
-    __order__ = "SEARCH, REVERSE, LOOKUP"
-
-    SEARCH = auto()  # Node
-    REVERSE = auto()  # Way
-    LOOKUP = auto()  # Relation
-
-    def __str__(self):
-        return str(self.name).lower()
-
-    @classmethod
-    def has_value(cls, value):
-        return any(value == item.value for item in cls)
-
-
 def save_to_cache(url, response_json):
     """
     Save an HTTP response json object to the cache.
@@ -237,7 +222,7 @@ def get_pause_duration(recursive_delay=5, default_duration=10):
     return pause_duration
 
 
-def nominatim_request(params, service = NominatimService.SEARCH, pause_duration=1, timeout=30, error_pause_duration=180):
+def nominatim_request(params, type = "search", pause_duration=1, timeout=30, error_pause_duration=180):
     """
     Send a request to the Nominatim API via HTTP GET and return the JSON
     response.
@@ -246,8 +231,8 @@ def nominatim_request(params, service = NominatimService.SEARCH, pause_duration=
     ----------
     params : dict or OrderedDict
         key-value pairs of parameters
-    service : NominatimService
-        one of the three available nominatim services: search, reverse and lookup
+    type : string
+        Type of Nominatim query. One of the following: search, reverse or lookup
     pause_duration : int
         how long to pause before requests, in seconds
     timeout : int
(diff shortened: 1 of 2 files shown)


PedrosWits:
The build failed on **py27**, because the auto method from the enum package is not recognised. This can be easily fixed by removing the auto() method call and assigning standard 1,2,3 values to the Enum fields.

PedrosWits:
I'm not sure why the coverage index decreased. Should I add extra unit tests for `NominatimService`?

gboeing:
@PedrosWits thanks! This looks really useful. A couple quick comments:

> I'm not sure why the coverage index decreased. Should I add extra unit tests for NominatimService?

It's a pretty small decrease, so don't worry about it. I usually go through and make sure we're not missing any major chunks of code before each minor release.

> The build failed on py27, because the auto method from the enum package is not recognised. This can be easily fixed by removing the auto() method call and assigning standard 1,2,3 values to the Enum fields.

For compatibility and clarity, I wonder if we might simplify the code a little bit. For example, rather than creating a class, we might just check that the requested service is a recognized value, and then format the string. Something like:

```python
known_services = ['search', 'lookup', 'reverse']
if service in known_services:
    url = url.format(service)
else:
    throw SomeUnknownServiceException
```

Thoughts?

PedrosWits:
@gboeing, thanks for the feedback. I agree, the `Enum` class does seem to be overkill. I get carried away sometimes. The method call simplifies to:

```python
ox.nominatim_request(params = params, type = "lookup", ...)
```

One quick question:
- should I make these commits on my fork and make a new pull request
- or is there a way to update the current pull request?

PedrosWits:
> One quick question:
> 
> * should I make these commits on my fork and make a new pull request
> * or is there a way to update the current pull request?

Oops, apparently this is as easy as making new commits to the master branch of my fork.

gboeing:
Push your commits to your fork and the PR should update itself automatically with them.

PedrosWits:
A quick observation: given an edge (u,v,k), its address details can now be quickly retrieved using `nominatim_request`:

```python
osmid = G[u][v][k]['osmid']

params = OrderedDict()
params['format'] = "json"
params['address_details'] = 1
params['osm_ids'] = "W{}".format(osmid)

response_json = ox.nominatim_request(params = params, type = 'lookup')
```

[Example of return response](https://nominatim.openstreetmap.org/lookup?osm_ids=W68876073?address_details=1) 

These details can now be added as attributes to every edge on the network and used in applications such as the ones in https://github.com/gboeing/osmnx-examples. Noteworthy fields: _suburb, city, county, postcode, place_rank, importance_

gboeing:
@PedrosWits thanks! One last thing. Would you update the tests to simplify them a bit? For example, just call the updated function and verify that we get a response back from the API. If we look for specific osmids in the response, the test will probably break over time as edits and additions occur on OSM. That is, a different set of osmid responses should not signify a broken test.

gboeing:
> I'm happy to just call ox.nominatim_request, with different parameters, and make sure it runs without errors.

@PedrosWits yeah would you mind making that final tweak, and then I'll merge the PR? The only reason I prefer to keep the tests loose is because I've had to do so much debugging and code maintenance over the past couple years due to precise tests (I've been surprised to see how often a little thing will change here or there). Hence I'm trying to keep it general for self-preservation.

PedrosWits:
Thanks @gboeing! I'm happy to do that asap.

gboeing:

Issue #198, opened 2018-08-24
Feature: Closest Road to a certain location's coordinates


miskolc:
I'm using osmnx and networkx for calculating shortest routes in Dubai and I'm having an issue when finding the closest road to a tracking location's coordinates. Even if I use coordinates that are right on a specific highway, sometimes it is matched to a close road due to the fact that I use the `osmnx.get_nearest_node` to find the closest road. This function will search for the closest graph node in the OSM map to that specific location. Due to the fact that for certain highways you have very large distances between consective node, the closest node might not be on the same road on which the  tracking location was recorded. It would be nice if there was a method `osmnx.get_nearest_edge` that mapped each location to the closest road.

Here is an example with 2 orig_coordinates that were recorded on the same highway but the second one is matched to a node on a different road.
I recommend using jupyter notebooks to visualize the problem on the map

```python
import networkx as nx
import osmnx as ox
import requests
from matplotlib import pyplot as plt
import matplotlib.cm as cm
import matplotlib.colors as colors
ox.config(use_cache=True, log_console=True)
ox.__version__

# get a graph for some city
SMALL_MARINA = [25.09, 25.06, 55.16, 55.11]
G2 = ox.graph_from_bbox(
    *SMALL_MARINA,
    simplify=False,
    retain_all=True,
    network_type='drive',

)
fig2, ax2 = ox.plot_graph(G2, fig_height=20, fig_width=20)

# get the nearest network node to each point

good_orig_node = ox.get_nearest_node(G2, (25.070661, 55.137911), method='euclidean')
bad_orig_node = ox.get_nearest_node(G2, (25.071764, 55.138978), method='euclidean')
dest_node = ox.get_nearest_node(G2, (25.079223, 55.146241), method='euclidean')



# find the route between these nodes then plot it
route = nx.shortest_path(G2, good_orig_node, dest_node, weight='length')
fig, ax = ox.plot_graph_route(G2, route, fig_height=20, fig_width=20)


# find the route between these nodes then plot it
route = nx.shortest_path(G2, bad_orig_node, dest_node, weight='length')
fig, ax = ox.plot_graph_route(G2, route, fig_height=20, fig_width=20)
```
### Coordinate(South-West) is matched to the right road
<img width="1006" alt="screen shot 2018-08-24 at 17 07 42" src="https://user-images.githubusercontent.com/2085350/44595220-4ea81d80-a7c0-11e8-90a6-61ba02d9578a.png">

[...]

PR #231, opened 2018-12-16
Enable finding nearest road to coordinates

miskolc:
Implementation of easy solution to https://github.com/gboeing/osmnx/issues/198

### DIFF
--- a/osmnx/utils.py
+++ b/osmnx/utils.py
@@ -489,6 +489,48 @@ def get_nearest_node(G, point, method='haversine', return_dist=False):
 
 
 
+def get_nearest_road(G, point):
+    """
+    Return the nearest road to a pair of coordinates.
+    Pass in the graph representing the road network and a tuple with the
+    coordinates. We first get all the roads in the graph. Secondly we
+    compute the distance from the coordinates to the segments determined
+    by each road. The last step is to sort the road segments in ascending
+    order based on the distance from the coordinates to the road.
+    In the end, the first element in the list of roads will be the closest
+    road that we will return as a tuple containing the shapely geometry and
+    the u, v nodes.
+
+    Parameters
+    ----------
+    G : networkx multidigraph
+    point : tuple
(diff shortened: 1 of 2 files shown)


gboeing:
Thanks @miskolc. Would you take a look at #234 which builds on this and offer any comments? If everything looks good to everyone after review, I will probably merge these PRs in tandem.

shafaypro:
Can we get coordinates out of this too?
I am pretty much stuck after calculating the distances, I tried using different methods.
```
import osmnx as osx
lat, lon  = 24.16746, 55.1117
G = osx.graph_from_point((lat,lon), dist=550, network_type='all')
osx.plot_graph(G)
gdf = osx.graph_to_gdfs(G, nodes=False, fill_edge_geometry=True)
roads = gdf[["geometry", "u", "v","ref","name","highway","lanes"]].values.tolist()
roads_with_distances = [(road, osx.geometries.Point(tuple(reversed((lat,lon)))).distance(road[0])) for road in roads]
roads_with_distances = sorted(roads_with_distances, key=lambda x: x[1])
for road in roads_with_distances:
    if str(road[0][-2]).lower() in ['tertiary']:
        print('General Extracted Information: ',road)
        #print('Specific u,v ID : ',road[0][1],road[0][2])
        print("Specific infromation ", G[road[0][1]][road[0][2]])
        #print('Edge Attributes OSMID: ',geo_information_edge_attributes)
        print("osmid :", geo_information_edge_attributes[0]['osmid'])
        break

Need to retrieve the lat and long of the respectable osmid before the break statemetn
```

gboeing:
> Can we get coordinates out of this too?

Yes. The coordinates of the edges are stored in each edge's `geometry` attribute (if it's not just a straightline from node to node). The coordinates of the nodes are stored in each node's `lat` and `lon` or `x` and `y` attributes. For how-to/usage questions, please ask on StackOverflow to reserve the issue tracker for bug reporting and feature development.

PR #232, opened 2018-12-17
Improve functionality of `get_route_edge_attributes`

RunOrVeith:
The behavior of `utils.get_route_edge_attributes` was un-intuitive in 2 ways:

1. The word "attributes" implies that multiple attributes can be retrieved at once.
2. If one edge is missing an attribute, the program exits with a `KeyError`

Both issues are fixed in this pull request:

1. If a specific attribute is not given, all attributes of nodes on the route are returned
2. A new parameter is introduced: It allows to pass a function named `retrieve_default` to `utils.get_route_edge_attributes`. This is a function that is called with the selected edge nodes  and should return the default value desired for this edge. The default behavior is to raise a `KeyError`, like before.

All features are tested in a new test case.
Additionally, I renamed `G.fresh_copy()` to `G.__class__()`  in `utils.induce_subgraph`, as suggested by networkx. This removes a lot of deprecation warnings when running the test suite.

The pull request is fully backwards-compatible, since only one new optional parameter is introduced, and one parameter that was positional is made into an optional parameter.


osmnx/utils.py
```diff
@@ -688,11 +688,14 @@ def get_route_edge_attributes(G, route, attribute, minimize_key='length'):
     route : list
         list of nodes in the path
     attribute : string
-        the name of the attribute to get the value of for each edge
+        the name of the attribute to get the value of for each edge.
+        If not specified, the list of dictionaries containing all attributes is returned.
```
gboeing:
```suggestion
        If not specified, the complete data dict is returned for each edge.
```

tests/test_osmnx.py
```diff
@@ -324,3 +326,33 @@ def test_nominatim():
...
+    import networkx as nx
+    import pytest
+
+    location_point = (37.791427, -122.410018)
+    G = ox.graph_from_point(location_point, distance=500, distance_type='network')
+
+    def on_no_lanes(u, v):
+        return "under_construction"
+
+    path = [65366915, 65326733, 65326736, 65326738, 65319946, 65319944, 65290756, 65312380, 65319942]
```
gboeing: This could change over time and break the tests if new nodes are digitized on OSM.

gboeing: This test feels a bit unnecessarily complex currently for testing `get_route_edge_attributes`.

gboeing:
@RunOrVeith thanks, added a few comments for you.

tests/test_osmnx.py
RunOrVeith: Do you care of there is no test?

gboeing: We can leave it without a test for now.

gboeing: We can leave it without a test for now.

gboeing: @RunOrVeith if you want to make this update, we can merge the PR after

RunOrVeith: I removed the test case

gboeing:

PR #234, opened 2018-12-19
Enable finding nearest edges from list of coordinates

samuelduchesne:
This in part builds on top of #231 by @miskolc and is an answer to issue #198.

The biggest change is in the implementation of the kdTree search applied to virtual points created along all edges of the graph. Therefore the function can take in a list of points (such as building centroids) and find the corresponding edges in a very quick fashion.

This is one of my first pull requests so I am not sure if I have done everything correctly!

### DIFF
--- a/osmnx/utils.py
+++ b/osmnx/utils.py
@@ -640,23 +640,23 @@ def get_nearest_edges(G, X, Y, method='kdtree', dist=10):
     G : networkx multidigraph
     X : list-like
         The vector of longitudes or x's for which we will find the nearest
-        edge in the graph
+        edge in the graph. For projected graphs, use the projected coordinates, usually in meters.
     Y : list-like
         The vector of latitudes or y's for which we will find the nearest
-        edge in the graph
-    method : str {None, 'kdtree'}
+        edge in the graph. For projected graphs, use the projected coordinates, usually in meters.
+    method : str {'kdtree', None}
         Which method to use for finding nearest edge to each point.
+        If 'kdtree' we use scipy.spatial.cKDTree for very fast euclidean search.
         If None, we manually find each edge one at a time using
-        osmnx.utils.get_nearest_edge. If 'kdtree' we use
-        scipy.spatial.cKDTree for very fast euclidean search.
+        osmnx.utils.get_nearest_edge.
     dist : float
-        spacing  distance for the point creation along edges. The smaller, the more points are created.
+        spacing length along edges. Units are the same as the geom; Degrees for unprojected
+        geometries and meters for projected geometries. The smaller the value, the more points are created.
 
     Returns
     -------
-    closest_road_to_point : tuple
-        A geometry object representing the segment and the coordinates of the two
-        nodes that determine the road section, u and v, the OSM ids of the nodes.
+    ne : ndarray
+        array of nearest edges represented by their startpoint and endpoint ids, u and v, the OSM ids of the nodes.
     """
     start_time = time.time()
 
@@ -716,8 +716,8 @@ def redistribute_vertices(geom, dist):
     geom: LineString or MultiLineString
         a Shapely geometry
     dist : float
-        point spacing along edges. The smaller, the more points are created.
-        Units must
+        spacing length along edges. Units are the same as the geom; Degrees for unprojected geometries and meters
+        for projected geometries. The smaller the value, the more points are created.
 
     Returns
     -------
--- a/osmnx/utils.py
+++ b/osmnx/utils.py
@@ -705,6 +705,36 @@ def get_nearest_edges(G, X, Y, method=None, dist=0.0001):
         eidx = extended.loc[idx, 'index']
         ne = edges.loc[eidx, ['u', 'v']]
 
+    elif method == 'balltree':
+
+        # check if we were able to import sklearn.neighbors.BallTree successfully
+        if not BallTree:
+            raise ImportError('The scikit-learn package must be installed to use this optional feature.')
+
+        # transform graph into DataFrame
+        edges = graph_to_gdfs(G, nodes=False, fill_edge_geometry=True)
+
+        # transform edges into evenly spaced points
+        edges['points'] = edges.apply(lambda x: redistribute_vertices(x.geometry, dist), axis=1)
+
+        # develop edges data for each created points
+        extended = edges['points'].apply([pd.Series]).stack().reset_index(level=1, drop=True).join(edges).reset_index()
+
+        # haversine requires data in form of [lat, lng] and inputs/outputs in units of radians
+        nodes = pd.DataFrame({'x': extended['Series'].apply(lambda x: x.x),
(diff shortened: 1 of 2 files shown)


osmnx/utils.py
```diff
@@ -573,6 +618,125 @@ def get_nearest_nodes(G, X, Y, method=None):
...
+    Y : list-like
+        The vector of latitudes or y's for which we will find the nearest
+        edge in the graph
+    method : str {None, 'kdtree'}
+        Which method to use for finding nearest edge to each point.
+        If None, we manually find each edge one at a time using
+        osmnx.utils.get_nearest_edge. If 'kdtree' we use
+        scipy.spatial.cKDTree for very fast euclidean search.
+    dist : float
+        spacing  distance for the point creation along edges. The smaller, the more points are created.
```
gboeing: Are these in the same units as the graph? That is, by default they are in degrees? And for projected graphs, in meters? Would be useful to explain here in the docstring.

osmnx/utils.py
```diff
@@ -573,6 +618,125 @@ def get_nearest_nodes(G, X, Y, method=None):
...
+
+    Returns
+    -------
+    closest_road_to_point : tuple
+        A geometry object representing the segment and the coordinates of the two
+        nodes that determine the road section, u and v, the OSM ids of the nodes.
```
gboeing: you use `road` terminology a couple times here, but probably best to keep it consistent as `edge`

JonasKlasen:
I just played with this and it look great! 

```
import osmnx as ox
import networkx as nx
import time

G = ox.graph_from_place('Piedmont, California, USA')

start1 = time.time()
id_edges = ox.get_nearest_edges(G, X=[37.824885], Y=[-122.243483])
end1 = time.time()

start2 = time.time()
id_edge = ox.get_nearest_edge(G, (37.824885, -122.243483))
end2 = time.time()

print(id_edge[1] == id_edges[0][0] and id_edge[2] == id_edges[0][1])

print(end1 - start1)
print(end2 - start2)
```
resulting timings for my run ware 0.0539 sec and 0.0532 sec

The resulting edge can be visually confirmed against google maps (37.824885, -122.243483).
```
route = nx.shortest_path(G, id_edges[0][0], id_edges[0][1])
fig, ax = ox.plot_graph_route(G, route)
```

osmnx/utils.py
```diff
@@ -573,6 +618,125 @@ def get_nearest_nodes(G, X, Y, method=None):
     return np.array(nn)
 
 
+def get_nearest_edges(G, X, Y, method=None, dist=10):
```
samuelduchesne: The default *dist* should be changed to degrees. Any suggestions for a default value?

samuelduchesne:
> I just played with this and it look great!
> 
> ```
> import osmnx as ox
> import networkx as nx
> import time
> 
> G = ox.graph_from_place('Piedmont, California, USA')
> 
> start1 = time.time()
> id_edges = ox.get_nearest_edges(G, X=[37.824885], Y=[-122.243483])
> end1 = time.time()
> 
> start2 = time.time()
> id_edge = ox.get_nearest_edge(G, (37.824885, -122.243483))
> end2 = time.time()
> 
> print(id_edge[1] == id_edges[0][0] and id_edge[2] == id_edges[0][1])
> 
> print(end1 - start1)
> print(end2 - start2)
> ```
> resulting timings for my run ware 0.0539 sec and 0.0532 sec
> 
> The resulting edge can be visually confirmed against google maps (37.824885, -122.243483).
> 
> ```
> route = nx.shortest_path(G, id_edges[0][0], id_edges[0][1])
> fig, ax = ox.plot_graph_route(G, route)
> ```

I don't think this uses the `kdtree` method. Also, you could try it with a list of points to test against a graph. Should return a list of edge coordinates for each point in the list.

samuelduchesne:
# Quick benchmark
I compare speeds between my method and @migurski's.

I used `ox.graph_from_point()` along with `ox.buildings_from_point()` with incremental distances which yields 0 to 1000 building centroids. These centroids are used for the X,Y coordinates for `ox.get_nearest_edges()`. The *kdtree* method is the vectorized method and the *None* method is the one from #231. Everything is done in projected space.

## Code used
A simple loop that tests both methods with different number of queried points
```
location_point = (40.748433, -73.985656)
times={}  # dict to receive speed results
for method in ['kdtree', None]:
    times[str(method)] = {}
    for distance in np.linspace(100,500,5):
        try:
        # create network from point, inside bounding box
            G = ox.graph_from_point(location_point, distance=distance, distance_type='bbox', retain_all=True)
            G = ox.project_graph(G)
            
            nodes, edges = ox.graph_to_gdfs(G)
            B = ox.buildings_from_point(location_point, distance=distance)
            B = ox.project_gdf(B)

            X = B.geometry.apply(lambda c: c.centroid.x)
            Y = B.geometry.apply(lambda c: c.centroid.y)

            start_time = time.time()
            ne = ox.get_nearest_edges(G,X,Y, method=method, dist=50)
            times[str(method)][len(X)] = time.time() - start_time
        except Exception as e:
            print("{}".format(e))
```

## Results
The kdtree method is consistently faster even for a large number of points as shown in the two figures below. The second figure displays the speeds on a log-scale.

No surprises here since the *None* method implements @migurski's method for each points instead of vectorizing the problem.

speedtest
speedtest-log

## Suggestions
We could test the impact of the *dist* parameter. For now, I used 50 meters, which means that the graph is subdivided in chunks of 50 meters, creating intermediary points.

osmnx/utils.py
gboeing: Yes that's right.

osmnx/utils.py
```diff
@@ -573,6 +618,125 @@ def get_nearest_nodes(G, X, Y, method=None):
...
+
+    Returns
+    -------
+        list of Point geometries : list
+    """
+    if geom.geom_type == 'LineString':
+        num_vert = int(round(geom.length / dist))
+        if num_vert == 0:
+            num_vert = 1
+        return [geom.interpolate(float(n) / num_vert, normalized=True)
```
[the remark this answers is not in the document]
gboeing: I would suggest using `none` as the default method as this works best for unprojected graphs (which are themselves the default) and does not require scipy, which is an optional dependency for OSMnx. A two-dimensional kdtree works only approximately for spheroidal lat-long data, and there should be a docstring comment here recommending it only be used on projected graphs.

Another option here would be to use scikit-learn's balltree with the haversine distance metric to more faithfully measure distances along the earth's curved surface. This would however require (another) new optional dependency in scikit-learn.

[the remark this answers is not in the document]
samuelduchesne: I pushed some changes to address your comments:
- Reverted to None method being the default one + better docstrings
- New: Added balltree method for unprojected graphs
- Adapted tests for new balltree method

samuelduchesne:
I updated the benchmark above to reflect the latest changes

gboeing:

PR #235, opened 2018-12-19
stop changing the type of the node ID to `float` and back to `int`, a…

JonasKlasen:
Avoids changing the type of the node ID to `float` and back to `int`. 
Allow other types than `int` as `osmid`


gboeing:
@JonasKlasen is there an example of node ID not being an integer?

JonasKlasen:
Not in the OSM data, as far as know. However, I have a combination of OSM data and a stop-network which I would like to work with. As I couldn't see a reason for this restriction, I thought this change wouldn't hurt.

JonasKlasen:
I figured, that there is more to do in order to remove the `int` restriction. I close this PR for now and reopen once I have a working PR together.

Issue #223, opened 2018-11-30
Buildings with holes & other polygon data


AtelierLibre:
Hi Geoff,

I wanted to thank you for OSMnx, it has really started to open up quantitative urban analysis for me.

I'm pretty new to python and coding in general but after following your course information I'm hacking together some osmnx, pandas, geopandas, and shapely to try to quantify urban form (Site Coverage, Floor Area Ratio etc.) using OSM data at the scale of city blocks. I have a couple of issues - one a possible bug, the other a proposal/feature request:

**1. Buildings with holes**
Am I right in thinking that osmnx.buildings currently doesn't handle multi-polygons properly i.e. buildings with holes? In OSM a relation tagged as 'building=yes' with two polygons, one tagged 'role=outer', one tagged 'role=inner' seems to end up in osmnx as two individual building polygons on top of each other. This leads to holes in buildings actually being double counted as building area sometimes generating Site Coverages of more than 1 (see attached image).

**2. Generalise building polygon tools**
I have seen that the osmnx.graph_from_* has been generalised to be able to work with infrastructure other than roads. My question is whether you would consider a similar generalisation for osmnx.buildings to enable it to download and work with other polygonal data for example 'landuse=*' or 'place=*' etc.? I am particularly interested in 'place=city_block' and have currently implemented that for myself by duplicating your building module 'building' with 'city_block' but can see that 'landuse=*' and 'amenity=*' etc. would also be useful.

Kind regards,

Nick Bristow

![coverage_ratio - margery_street](https://user-images.githubusercontent.com/36126654/49305728-d03dd480-f4ae-11e8-8e82-a32b6ab0492f.png)

gboeing:
Regarding 1, that's possible. Would you mind whipping up a MWE code snippet to demonstrate this issue? Regarding 2, that could be a nice idea. Perhaps generalizing it from `buildings` to `footprints`. Pull requests welcome!

AtelierLibre:
Regarding 1. Sure, will do.
Regarding 2, that's great, wanted to check you were open to the idea. In reality it's going to take me a while to generate a pull request with the knowledge I have but would like to give it a go.
All the best,
Nick

AtelierLibre:
# Buildings with holes

I've checked the tagging and this building is tagged as OSM guidelines i.e. building tag on the multipolygon relation which contains two polygons, one tagged role=outer the other tagged role=inner

```
import osmnx as ox
%matplotlib inline

point=(51.52753,-0.10999)

[...]

PR #236, opened 2018-12-19
Modify buildings module to work with multipolygon buildings

AtelierLibre:
**Please note I am putting this forward as a suggestion for comment/review**
This is my first pull request. I have not tested it extensively but would appreciate a review / comments to let me know if I am on the right track with the code. If a pull request is not the right way to do this please let me know.

The current buildings.py converts
- polygons tagged as buildings
- polygons part of relations tagged as buildings
to buildings.

This leads to overlapping building footprints where each member of a multipolygon relationship is converted to a building in its own right.

The proposal keeps this behaviour but additionally
- creates multipolygon buildings from the polygons already generated
- deletes the buildings that were created from supporting geometry polygons (i.e. not tagged in their own right)

I haven't included the nodes for the multipolygon buildings as am I unclear how useful that would be.

### DIFF
--- a/osmnx/buildings.py
+++ b/osmnx/buildings.py
@@ -185,6 +185,47 @@ def create_buildings_gdf(polygon=None, north=None, south=None, east=None,
 
                 buildings[result['id']] = building
 
+    # Create multipolygon buildings and pop untagged supporting polygons from buildings
+    pop_list = []
+
+    for response in responses:
+        for result in response['elements']:
+            if 'type' in result and result['type']=='relation':
+
+                outer_polys = []
+                inner_polys = []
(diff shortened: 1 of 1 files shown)


gboeing:
@AtelierLibre thanks! Could you add a commit to include a test for this new functionality in the `tests` folder? For example, using the use case in #223.

AtelierLibre:
@gboeing sure yes, will have a look into this

gboeing:
closes #223

gboeing:
@AtelierLibre just checking back on the test. We should be able to merge this PR after that is committed.

gboeing:

PR #238, opened 2018-12-30
Avoids changing the type of the `osmid`

JonasKlasen:
Avoids changing the type of the `osmid`.
In addition, allow other types than `int` as `osmid`.

### DIFF
--- a/osmnx/save_load.py
+++ b/osmnx/save_load.py
@@ -249,10 +249,7 @@ def load_graphml(filename, folder=None, node_type=int):
     # convert numeric node tags from string to numeric data types
     log('Converting node and edge attribute data types')
     for _, data in G.nodes(data=True):
-        if node_type == int:
-            data['osmid'] = int(data['osmid'])
-        else:
-            data['osmid'] = ast.literal_eval(data['osmid'])
+        data['osmid'] = node_type(data['osmid'])
         data['x'] = float(data['x'])
         data['y'] = float(data['y'])
 
@@ -278,13 +275,11 @@ def load_graphml(filename, folder=None, node_type=int):
                     pass
 
         # osmid might have a single value or a list
-        if 'osmid' in data and node_type == int:
+        if 'osmid' in data:
             if data['osmid'][0] == '[' and data['osmid'][-1] == ']':
-                data['osmid'] = ast.literal_eval(data['osmid'])
+                data['osmid'] = [node_type(i) for i in ast.literal_eval(data['osmid'])]
             else:
-                data['osmid'] = int(data['osmid'])
-        elif 'osmid' in data: 
-            data['osmid'] = ast.literal_eval(data['osmid'])
+                data['osmid'] = node_type(data['osmid'])
 
         # if geometry attribute exists, load the string as well-known text to
         # shapely LineString
--- a/osmnx/save_load.py
+++ b/osmnx/save_load.py
@@ -249,7 +249,10 @@ def load_graphml(filename, folder=None, node_type=int):
     # convert numeric node tags from string to numeric data types
     log('Converting node and edge attribute data types')
     for _, data in G.nodes(data=True):
-        data['osmid'] = node_type(data['osmid'])
+        if node_type == int:
+            data['osmid'] = int(data['osmid'])
+        else:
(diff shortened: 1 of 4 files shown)


JonasKlasen:
I went thru the package and fixed all the `omsid` type alterations. 
Test are finally passing again and the example notebooks work as well.

The changes came along with some small speedups. 

Let me know if this is fine as is?

osmnx/utils.py
```diff
@@ -477,7 +477,7 @@ def get_nearest_node(G, point, method='haversine', return_dist=False):
         raise ValueError('method argument must be either "haversine" or "euclidean"')
 
     # nearest node's ID is the index label of the minimum distance
-    nearest_node = int(distances.idxmin())
+    nearest_node = distances.idxmin()
```
gboeing: Will this invariably return the min index label as an int when working with OSM data? This was originally here because under certain circumstances it would return a float version of the label.

osmnx/save_load.py
```diff
@@ -531,13 +533,12 @@ def graph_to_gdfs(G, nodes=True, edges=True, node_geometry=True, fill_edge_geome
 
         start_time = time.time()
 
-        nodes = {node:data for node, data in G.nodes(data=True)}
-        gdf_nodes = gpd.GeoDataFrame(nodes).T
+        nodes_data = [data for _, data in G.nodes(data=True)]
+        gdf_nodes = gpd.GeoDataFrame(nodes_data, index=G.nodes(data=False))
```
gboeing: Will `G.nodes` return the elements in the same order each time? If it doesn't, then nodes_data and the index will be misaligned. Per the networkx docs, `G.nodes` returns a dict-like interface, and Python dicts are not ordered: https://networkx.github.io/documentation/stable/reference/classes/generated/networkx.Graph.nodes.html

gboeing:
@JonasKlasen thanks, added a couple comments for you.

JonasKlasen:
Did I miss something, where did you add the comments?

gboeing:
@JonasKlasen 

> Did I miss something, where did you add the comments?

No -- sorry I forgot to actually submit the code review! Whoops. It's there now.

osmnx/save_load.py
JonasKlasen: Thanks, now I'm likewise not sure anymore. I will push a fix.

By the way: Happy new year!

osmnx/utils.py
JonasKlasen: This was introduced by the array above, see: `np.array([1,  .2])`

gboeing: Ah right. Thanks.

JonasKlasen:
is there anything I can do to get this going?

gboeing:
@JonasKlasen sorry things are always slow at the beginning of the semester. Lemme take a look.

osmnx/save_load.py
gboeing: Which commit resolved this?

osmnx/save_load.py
```diff
@@ -248,7 +249,10 @@ def load_graphml(filename, folder=None):
     # convert numeric node tags from string to numeric data types
     log('Converting node and edge attribute data types')
     for _, data in G.nodes(data=True):
-        data['osmid'] = int(data['osmid'])
+        if node_type == int:
+            data['osmid'] = int(data['osmid'])
+        else:
+            data['osmid'] = ast.literal_eval(data['osmid'])
```
gboeing: One last question: I see that your logic either converts `data['osmid']` to `int` if `node_type==int`, or it does a `literal_eval` on `data['osmid']`... however the docstring says "node_type : type (Python type (default: int)) – Convert node ids to this type". So if a user passes `node_type=str`, they would expect it to convert `data['osmid']` to a string, but your current logic would `literal_eval` it instead, right?

JonasKlasen: Thanks
Uou are right, I change it, to be more consistent

JonasKlasen:
@gboeing can you have a look. Tests are passing

gboeing:
@JonasKlasen ok thanks, let me take a look

gboeing:

PR #242, opened 2019-01-15
Add support for plotting planar spatial graphs

PedrosWits:
When plotting a graph, if `G` is a planar graph and hence of type `nx.DiGraph`, instead of `nx.MultiDiGraph`, then `plot_graph` will throw a `KeyError` exception for the `keys` attribute. The workaround, which does not affect the output, is to cast to `G` to a Graph when iterating through the edges and their geometries to get the list of lines (which is kind of what is being done anyway).

An example of _downgrading_ `G`, is when using `clean_intersections()` to obtain the intersection points' centroids and then trying to construct a new graph, planar for simplification, using the centroids as nodes.

### DIFF
--- a/osmnx/plot.py
+++ b/osmnx/plot.py
@@ -359,8 +359,8 @@ def plot_graph(G, bbox=None, fig_height=6, fig_width=None, margin=0.02,
     # get north, south, east, west values either from bbox parameter or from the
     # spatial extent of the edges' geometries
     if bbox is None:
-        edges = graph_to_gdfs(G, nodes=False, fill_edge_geometry=True)
-        west, south, east, north = edges.total_bounds
+        nodes = graph_to_gdfs(G, nodes=True, edges=False)
+        west, south, east, north = nodes.total_bounds
     else:
         north, south, east, west = bbox
 
--- a/osmnx/plot.py
+++ b/osmnx/plot.py
@@ -10,6 +10,7 @@ import os
 import json
 import numpy as np
 import pandas as pd
+import networkx as nx
 import matplotlib.pyplot as plt
 import matplotlib.cm as cm
 from matplotlib.collections import LineCollection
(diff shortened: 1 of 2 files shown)


gboeing:
Hi @PedrosWits, thanks for suggesting this. A couple questions:

>  if G is a planar graph and hence of type nx.DiGraph, instead of nx.MultiDiGraph

DiGraph vs MultiDiGraph types shouldn't have any impact on planar vs nonplanar, or vice versa. For instance, a planar graph could be a DiGraph or a MultiDiGraph. Likewise, a MultiDiGraph could be a planar graph or a nonplanar graph. A MultiDiGraph simply allows for parallel edges (hence the `key` attribute) and it certainly could be planar (e.g., if it contains no overpasses or underpasses).

> An example of downgrading G, is when using clean_intersections() to obtain the intersection points' centroids and then trying to construct a new graph, planar for simplification, using the centroids as nodes.

When constructing this new graph (whether planar or not), why not just make it a MultiDiGraph to use OSMnx's plotting functionality as-is?

PedrosWits:
Hi @gboeing,

Thanks for you reply. Yes, you're right. I didn't realise that one can just cast any `nx.Graph` or `nx.DiGraph` to a `nx.MultiDiGraph`, with:
```
G = nx.DiGraph()
nx.MultiDiGraph(G)
```
and do this before calling `plot_graph()`. This was silly, I'm sorry if I wasted any of your time.

Anyway, I wonder if you're interested in extending `clean_intersections()` (or writing a new function) to return the simplified graph rather than just the intersections. I can make a separate pull request once this is ready.

Thanks.

gboeing:
> Anyway, I wonder if you're interested in extending clean_intersections() (or writing a new function) to return the simplified graph rather than just the intersections. I can make a separate pull request once this is ready.

Yes I'd certainly be interested in reviewing this.

PR #249, opened 2019-01-28
Extend clean_intersections to return a new graph, topologically equal to the input graph, but with clusters of nodes merged into a single centroid node.

ppintosilva:
Hi @gboeing. I'm happy to put forward the first version of the new `clean_intersections()` method. The method now returns the simplified graph instead of just the intersection centroids.

I've added comments to the code, which hopefully will be easy to follow, but I'm happy to write down the main steps of the algorithm. The method doesn't appear to be (too) slow, but there is probably room for improvement.

There are a few nuances to look out for, e.g. when choosing between multiple geometries, but I think that the final decision was sensible.

Finally, the simplified graph can end up having a lot of nodes of degree 2 (in its undirected version). Therefore, further simplification can be achieved by removing these nodes. I'm working on this, but maybe this should be a separate method.

I'm looking forward to your input. Cheers, Pedro.


PedrosWits:
Before and after pics: `2700 Shattuck Ave, Berkeley, CA`.
shattuck_before
shattuck_after

samuelduchesne:
This is a feature I was looking forward to! I will try it out and return back with any issues

PedrosWits:
Thanks for trying it out, @samuelduchesne! I'm sure that there a few glitches that I may have miss and room for improvement overall.

osmnx/simplify.py
```diff
@@ -354,4 +371,160 @@ def clean_intersections(G, tolerance=15, dead_ends=False):
...
+                out_edges_neighbor = set(filter(lambda x: x[1] in neighbor_elements, out_edges))
+
+                if len(out_edges_neighbor) > 1:
+                    # If there are multiple edges that fit this criteria, then we need
+                    # to combine their attributes, except for geometry and length
+                    # for which we need to pick a single value
+
+                    edges_attrs = [ G.edges[edge[0], edge[1], 0] for edge in out_edges_neighbor ]
+
+                    all_keys = []
```
gboeing: Can you add more comments to this code block and those that follow to document the logic?

gboeing:
@PedrosWits also, can you explain a bit more about why this is the case:

> Finally, the simplified graph can end up having a lot of nodes of degree 2 (in its undirected version).

osmnx/simplify.py
```diff
@@ -354,4 +371,160 @@ def clean_intersections(G, tolerance=15, dead_ends=False):
...
+            data['geometry'] = LineString([(x1, y1), (x2, y2)])
+
+    # First, we need to find the nearest centroid to each node
+    if cKDTree:
+        # the search is faster if we use a cKDTree
+        # build a k-d tree for euclidean nearest node search
+        centroids = pd.DataFrame({'x': list(map(lambda c: c.x, intersection_centroids)),
+                                  'y': list(map(lambda c: c.y, intersection_centroids))})
+
+        tree = cKDTree(data=centroids[['x', 'y']], compact_nodes=True, balanced_tree=True)
```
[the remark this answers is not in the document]
PedrosWits: We could generalize the method `get_nearest_nodes()` to take in as first argument the points used to form the kd-tree. But that would mean changing every existing call to `get_nearest_nodes` to pass the graph's nodes as input instead of `G`:

```python
nodes = pd.DataFrame({'x':nx.get_node_attributes(G, 'x'),
                      'y':nx.get_node_attributes(G, 'y')})
```
which is a bit of a hassle.

Ideally, we would overload `get_node_attributes()` to keep a single signature, whilst supporting different input types. As overloading is not really a thing in python, we can either:

1. Add a second parameter `points` that defaults to None . Then, either `G` or `points` could be passed by the user, but not both. However, I think this can be a bit confusing and not very elegant.
2. Have a generic method that takes points as input and does nn search, and modify `get_nearest_nodes` internally to call the first method. 

For instance (option 2):

```suggestion
def nearest_neighbors_k2tree(nodes, X, Y):
    """
    Return the nearest neighbors to a list of points, in a 2-dimensional
    space indexed by nodes.
    
    Parameters
    ----------
    nodes : pandas DataFrame with columns 'x' and 'y'
        The data points of dimension 2 to be indexed

    points : ndarray, shape (n,2)
        The n data points of dimension 2 for which we want to find the
        nearest neighbors.

    Returns
    -------
    nn : array
        list of nearest neighbors
    """

    start_time = time.time()

    if not cKDTree:
        raise ImportError('The scipy package must be installed to use this optional feature.')

    # build a k-d tree for euclidean nearest node search
    tree = cKDTree(data=nodes[['x', 'y']], compact_nodes=True, balanced_tree=True)

    # query the tree for nearest node to each point 
    points = np.array([X,Y]).T   
    dist, idx = tree.query(points, k=1)
    nn = nodes.iloc[idx].index

    log('Found nearest nodes to {:,} points in {:,.2f} seconds'.format(len(nodes), time.time()-start_time))

    return np.array(nn)
    
def get_nearest_nodes(G, X, Y, method=None):
...
 
    elif method == 'kdtree':
        
        nodes = pd.DataFrame({'x':nx.get_node_attributes(G, 'x'),
                              'y':nx.get_node_attributes(G, 'y')})
        nn = nearest_neighbors_k2tree(nodes,X,Y)
...   
```

We could do the same for the other methods and even generalise further to k dimensions. But I'm not sure that we gain a lot by this kind of generalisation, especially if it's not used very often. We're effectively building a very thin wrapper around the scipy API.

osmnx/simplify.py
```diff
@@ -325,11 +334,20 @@ def clean_intersections(G, tolerance=15, dead_ends=False):
 
     Returns
```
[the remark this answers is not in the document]
PedrosWits: I was aiming for option a), as the centroids can be easily retrieved from the new graph:

```python
G__ = ox.clean_intersections(G)
intersection_centroids = list(G__.nodes(data = True))
```

PedrosWits:
> @PedrosWits also, can you explain a bit more about why this is the case:

Take for instance the network below. If we only care about intersections where we have multiple turning options (and ignore u-turns), then a lot of these nodes are redundant as the only option is moving forward, along the same geometry. I realise that this is a gross oversimplification of the road network, but I'm guessing it can be useful when analysing major road infrastructure. On the other hand, it may be important to keep these nodes as they represent entry/exit points from the major network. I'm not a 100% sure on this one.

However, identifying these nodes is a bit tricky as not all nodes of degree 2 are redundant, especially the ones near the edge of the graph.

candidate_nodes_degree2_to_remove

PedrosWits:
Main logic behind the method:

1. For each node in G, we find the closest centroid (the cluster to which that node belongs)
2. We create the new graph G' and add the nodes, labelled 1 to n, where n is the number of centroids.
3. To create the edges in the new graph, we iterate through each centroid:
  a. We get the list of nodes that belong to that centroid (we call these 'elements')
  b. For each element of our current centroid, we get its out going edges.
  c. We map the second node of each out-edge (successor), into nodes of G'.
  d. We compute the successors of the current centroid as the set of (mapped) successors of its elements, thus removing duplicates.
  e. For each successor centroid, we merge the existing attributes of the out going edges to that particular centroid.
  f. The 'geometry' with largest length is kept and "fixed" to avoid gaps in the resulting graph (is there a better way of choosing the geometry?)
  g. Finally, we add the edge (centroid, successor) to G'. There is no need to handle predecessors, as the current centroid will be a successor to other centroids.

osmnx/simplify.py
```diff
@@ -354,4 +371,160 @@ def clean_intersections(G, tolerance=15, dead_ends=False):
...
+            y1 = G.nodes[u]['y']
+            x2 = G.nodes[v]['x']
+            y2 = G.nodes[v]['y']
+            data['geometry'] = LineString([(x1, y1), (x2, y2)])
+
+    # First, we need to find the nearest centroid to each node
+    if cKDTree:
+        # the search is faster if we use a cKDTree
+        # build a k-d tree for euclidean nearest node search
+        centroids = pd.DataFrame({'x': list(map(lambda c: c.x, intersection_centroids)),
```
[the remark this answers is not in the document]
PedrosWits: Addressed in eea1bd

osmnx/simplify.py
```diff
@@ -354,4 +371,160 @@ def clean_intersections(G, tolerance=15, dead_ends=False):
...
+    G__ = nx.DiGraph(name = G.graph['name'],
+                     crs = G.graph['crs'])
+
+    # Add the nodes first
+    # The centroids are given new ids = 0 .. total_centroids-1
+    for centroid, elements in centroid_elements.items():
+        x = intersection_centroids[centroid].x
+        y = intersection_centroids[centroid].y
+
+        osmid = elements[0] if len(elements) == 1 else elements
```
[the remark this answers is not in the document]
PedrosWits: Addressed in [eea1bd](https://github.com/gboeing/osmnx/pull/249/commits/eea1bd636509aa4fd5fa7ec9f3dfb5c4ff12066c)

osmnx/simplify.py
PedrosWits: Addressed in [eea1bd](https://github.com/gboeing/osmnx/pull/249/commits/eea1bd636509aa4fd5fa7ec9f3dfb5c4ff12066c)

samuelduchesne:
Hi @PedrosWits, when I use the new function and then apply a `ox.get_undirected()` on the new graph, I get an error inside the `is_duplicate_edge()` function. Your method seems to add osmid attributes as **ndarrays** instead of **lists**. (I have highlighted the array below for *data_other*)


>data= {'lanes': '1', 'bridge': 'yes', 'highway': 'primary', 'oneway': True, 'geometry': <shapely.geometry.linestring.LineString object at 0x1a296900b8>, 'to': 94, 'length': 271.10933596417567, 'ref': 'A695', 'from': 98, 'osmid': [5016040, 95817240], 'name': 'Derwenthaugh Road', 'maxspeed': '70 mph'}

>data_other= {'lanes': ['1', '1'], 'bridge': 'yes', 'highway': ['primary', 'primary'], 'oneway': [True, True], 'from': 94, 'to': 98, 'length': 406.7024179676412, 'geometry': <shapely.geometry.linestring.LineString object at 0x1a29693128>, 'ref': ['A695', 'A695'], **'osmid': [[114882696, 60893148], 505407462]**, 'name': 'Derwenthaugh Road', 'maxspeed': ['70 mph', '70 mph']}

These values come from to same example you provided in tests.

PedrosWits:
Thanks for noticing this, @samuelduchesne. I don't think it is a **ndarray**, but just a nested list. I wasn't sure whether to flatten the lists when merging attributes of several candidate edges between the centroids, but now I have my answer! I wasn't sure because we lose the mapping of attributes to its prior edges, that is:

```
[114882696, 60893148] -> osmids or previous edge A
505407462 -> osmid of previous edge B
```

So, we could potentially get the previous edges back from each new edge, and reverse the clean_intersections transformation if we wanted to. But this is probably not a big deal, as we don't care about the reverse transformation, i.e. going from the cleaned intersections graph back to the original one. We just have the two graphs in memory or store them to disk.

I'll just flatten the resulting lists and submit a new commit.

PedrosWits:
I also want to do a quick speed benchmark to see how the method performs with increasing size of G.

osmnx/simplify.py
```diff
@@ -325,11 +334,20 @@ def clean_intersections(G, tolerance=15, dead_ends=False):
...
-        a GeoSeries of shapely Points representing the centroids of street
-        intersections
+    G__ : networkx multidigraph
+        a new spatial graph, topologically equal to G (approximately), but with
+        clusters of nodes merged into a single centroid node.
     """
 
+    start_time = time.time()
+
+    if 'proj' not in G.graph['crs'] and G.graph['proj'] != "utm":
```
samuelduchesne: Couldn't there also just be the 'init' key in G.graph['crs']?

PedrosWits: At the start of the method, I wanted to test if the graph is projected, otherwise the method won't work as excepted. In my view, an exception should be thrown if that is the case, but I just logged as a warning and moved on.

But I guess you're right, the init key can have some value that specifies that the graph is projected, rather than the default `epsg:4326`. But can that happen in the standard osmnx program flow? You always get an unprojected graph from methods like `graph_from_bbox,` and you need to project it through `ox.project_graph`, thus setting `G.graph['crs']['proj'] = 'utm'`. Or do I have this wrong?

I've also realised that that statement is also wrong as it stands. It should be:

```python
if 'proj' not in G.graph['crs'] or G.graph['crs']['proj'] != "utm":
```

Is there a better way of checking if a graph is unprojected?

gboeing: It might be best to check if it's projected and, if not, issue a warning to the user but proceed anyway. The docstring should make clear that cleaning will occur in the graph's current units (degrees, meters, feet, whatever they may be---but user is advised to not use unprojected/degrees).

But, we shouldn't force them to use a UTM projection. Although the default is UTM, OSMnx allows users to project to any CRS of their choice. So we don't want to hard code anything that requires UTM or that requires meters (as other CRSs can use other units).

To check if the graph differs from the default projection, you can use:

```python
G.graph['crs'] == ox.settings.default_crs
```

PedrosWits: Yes, that is sensible. Thanks, I'll update the statement and docstring.

osmnx/simplify.py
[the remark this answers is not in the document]
PedrosWits: @gboeing I might have found an alternative that reuses `ox.get_nearest_nodes` without the need to modify it. We first create the new graph, add the nodes, and then call `ox.get_nearest_nodes` on that new graph:

```python
# Add the nodes first
# The centroids are given new ids = 0 .. total_centroids-1
for centroid, elements in centroid_elements.items():
    x = intersection_centroids[centroid].x
    y = intersection_centroids[centroid].y

    # we then have to do this later as we don't have the mapping node -> centroid yet
    # osmid = elements[0] if len(elements) == 1 else elements
    G__.add_node(centroid, x = x, y = y)

# get the nodes of G as points
X = [ x for node, x in G.nodes(data = 'x') ]
Y = [ y for node, y in G.nodes(data = 'y') ]
points = np.array([X, Y]).T

nearest_centroids = ox.get_nearest_nodes(G__, points)

# now we have the mapping node -> centroid
```

gboeing:
@PedrosWits would you mind resolving the branch conflicts so we can run the CI tests?

PedrosWits:
> @PedrosWits would you mind resolving the branch conflicts so we can run the CI tests?

Whooops, I'm sorry, I didn't realise this was something that I had to do. I thought I had made sure that the two master branches didn't diverge, but I guess this wasn't the case. And this right was front of me for so long.. erm.. anyway.. it should be ok now.

PedrosWits:
> I also want to do a quick speed benchmark to see how the method performs with increasing size of G.

Looks like the method scales linearly with the number of nodes ([code](https://gist.github.com/PedrosWits/2c52728696397182b34ab96caf99da4d))

quick_benchmark_clean_intersections

There's probably some room for optimisation. I'm guessing the crux is this bit here:

```python
# We also build the reverse mapping: centroid -> osmids
centroid_elements = {i : [] for i in range(len(intersection_centroids))}
for i, osmid in enumerate(G.nodes(data = False)):
    centroid = nearest_centroid[i]
    centroid_elements[centroid].append(osmid)
```

But maybe that is fine for a first iteration. What do you think @samuelduchesne @gboeing?

schoeller:
@PedrosWits 
When trying
```
G = ox.graph_from_address('Bandung, Jawa, Indonesia', distance=3000, network_type='drive', infrastructure='way["highway"~"trunk|primary|secondary|tertiary"]', clean_periphery=True, simplify=True)
G = ox.project_graph(G, to_crs='+proj=utm +zone=48 +ellps=WGS84 +datum=WGS84 +units=m +no_defs +south')
G = ox.clean_intersections(G, tolerance=50, dead_ends=False)
gdf_nodes, gdf_edges = ox.graph_to_gdfs(G)
G = ox.gdfs_to_graph(gdf_nodes, gdf_edges)
```
I receive the following error
```
  File "<ipython-input-6-22a16ce1b815>", line 1, in <module>
    G = ox.gdfs_to_graph(gdf_nodes, gdf_edges)

  File "C:\ProgramData\Miniconda3\lib\site-packages\osmnx\save_load.py", line 610, in gdfs_to_graph
    attribute_values = {k:v for k, v in attributes[attribute_name].items() if pd.notnull(v)}

  File "C:\ProgramData\Miniconda3\lib\site-packages\osmnx\save_load.py", line 610, in <dictcomp>
    attribute_values = {k:v for k, v in attributes[attribute_name].items() if pd.notnull(v)}

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

PedrosWits:
> I receive the following error
> ...
> ```python
> attribute_values = {k:v for k, v in attributes[attribute_name].items() if pd.notnull(v)}
> ```
Hi @schoeller, thanks for trying out the new method!

__Issue__:
The 'osmid' attribute for nodes in the new cleaned graph can be a list as opposed to a single value. This is because the centroids (nodes in the cleaned graph) can be composed by multiple nodes (of the original graph), each with its own osmid. Thus when the condition `pd.notnull(v)` is evaluated and v is a list of osmids, we get an array of booleans, instead of a single true/false value, which causes the ValueError exception. 

__Solution__:
We can store the entire list in a separate attribute, say 'osmids', and store a single value of that list on the 'osmid' key. Does this sound reasonable @gboeing?

---

But other than that it looks like the method is doing the job right.

Before |  After
:-------------------------:|:-------------------------:
bandung_before|bandung_after

gboeing:
That sounds like a good solution.

PedrosWits:
> That sounds like a good solution.

Oops, that solution didn't actually work as we were iterating over every _k,v_ pair.
A few lines below in the method `gdfs_to_graph` the condition `(isinstance(v, list) or pd.notnull(v))`, is used instead of `pd.notnull(value))`, and that works for cases where _v_ is a list. Matching the two seems to solve the issue.

PedrosWits:
I'm not sure why the travis builds are failing when running `conda create --name TESTENV ...`:
```
PackagesNotFoundError: The following packages are not available from current channels:

  - geopandas[version='>=0.4']
  - rtree
  - libspatialindex==1.8
  - rtree[version='>=0.8.3']
  - libspatialindex==1.8
```
I've checked and the packages are available at the listed channels:
```
  - https://conda.anaconda.org/conda-forge/linux-64
  - https://conda.anaconda.org/conda-forge/noarch
```

Any ideas, @gboeing?

gboeing:
@PedrosWits I believe this is resulting from https://github.com/conda-forge/geopandas-feedstock/issues/53

osmnx/simplify.py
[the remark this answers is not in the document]
PedrosWits: @gboeing shall I go ahead with this improvement?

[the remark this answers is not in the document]
gboeing: @PedrosWits I think this looks like a good idea. See also #261 for some related discussion regarding finding nearest edges. cc @juliamgomes

PedrosWits:
Are there any other improvements that you'd like to see @gboeing? I can try to look for performance bottlenecks for large networks as it starts to get rather slow on networks with 10K+ nodes. But other than that, what do you think that is needed to merge this pr?

gboeing:
I think we're nearly there. Let me know when any final changes resulting from code review are finished and I'll take a look. Then we're just waiting on that conda-forge feedstock issue I linked above to run the tests.

PedrosWits:
> I think we're nearly there. Let me know when any final changes resulting from code review are finished and I'll take a look. Then we're just waiting on that conda-forge feedstock issue I linked above to run the tests.

Great, thanks!

gboeing:
@PedrosWits I just re-ran the tests successfully.

osmnx/simplify.py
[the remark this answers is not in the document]
PedrosWits: addressed in [b4caee6](https://github.com/gboeing/osmnx/pull/249/commits/b4caee6902410aa331a926dfac1256f2d330447f)

PedrosWits:
Here's a summary of the changes made at the end of the first review round:
- [x] __New__: Updated docstring and added comments
- [x] __Refactoring__: Simplified the code by reusing `ox.get_nearest_nodes` and `pd.apply`
- [x] __Bugfix__: The new nodes contain both the attributes _osmid_ (single value) and _osmids_ (list).
- [x] __Bugfix__: Flattened the attribute _osmids_ in the new edges (nested lists were causing issues with nx methods)
- [x] __Profiling__: Initial benchmark shows performance issues with networks containing 10k+ nodes (which aren't very large). Further investigation is needed.

We're probably ready for a second round @gboeing . Some things to consider:
- [ ] More detailed performance profiling
- [ ] Routes and their corresponding distance should remain similar for the original and cleaned graphs. Is this the case?
- [ ] Is the choice of geometry for the new edges adequate? It seems to be the case by looking at the plots. However, I notice many instances of new edges with length greater than the expected maximum: max length (among the candidate edges) + 2*tolerance (maximum distance of old node to its centroid, for _u_ and _v_ centroids). So far I haven't come up with an explanation for this.

schoeller:
@PedrosWits :+1: Works for me. Thanks a ton!

gboeing:
@PedrosWits 

> Routes and their corresponding distance should remain similar for the original and cleaned graphs. Is this the case?

Yes, within some tolerance they should be similar (with the distance changing only as a result of intersection centroids shifting a few meters from the original unclustered intersection points.

> Is the choice of geometry for the new edges adequate? It seems to be the case by looking at the plots. However, I notice many instances of new edges with length greater than the expected maximum: max length (among the candidate edges) + 2*tolerance (maximum distance of old node to its centroid, for u and v centroids). So far I haven't come up with an explanation for this.

Yes, they're `LineString` objects, right? That should work fine... which makes me wonder if some other edge case isn't being caught algorithmically and might be working incorrectly in certain circumstances?

PedrosWits:
> Yes, within some tolerance they should be similar (with the distance changing only as a result of intersection centroids shifting a few meters from the original unclustered intersection points.
>
> Yes, they're `LineString` objects, right? That should work fine... which makes me wonder if some other edge case isn't being caught algorithmically and might be working incorrectly in certain circumstances?

Cheers, @gboeing. I think I can look into this, some point at the end of next week. I'll write some tests and let you know the results.

EricAski:
Hi,
I am a newbie with osmnx, can you give a code snippet explaining how to display the simplified graph starting with the following graph 
`G = ox.graph_from_address('Bandung, Jawa, Indonesia', distance=3000, network_type='drive', infrastructure='way["highway"~"trunk|primary|secondary|tertiary"]', clean_periphery=True, simplify=True)
`

Thank you very much

PedrosWits:
> Hi,
> I am a newbie with osmnx, can you give a code snippet explaining how to display the simplified graph starting with the following graph
> `G = ox.graph_from_address('Bandung, Jawa, Indonesia', distance=3000, network_type='drive', infrastructure='way["highway"~"trunk|primary|secondary|tertiary"]', clean_periphery=True, simplify=True) `
> 
> Thank you very much

Hi @EricAski , have a look at the [test file](https://github.com/PedrosWits/osmnx/blob/master/tests/test_osmnx.py) for a couple of examples. All you have to do really is project the graph and then call the clean method. For instance:

```python
G  = ox.graph_from_address('Bandung, Jawa, Indonesia', distance=3000, network_type='drive', infrastructure='way["highway"~"trunk|primary|secondary|tertiary"]', clean_periphery=True, simplify=True)

G_proj    = ox.project_graph(G)
G_cleaned = ox.clean_intersections(G_proj, tolerance = 15)
```

Please note that the pull request has not been merged yet.

gboeing:
@ppintosilva just checking in quick on this PR's status?

ppintosilva:
> @ppintosilva just checking in quick on this PR's status?

Hi @gboeing. I haven't had much time to work on this, but I'll try to look into it over the weekend, or early next week.

ppintosilva:
>     * [ ]  Routes and their corresponding distance should remain similar for the original and cleaned graphs. Is this the case?
> 
>     * [ ]  Is the choice of geometry for the new edges adequate? It seems to be the case by looking at the plots. However, I notice many instances of new edges with length greater than the expected maximum: max length (among the candidate edges) + 2*tolerance (maximum distance of old node to its centroid, for _u_ and _v_ centroids). So far I haven't come up with an explanation for this.

@gboeing , so I just had a look into these two points that I've highlighted previously.

So.. I've computed the shortest paths distances for all pairs in both the original and cleaned graphs and then, for each pair, I took the absolute difference between the two.

One thing that I've noticed is that some nodes in the original graph had a shortest distant greater than 2 times the threshold, but after cleaning, they belong to the same cluster! It's obvious to me now why the second bullet point quoted above doesn't make sense. Two nodes might be close together on a straight line, and hence be assigned to the same cluster, but that doesn't imply that their shortest distance along the network, is anywhere close to that value. That obviously depends on the topology of the network. Please see example below.

So that solves that mystery. But that makes it really difficult to test for correctness. By cleaning the graph we're effectively throwing away information, so there will always be some differences between paths computed in the original vs cleaned graphs. But can we do better than just visual inspection or is that enough?

---
**Example**

- The dictionary key represents the ids of the nodes in the original graph (source, target) of the shortest path.
- `new_edge` is the ids of the nodes in the cleaned graph
- `value` is the shortest distance difference between the two (zero in the cleaned graph because they belong to the same cluster)

(4550426634, 13800765): {'new_edge': (277, 277), 'value': 455.99300000000005}

close_but_far

ppintosilva:
>     * [ ]  More detailed performance profiling

Oh, the performance test that I did previously was actually silly of me. I forgot to take out the time that took to retrieve the data from the server. As the bbox gets larger it obviously takes longer to get the data from the server. Then, the time it takes to clean the graph is actually negligible in comparison.

So these both of the issues that I've highlighted previously don't appear to be issues at all.
I think that we're in a good position to merge this PR @gboeing

gboeing:
@ppintosilva thanks! Can you resolve the merge conflict and I'll take a final reviewing pass then merge?

osmnx/simplify.py
```diff
@@ -354,4 +384,165 @@ def clean_intersections(G, tolerance=15, dead_ends=False):
...
+            # if it doesn't have a geometry attribute, the edge is a straight
+            # line from node to node
+            x1 = G.nodes[u]['x']
+            y1 = G.nodes[u]['y']
+            x2 = G.nodes[v]['x']
+            y2 = G.nodes[v]['y']
+            data['geometry'] = LineString([(x1, y1), (x2, y2)])
+
+    # Let's first create the Graph first
+    G__ = nx.DiGraph(name = G.graph['name'],
```
gboeing: Why did you choose nx.DiGraph here? OSMnx typically works with nx.MultiDiGraph as parallel edges sometimes occur in street networks. Ideally, this would consistently use nx.MultiDiGraph unless there's a clear constraint.

ppintosilva: I think I chose to use a `nx.DiGraph` here, because it's simpler to add the edges, as I don't have to set the key parameter. I do cast it to a nx.MultiDiGraph on the return statement `return nx.MultiDiGraph(G__)`, but yes it might be better to use `nx.MultiDiGraph` from the start to make it consistent. I'll make that update.

gboeing: Thanks. We'd want to make sure we capture any parallel edges from the get-go (including their `key` attributes), or they could get lost in translation.

ppintosilva: I think these will be always lost because effectively we're also clustering (or merging) the edges, not just the nodes.

ppintosilva: Yes, I wasn't sure about this at first, and in fact, reviewing the code now, I'm only scanning the first layer (k = 0) of the input nx.MultiDiGraph, when merging the attributes of multiple edges between two centroids. Trying to recall why I coded it this way:

```python
    edges_attrs = [ G.edges[edge[0], edge[1], 0] for edge in out_edges_neighbor ]
```

gboeing: I sketched out a quick and dirty example... edges A and B are parallel edges (in the graph theory definition) and thus necessitate a MultiDiGraph rather than a DiGraph to model them. We'd want to retain both A and B in the cleaned graph, but if we use a DiGraph at some point in the workflow we'll lose one or the other.

Untitled-1

ppintosilva: So far, given two centroids (large nodes) and the corresponding members (small nodes), the current behavior of `clean_intersections` is to create a **single** new edge in the new graph, connecting the two large circles. And the attributes of this edge have to be chosen in some fashion among the multiple edges that exist between the members of each cluster.

But a better idea, might be to add as many new edges between the two new nodes (centroids) as there are edges between the members of each cluster in the original graph. I think with this solution we would most of the edge information in the original graph.

osmnx_clean_intersections

gboeing: > But a better idea might be to add as many new parallel edges between the two centroids as there are "old" edges.

I'm not sure about that. I think adding a single new edge is the right approach. However, we don't want to lose parallel edges from *non-clustered* nodes, such as in the example image I gave. Can we continue using the current single-edge implementation you've developed, but do a quick check for parallel edges in the original graph, and add them if found?

ppintosilva: Yeah, I can check for parallel edges between clusters with a single member and try to keep those. I need to find a test case, but it shouldn't be to hard to make these changes. I'll let you know.

ppintosilva: @gboeing I've implemented these changes. I've added/updated existing comments and also cleaned up the code a little bit (this includes the addition of an inner function 'fix_geometry' -- which is a bit ugly, but avoids repetition and makes the main loop easier to read). Please let me know what you think.

ppintosilva:
One of the travis builds (py36) timed-out. I guess there's a way to restart/re-run the build? I've noticed that this also happened in other builds, so it shouldn't be a problem with the latest commit.

gboeing:
@ppintosilva thanks!

ppintosilva:
> @ppintosilva thanks!

@gboeing thanks for the support, patience and guidance getting this PR through! Cheers :raised_hands:

psohouenou:
Hi @gboeing, it seems that the new function is not integrated/merged with the master (the latest version of Osmnx still have the version that return a Geoseries). Is there a reason for that?

gboeing:
@psohouenou it was merged into a feature branch that went stale due to some unresolved issues (see discussion in #299). A new PR for similar functionality is now open (see #339) but needs a little more work before it's ready.

psohouenou:
All right. Thank you very much for your prompt reply!

gboeing:
Superseded by #450

Issue #257, opened 2019-02-13
export to osm xml or .pbf?


mxndrwgrdnr:
Hey @gboeing, any plans to add support for exporting an osmnx network back out to an OSM-based format? Many network modeling platforms expect an OSM-formatted network, so it would be great to be able to load a network with osmnx, use osmnx's geometry simplification capabilities, for example, and export back out to an OSM-supported format to use with a different platform. Do you have any suggestions for accomplishing this kind of workflow in the meantime?

gboeing:
No current plans to add functionality to export back to OSM XML. However, if you end up putting this sort of workflow together for your own purposes, a PR would certainly be welcome. Regarding suggestions, you might want to check out the class here: https://github.com/gboeing/osmnx/blob/master/osmnx/utils.py#L1165

mxndrwgrdnr:
OK cool. I'm gonna take a stab at something in the couple days for my own use and I'll submit a PR if it seems to work reasonably well.

gboeing:
closed by #259

PR #259, opened 2019-02-15
New export function to write graph as OSM XML-formatted file

mxndrwgrdnr:
per #257 

- I tested this on a network for Piedmont, CA and validated the output by using Osmosis to convert the output to .pbf without errors. Is there a better way to validate? Open to suggestions here. 
- The hackiest part of this code is [here](https://github.com/mxndrwgrdnr/osmnx/blob/master/osmnx/save_load.py#L187-L192). As best I can tell, osmnx does not currently obtain or preserve these element attributes when querying the Overpass API. Osmosis, however, requires these attributes to be present when it does it's thing and will error out otherwise. Very much open to suggestions here for better ways to handle this.
- The other iffy part of the code is [here](https://github.com/mxndrwgrdnr/osmnx/blob/master/osmnx/utils.py#L123-L126). In the OSM XML format it seems that certain tags are stored as element attributes, while others are stored as sub-elements of their own. I account for this by explicitly specifying which is which in the osmnx settings [here](https://github.com/mxndrwgrdnr/osmnx/blob/master/osmnx/settings.py#L36-L40). This is all fine except in the case that any of these attributes and tags don't also appear in the `useful_tags_node` and `useful_tags_path` settings. They all need to be in one of those lists of useful tags in order to be retained during creation of the graph. Perhaps the more elegant solution would be to include the osm tag and attribute lists in the [`get_node()`](https://github.com/gboeing/osmnx/blob/master/osmnx/core.py#L795-L817) and [`get_path()`](https://github.com/gboeing/osmnx/blob/master/osmnx/core.py#L820-L845) commands, since that's where the actual filtering of tags takes place, but it seemed like it involved fewer changes to the code to just force the osm tags and attributes into the "useful tags" lists whenever `ox.config()` gets called. That being said, it does force those default fields into the "useful tags" lists and that's probably not ideal either. Open to suggestions here as well.

Those are the main pitfalls I see with this PR, but I may be overlooking some stuff so please let me know.

### DIFF
--- a/osmnx/settings.py
+++ b/osmnx/settings.py
@@ -33,11 +33,11 @@ useful_tags_path = ['bridge', 'tunnel', 'oneway', 'lanes', 'ref', 'name',
                     'landuse', 'width', 'est_width', 'junction']
 
 # tags and attributes for generating an OSM XML file
-osm_node_attrs = [
+osm_xml_node_attrs = [
     'id', 'timestamp', 'uid', 'user', 'version', 'changeset', 'lat', 'lon']
-osm_node_tags = ['highway']
-osm_way_attrs = ['id', 'timestamp', 'uid', 'user', 'version', 'changeset']
-osm_way_tags = ['highway', 'lanes', 'maxspeed', 'name', 'oneway']
+osm_xml_node_tags = ['highway']
+osm_xml_way_attrs = ['id', 'timestamp', 'uid', 'user', 'version', 'changeset']
+osm_xml_way_tags = ['highway', 'lanes', 'maxspeed', 'name', 'oneway']
 
 
 # default filter for OSM "access" key. filtering out "access=no" ways prevents
--- a/osmnx/settings.py
+++ b/osmnx/settings.py
@@ -32,6 +32,14 @@ useful_tags_path = ['bridge', 'tunnel', 'oneway', 'lanes', 'ref', 'name',
                     'highway', 'maxspeed', 'service', 'access', 'area',
                     'landuse', 'width', 'est_width', 'junction']
 
+# tags and attributes for generating an OSM XML file
+osm_node_attrs = [
+    'id', 'timestamp', 'uid', 'user', 'version', 'changeset', 'lat', 'lon']
+osm_node_tags = ['highway']
+osm_way_attrs = ['id', 'timestamp', 'uid', 'user', 'version', 'changeset']
+osm_way_tags = ['highway', 'lanes', 'maxspeed', 'name', 'oneway']
+
+
 # default filter for OSM "access" key. filtering out "access=no" ways prevents
 # including transit-only bridges like tilikum crossing from appearing in drivable
 # road network (e.g., '["access"!~"private|no"]'). however, some drivable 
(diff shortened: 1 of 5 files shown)


mxndrwgrdnr:
Two other comments:

- Because the `ox.save_graph_osm()` function has to iterate over every node and every id in order to parse into XML, the runtimes can be pretty severe. I clocked ~30 min to export a network with ~600k nodes and ~1.6M edges. When I was running this myself I was using a [`tqdm`](https://github.com/tqdm/tqdm) wrapper around the iterator to track progress as it went so that I wouldn't freak out and think it had stalled. I didn't want to further bog down the requirements.txt file though so I didn't include that functionality in the PR although I think it made it much nicer.
- On the topic of runtimes, one future improvement might be to skip the `graph_to_gdfs()` conversion and parse the graph object itself directly into XML. Pandas makes life easier for the person writing the code but it definitely slows things down having to iterate over huge dataframes.

mxndrwgrdnr:
@gboeing

osmnx/settings.py
```diff
@@ -32,6 +32,14 @@
                     'highway', 'maxspeed', 'service', 'access', 'area',
                     'landuse', 'width', 'est_width', 'junction']
 
+# tags and attributes for generating an OSM XML file
+osm_node_attrs = [
```
gboeing: maybe name these like `osm_xml_node_attrs` etc to make it a little more explicit for re-use?

requirements.txt
```diff
@@ -1,5 +1,6 @@
 descartes>=1.1
 geopandas>=0.4
+lxml>=4.2
```
gboeing: is it necessary to add `lxml` as a dependency? is there a way to add this functionality (reasonably efficiently) using the python standard library?

gboeing:
> Because the ox.save_graph_osm() function has to iterate over every node and every id in order to parse into XML, the runtimes can be pretty severe.

It might be useful to add a comment that this can be time-intensive into the docstring.

> On the topic of runtimes, one future improvement might be to skip the graph_to_gdfs() conversion and parse the graph object itself directly into XML. 

This would also be more memory efficient.

> I tested this on a network for Piedmont, CA and validated the output by using Osmosis to convert the output to .pbf without errors. Is there a better way to validate?

That seems reasonable. Would you also add a new test to tests/test_osmnx.py to test this function for CI, using a small graph for relatively quick runtime?

osmnx/settings.py
mxndrwgrdnr: Done.

requirements.txt
mxndrwgrdnr: Done.

mxndrwgrdnr: Done.

mxndrwgrdnr:
@gboeing travis is failing trying to create the TESTENV but the only change I made to requirements.txt was to remove lxml. Have you seen this before?

gboeing:
> travis is failing trying to create the TESTENV but the only change I made to requirements.txt was to remove lxml. Have you seen this before?

@mxndrwgrdnr I think this is happening because of recent changes in how conda and conda-forge handle channel priority. I've updated the travis config in the master branch and the tests are working again there. It should fix it here if you rebase your PR from master to incorporate those changes.

mxndrwgrdnr:
@gboeing we good.

gboeing:
Thanks!

Issue #270, opened 2019-04-10
TopologicalError：footprint_type='landuse'


melove93:
Hallo, gboeing . Thank you for osmnx . I used 'footprints_from_place' function
I typed 
point = (48, 10)
dist =2000
gdf = ox.footprints.footprints_from_point(point=point, distance=dist,footprint_type='landuse')
gdf_proj = ox.project_gdf(gdf)

and it works well. But wenn i dist=10000 change,occurs TopologyException.
Error is:
TopologyException: Input geom 0 is invalid: Self-intersection at or near point 10.060335240950593 48.128294710088241 at 10.060335240950593 48.128294710088241
TopologicalError: The operation 'GEOSDifference_r' could not be performed. Likely cause is invalidity of the geometry <shapely.geometry.polygon.Polygon object at 0x00000155BF9DD978>

But wenn i change footprint_type='building', dist=10000, it works well. 
In different latitude and longitude,wenn i type footprint_type='landuse'    ,dist have different upper limits.But footprint_type='building' works well.
Whats the Problem?
![image](https://user-images.githubusercontent.com/49486333/55954462-0bbb8d00-5c5f-11e9-9127-983fe1a86fbe.png)
Thanks

melove93:
I think perhaps the problem is outer_polys and  inner_polys, because the lag-long point nearby is always a large forest belt, in this polygon there are some small polygons such as park,lake. I think may be this causes TopologicalError.

gboeing:
@AtelierLibre recently did some nice work on adapting the footprints module to handle multipolygons... he might have some quick insight into what's going on here?

AtelierLibre:
Hello both, I'm happy to have a look and see what I can find.

AtelierLibre:
@melove93 I haven't been able to resolve this but it does seem to be Relation: Kirchberger Holz (14650) that is causing the problem. I exported it to GQIS and couldn't find a problem with its geometry there. @gboeing it is a multipolygon about 20km long with several ways as outer members and inner polys. I'm speculating at the moment but it would be subdivided by osm_footprints_download wouldn't it? Could this affect it?

AtelierLibre:
Having another look, I think what is happening is that in 'create_footprints_gdf' the part 'try ...' 'exception ...' takes the vertices of every way received from the request and tries to turn them into a polygon.
[...]

Issue #279, opened 2019-04-28
`Polygon has invalid geometry` errors when downloading building footprints


00krishna:
#### Problem description (what did you do, what did you expect to happen, and what actually happened)

I am using the `ox.footprints.footprints_from_polygon()` function to download the building footprints from Tanzania. I pulled the Shapely polygon for the Tanzanian state of Dar-es-Salaam and put that into the function `osmnx` function. 

Here is a snippet of the MULTIPOLYGON:

```
MULTIPOLYGON (((39.49763870239263 -7.187915802001953, 39.49763870239263 -7.189027786254883, 39.49791717529303 -7.189027786254883, 39.49791717529303 -7.189303874969482, 39.49819564819342 -7.189303874969482, 39.49819564819342 -7.189027786254883, 39.49847030639677 -7.189027786254883, 39.49847412109375 -7.18874979019165, 39.49902725219755 -7.18874979019165, 39.49902725219755 -7.188471794128418, 39.50180435180664 -7.188471794128418, 39.50180435180664 -7.18874979019165, 39.50291824340837 -7.18874979019165, 39.50291824340837 -7.189303874969482, 39.50319290161156 -7.189305782318058, 39.50319290161156 -7.189859867095947, 39.50347137451195 -7.189859867095947, 39.50347137451195 -7.19013786315918, 39.50374984741217 -7.19013786315918, 39.50374984741217 -7.190972805023193, 39.50152587890625 -7.190972805023193, 39.50152587890625 -7.190693855285645, 39.50069427490229 -7.190693855285645, 39.50069427490229 -7.190415859222412, 39.5004158020019 -7.190415859222412, 39.5004158020019 -7.190693855285645, 39.50013732910173 -7.190693855285645, 39.50013732910173 -7.190972805023193, 39.49986267089849 -7.190972805023193, 39.49986267089849 -7.191250801086426, 39.49930572509794 -7.191250801086426, 39.49930572509794 -7.190972805023193, 39.49874877929716 -7.190972805023193, 39.49874877929716 -7.190693855285645, 39.49847412109375 .... 
```
dar2 = tanz_data.loc[tanz_data.ID_1==2,'geometry'].unary_union
gdf = ox.footprints.footprints_from_polygon(dar2, retain_invalid=False)

The function will start running and download a bunch of footprints. But then towards the end, I will get errors like the one below. I can see these errors in the terminal associated with the jupyter notebook.

```
[...]

Issue #278, opened 2019-04-26
Input geom 0 is invalid: Self-intersection at or near point 


PierreAttard:
Hello gboing. 

I have seen the issue #270 but I'm not sure it is the same issue. 

# Request

For some location, the method *create_footprints_gdf* has an issue. 

You can find below an example of request. 

results = osmnx.footprints.create_footprints_gdf(polygon=None, north=51.26658135394529, south=51.152831680560726, east=4.55191769225323, west=4.4076010491502684, footprint_type='building', retain_invalid=False)

# The issue
I got these errors when I do this request :
- TopologyException: Input geom 0 is invalid: Self-intersection at or near point 
- shapely.errors.TopologicalError: The operation 'GEOSDifference_r' could not be performed. Likely cause is invalidity of the geometry <shapely.geometry.polygon.Polygon object at >

If I add .buffer(0) for temp_poly=footprints[outer_poly]['geometry'] and temp_poly=temp_poly.difference(footprints[inner_poly]['geometry'].buffer(0)), the issue disappear but it is not satisfactory. 

I use Python 3.7 on windows. 

Thank you a lot.

AtelierLibre:
I think this is the same as issue #270.

Your code generates this error:
TopologyException: Input geom 0 is invalid: Self-intersection at or near point 4.4621621060636354 51.214826758346931 at 4.4621621060636354 51.214826758346931

If you look at OpenStreetMap at that point
(https://www.openstreetmap.org/#map=19/51.21486/4.46238) the castle's outer boundary is made up of two lines.

As in #270 OSMnx currently tries to turn them both into individual polygons but the shorter one creates a self-intersection. buffer(0) may 'fix' the self-intersection but it still won't return the correct geometry for the outer edge of the castle.
![Screenshot from 2019-04-30 09-46-34](https://user-images.githubusercontent.com/36126654/56951907-e240a380-6b30-11e9-912f-621f8d23aedf.png)

gboeing:
Thanks @AtelierLibre, did that `geom.buffer(0)` suggestion resolve it? (That fix has often worked for me in other, previous, but similar circumstances). If so, would you like to create a PR?

AtelierLibre:
@gboeing The exisiting `create_footprints_gdf` only correctly handles polygons formed from single closed ways. The `geom.buffer(0)` may 'fix' the invalid polygons created from chained open ways and avoid the error but it won't create the correct shape for the building.

[...]

PR #283, opened 2019-05-08
Catch key errors when deleting supporting geometry

AtelierLibre:
Delete instead of pop supporting polygons from the final dictionary of footprints. Catch key errors if items in the pop_list are not found in the footprints dictionary - this may be caused by duplicate items in pop_list attempting to delete items from the footprints dictionary multiple times.


AtelierLibre:
@gboeing running it through as a set does seem to resolve the key error, however it may be best to wrap it up in a more complete fix. So happy to hold off on this for now if you think that's a better idea.

gboeing:
@AtelierLibre sounds good

AtelierLibre:
@gboeing I have a new version of `footprints.py` in a branch called `footprints_fix` in my repository. I think it succesfully resolves issues #270, #278, #279 however the change to the code is quite significant.

[https://github.com/AtelierLibre/osmnx/tree/footprints_fix](https://github.com/AtelierLibre/osmnx/tree/footprints_fix)

On my machine #279 runs out of memory, however a smaller area (which still generates 4 responses) does complete.

```
point = (-6.5652,39.1172)
dist =26000
gdf = ox.footprints_from_point(point=point, distance=dist)
```

I introduced a `responses` parameter into `create_footprints_gdf` to allow for pickling between `osm_footprints_download` and `create_footprints_gdf` as some of the `responses` generated are quite large (approx. 500MB).

It needs more testing before I submit a pull request however I wanted to get it out there incase anyone else is able to test it and provide feedback.

Nick

gboeing:
superseded by #289

Issue #280, opened 2019-04-29
Return distances to nearest edges


dHannasch:
It would be convenient to have the capability to keep the distances to nearest edges found by `get_nearest_edges`.
One possible way to do this would be to have `get_nearest_edges` internally call a function that does return the distances returned by the tree query, as shown in https://github.com/dHannasch/osmnx/commit/8f0cc95ca80cb3b0bf0e0322ec0771033fb127b1.

This would, obviously, affect Issue #261. But from reading https://github.com/juliamgomes/osmnx/commit/72b887a53d045ce8468cb1e5766f85abda036d4d, it *looks* like it would be reasonably straightforward to return distances in @juliamgomes' new scheme the same way we can in the current scheme.

juliamgomes:
@dHannasch that makes sense to me! @gboeing @PedrosWits it sounds like we're all on the same page about the implementation so I'll try to get a PR together over the weekend :)

gboeing:
Thanks!

gboeing:
closing this issue here in favor of discussion in #261 and #284

Issue #261, opened 2019-02-23
Feature Request - RTree Implementation of Nearest Edge


juliamgomes:
I would like to create a more efficient implementation of the function `osmnx.utils.get_nearest_edge`, which currently uses the following algorithm:

def get_nearest_edge(G, point):
    """
    Return the nearest edge to a pair of coordinates. Pass in a graph and a tuple
    with the coordinates. We first get all the edges in the graph. Secondly we compute
    the euclidean distance from the coordinates to the segments determined by each edge.
    The last step is to sort the edge segments in ascending order based on the distance
    from the coordinates to the edge. In the end, the first element in the list of edges
    will be the closest edge that we will return as a tuple containing the shapely
    geometry and the u, v nodes.
    Parameters
    ----------
    G : networkx multidigraph
    point : tuple
        The (lat, lng) or (y, x) point for which we will find the nearest edge
        in the graph
    Returns
    -------
    closest_edge_to_point : tuple (shapely.geometry, u, v)
        A geometry object representing the segment and the coordinates of the two
        nodes that determine the edge section, u and v, the OSM ids of the nodes.
    """

When a user has a large graph and must repeatedly query this graph for the nearest edge, it is time-consuming to calculate the distance between the point and every edge of the graph. A faster method would be to generate and cache a rtree corresponding to the graph, then optionally pass it in as an argument to the `get_nearest_edge()` function at query time. 

My feature proposal is the following:
1) Create a GraphRTree class that can convert a graph into an rtree data structure and store the relevant information
2) Either add an rtree as an optional argument in the get_nearest_edge() function, or have an addition get_nearest_edge() method under the GraphRTree class

I implemented this for a different project and found that this implementation allows the query to be completed in just a few milliseconds, even for very large graphs.

gboeing:
@juliamgomes thanks, this might be a good idea. Note that the `get_nearest_edges()` function already [uses trees](https://github.com/gboeing/osmnx/blob/master/osmnx/utils.py#L621) (optionally either a balltree or a kdtree) for faster search. One of those index data structures might work better than an rtree. But more broadly, when you say:

> When a user has a large graph and must repeatedly query this graph for the nearest edge

The `get_nearest_edges()` function already exists to allow for multiple such queries using a single tree data structure to accelerate.

juliamgomes:
[...]

PR #284, opened 2019-05-12
Feature/update get nearest edge methods

juliamgomes:
This is how I envisioned writing a SpatialIndex class that can be passed in to the get_nearest_edge() and get_nearest_edges() functions. The functionality should be the same as before with a few optional additional parameters. Please see feature request #261 for details. 

I have not yet added the option to return distances, as requested in #280 , since that would change the number of items returned by the function, which could cause problems for users who are relying on the current implementation. Happy to add this feature if desired.

It might also be a good idea to update the get_nearest_nodes() function similarly.

Lastly, on the current upstream / master implementation of the get_nearest_edges() function, I noticed on a test case that I received different outputs when using the methods kdtree, balltree and None. This seems like a potential problem and I'm happy to share the test script that gave these results. The implementation in this PR is simply a re-arrangement of the original implementation, so the outputs match those of the current upstream / master version.

Lmk your thoughts, thanks!


dHannasch:
As far as returning distances for #280, could we extract the internals of `KDTreeIndex.get_nearest_edges` to another function and simply discard the distances for the function that returns only the node IDs?
https://github.com/dHannasch/osmnx/commit/e28b9782164b0c8e63024902ccb60e7ad7f1b0e3

It seems like that is in effect what the function does anyway, since `cKDTree.query` returns the distances which are then discarded. We could make a column of the existing GeoDataFrame point to those distances, and then still discard them when they are not wanted.

gboeing:
Thanks @juliamgomes.

> Lastly, on the current upstream / master implementation of the get_nearest_edges() function, I noticed on a test case that I received different outputs when using the methods kdtree, balltree and None. This seems like a potential problem and I'm happy to share the test script that gave these results. 

Can you provide more details on that? If you were working in lat-lng, I suppose we should expect different results between kdtree (since it's Euclidean) and balltree (since it's haversine).

gboeing:
> I have not yet added the option to return distances, as requested in #280 , since that would change the number of items returned by the function, which could cause problems for users who are relying on the current implementation. Happy to add this feature if desired.

@juliamgomes we could handle this by adding a new argument `return_distances=False`, which if set to True would return the distances as well. But by default it would not return them so the behavior would remain consistent with the current implementation.

juliamgomes:
Cool @dHannasch I'll make that change tomorrow in the way that @gboeing suggested, by adding a new argument `return_distanceas=False`. Sorry for the late reply!

@gboeing to elaborate, I tried to test the original and modified function `ox.get_nearest_edges()` on 2 sample coordinates. The results between master and my modified function were consistent. But the results across the methods `balltree`, `kdtree`, and `none` were inconsistent, with `nearest_edges_none = ox.get_nearest_edges(graph, X, Y)` producing unexpected results. This is most likely a mistake on my part, but I'm including the visuals and sloppy debug script just in case. 

Balltree (coordinates in yellow, nearest edge nodes in red):
[
Balltree
](url)

Kdtree:
KdTree

None:
None

```
import osmnx as ox
import matplotlib.pyplot as plt
import numpy as np


def debug_visualize(graph, X, Y, nearest_edges):
    nc = []
    for node in graph.nodes():
        color = "k"
        if node in list(np.array(nearest_edges).flatten()):
            color = "r"
        nc.extend(color)
    fig, ax = ox.plot_graph(graph, node_color=nc, node_size=30, node_zorder=3, show=False, close=False)
    ax.scatter(X, Y, c='y', s=60)
    plt.show()


def main():
    latitude = 37.765860
    longitude = -122.433310
    duboce_triangle = (latitude, longitude)

    philz_latitude = 37.780079
    philz_longitude = -122.420174
    philz = (philz_latitude, philz_longitude)

    graph = ox.graph_from_point(philz, distance=2000, network_type="drive")

    coordinates = [duboce_triangle, philz]
    Y = [coord[0] for coord in coordinates]
    X = [coord[1] for coord in coordinates]
    print("Y coordinates are latitude: {}\nX coordinates are longitude: {}\n".format(X, Y))

    nearest_edges_balltree = ox.get_nearest_edges(graph, X, Y, method='balltree')
    print("balltree:")
    for i in [0, 1]:
        for j in [0, 1]:
            print(graph.node[nearest_edges_balltree[i][j]])
    print()

    nearest_edges_kdtree = ox.get_nearest_edges(graph, X, Y, method='kdtree')
    print("kdtree")
    for i in [0, 1]:
        for j in [0, 1]:
            print(graph.node[nearest_edges_kdtree[i][j]])
    print()

    nearest_edges_none = ox.get_nearest_edges(graph, X, Y)
    print("haversine")
    for i in [0, 1]:
        for j in [0, 1]:
            print(graph.node[nearest_edges_none[i][j]])
    print()

    debug_visualize(graph, X, Y, nearest_edges_balltree)
    debug_visualize(graph, X, Y, nearest_edges_kdtree)
    debug_visualize(graph, X, Y, nearest_edges_none)


if __name__ == "__main__":
    main()
```

PR #289, opened 2019-05-26
Footprints - handle OSM multipolygon relations & fix key error deleting supporting geometry

AtelierLibre:
@gboeing 
Hi Geoff,

I'm putting this forward to replace pull request #283. Essentially it adds new functionality which is to handle multipolygon relations including those made up of chains of open ways. I have also split create_footprints_gdf down into smaller functions to make it easier to read. It also resolves the key error when deleting supporting geometry.

I've tested it against issues #270, #278, #279. As before, on my machine I run out of memory trying to run #279 however a smaller area (which still generates 4 responses) does complete.

point = (-6.5652,39.1172)
dist =26000
gdf = ox.footprints_from_point(point=point, distance=dist)

I have introduced a responses parameter into create_footprints_gdf to allow for pickling between osm_footprints_download and create_footprints_gdf as some of the responses generated are quite large (approx. 500MB).

I remember last time I had to add some test functions so if necessary I'll have a look at that too.

Let me know if you have any comments, otherwise hope it looks good,

Nick


gboeing:
Thanks @AtelierLibre this is looking good. I've closed #283 as this supersedes it. Can you let me know when the tests are ready/committed, and then resolve any merge commits so the CI will run? Thanks!

AtelierLibre:
@gboeing yes, will do. Apologies if this is a basic question but why does the additional `responses=None` parameter in `create_footprints_gdf` cause a conflict?

AtelierLibre:
@gboeing don't worry, I think I'm starting to work it out - I need to resolve the branches first.

AtelierLibre:
@gboeing I think this is ready to go now. I've added some small sample jsons to input_data to help with the testing. Let me know if you have any comments.

Nick

gboeing:
@AtelierLibre thanks! Could you resolve the code conflict, and then I'll test and merge it.

AtelierLibre:
@gboeing no problem, that's done now.

PR #299, opened 2019-06-10
Clean intersections feature + tweaks

gboeing:


gboeing:
@ppintosilva thanks again for all your hard work! I merged your PR into this feature branch to do a little more testing. I found some inconsistencies that maybe you'll have some quick insight into.

Here's the code I'm using to produce two graphs for comparison:
```python
import osmnx as ox

G = ox.graph_from_place('Piedmont, CA, USA', network_type='drive')

G_proj = ox.project_graph(G)
print(len(G_proj.nodes())) #prints 346

G_clean = ox.clean_intersections(G_proj, rebuild_graph=True)
print(len(G_clean.nodes())) #prints 302

ox.save_graph_shapefile(G_proj, filename='G_proj')
ox.save_graph_shapefile(G_clean, filename='G_clean')
```

I load both graphs into QGIS (G_proj in red and G_clean in gray):
graphs-compare


### A few issues jump out at me:

1. The intersection cleaning seems to have lost the self-looping edges (see toward the center-bottom of the image).

2. Toward the top, the cleaned graph is missing a parallel edge that connects to a traffic circle:
graph-edge

3. There's a similar problem here, at the bottom-left of the cleaned graph:
graph-edge-2

4. Lastly, some of the cleaned-up graph nodes seem to be oddly located (or, more accurately, the edges that connect to the new cleaned intersection point are oddly aligned). See here, viewing only the cleaned graph:
graph-node

### Do you have any insight about these issues?

I want to make sure it's ready for prime-time before merging into master.

ppintosilva:
Thanks @gboeing. It's definitely a good idea not to merge straight into master!

Yes, there still seem to be a few issues hanging around. Definitely a tricky one to implement.

### 1 - Self looping edges

I completely forgot about the self-looping edges, so I guess it's no wonder that they're missing. I had to make sure we didn't have self-loops due to occurrences of edges between members of the same centroid, but this is different. Hopefully this will be a quick fix.

### 2- Parallel edges

I believe that the parallel edges disappear in the first example (figure following issue 2) because the two centroids have 3 and 1 members each, and there are two total edges between the combination of its members. As only one edge is kept, the one with longest length is kept and the one with shortest length is discarded.

osmnx_clean_intersections(1)

Similarly in the next figure, the centroids have 2 and 1 members each, and there are two total edges, only one is kept.

The condition to keeping parallel edges (key=1+) is that of a single edge existing between two centroids (either because each centroid has only one member, or because there is only one edge out of the many possible combinations of members between centroids). Then, if that edge has multiple keys, we add all of them to the new graph. However, this may not be the best approach because these sort of _parallel edges_ (not parallel in the sense that they exist with key=1)  or alternatives routes are lost. 

We agree that the alternative, mentioned in the previous PR, is not great. But I'm thinking that a better idea might be to look for non-intersecting geometries (or do some sort of linemerge) and add one element of this set per key to the final graph. I believe this could retain this sort of parallel, non-intersecting edges/routes. We could also maybe add an argument to the function and let the user decide which behaviour is preferable based on trial/error, knowledge of the network, etc.

### 3 - Oddly located nodes

I believe this is one of the pitfalls of reducing multiple nodes to its centroid. It works nicely on roundabouts, due to symmetry, as we can see on pictures 2 and 3; but it doesn't work very well if we have, for instance, two intersections on a straight line. In that case, the centroid will be be the point between those two intersections, which may not represent an actual intersection. We can see this happening in figures 3 and 4. In figure 3, the two red nodes are the members of the centroid, which is located in between the two. Then, the algorithm, out of the two candidate edges, picks the one that has longer length, discarding the second, parallel, edge. Then, because the resulting geometry does not touch the centroid, we find the straight line from the last coordinate to the centroid's centre and complete ("fix") the geometry. This actually makes the resulting graph coincide with its original, even though it looks awkward, and I'm not entirely sure if this is a feature or just a happy coincidence, but it seems to work in general.

Not entirely sure if there is a better way to handle this one.

gboeing:
Superseded by #450

PR #301, opened 2019-06-11
Added which_result list parameter for gdf_from_places; closes #295

### LINKED ISSUE #295: Feature Request - which_result in gdf_from_places
Hey Guys!

I need to query a lot of cities, so I'm using the `gdf_from_places` function, but as you could expect I don't always want the first returned result, so it would be needed to add the `which_result` (available in `gdf_from_place`) to the gdf_from_places function as well.

I have implemented this feature locally, just checking to see if this is something it would be wanted to be implemented upstream.

If so, there is a project decision to be made:
Do we restrict the same which_result for all places (this way we just need to add an optional parameter in the function), or we use a specific parameter for each of the places (which is what I used in my case) that must be passed together in the dictionary, or maybe another idea?

rafaeelaudibert:
Closes issue #295 using the method suggested by @gboeing 

Adds a new parameter `which_results` for `gdf_from_places` function, so that we can choose the `which_result` parameter passed to the `gdf_from_place` function

### DIFF
--- a/osmnx/core.py
+++ b/osmnx/core.py
@@ -489,7 +489,7 @@ def gdf_from_place(query, gdf_name=None, which_result=1, buffer_dist=None):
         return gdf
 
 
-def gdf_from_places(queries, gdf_name='unnamed', buffer_dist=None):
+def gdf_from_places(queries, gdf_name='unnamed', buffer_dist=None, which_results=None):
     """
     Create a GeoDataFrame from a list of place names to query.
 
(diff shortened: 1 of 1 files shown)


osmnx/core.py
```diff
@@ -503,15 +503,23 @@ def gdf_from_places(queries, gdf_name='unnamed', buffer_dist=None):
         later)
     buffer_dist : float
         distance to buffer around the place geometry, in meters
+    which_results : list
+        list (with the same size of queries) of max number of results to return
```
gboeing: if not None, a list of max number of results to return and which to process upon receipt, for each query in queries

osmnx/core.py
```diff
@@ -503,15 +503,23 @@ def gdf_from_places(queries, gdf_name='unnamed', buffer_dist=None):
...
     Returns
     -------
     GeoDataFrame
     """
-    # create an empty GeoDataFrame then append each result as a new row
+    # create an empty GeoDataFrame then append each result as a new row, checking for the presence of which_results
     gdf = gpd.GeoDataFrame()
-    for query in queries:
-        gdf = gdf.append(gdf_from_place(query, buffer_dist=buffer_dist))
+    if which_results != None:
```
gboeing: if which_results is not None:

osmnx/core.py
```diff
@@ -503,15 +503,23 @@ def gdf_from_places(queries, gdf_name='unnamed', buffer_dist=None):
...
+        assert len(queries) == len(which_results), 'Which_results list length must be the same as queries list length'
```
gboeing: assert len(queries) == len(which_results), 'which_results list length must be the same as queries list length'

osmnx/core.py
rafaeelaudibert: Done

osmnx/core.py
rafaeelaudibert: Done

osmnx/core.py
rafaeelaudibert: Done

gboeing:
Thanks!

PR #302, opened 2019-06-13
Allow change overpass endpoint

andreytyu:
Similar to [this](https://github.com/gboeing/osmnx/commit/68fc1065734f792edbafd67d119ba01fd9e0d5e3). 

It's a crucial change for osmnx users from Russia – http://overpass-api.de is unavailable here without a proxy. Some of the other public endpoints (for example https://overpass.kumi.systems/api/ and others provided on the [overpass-turbo](https://overpass-turbo.eu/) website) work fine.

This is my first pull request, I'll be happy to discuss and improve it. It seems to work for me locally.

### DIFF
--- a/osmnx/core.py
+++ b/osmnx/core.py
@@ -185,13 +185,13 @@ def get_pause_duration(recursive_delay=5, default_duration=10):
     """
 
     try:
-        response = requests.get(settings.overpass_endpoint.replace('interpteter', 'status'), headers=get_http_headers())
+        response = requests.get(settings.overpass_endpoint.replace('interpreter', 'status'), headers=get_http_headers())
         status = response.text.split('\n')[3]
         status_first_token = status.split(' ')[0]
     except Exception:
         # if we cannot reach the status endpoint or parse its output, log an
         # error and return default duration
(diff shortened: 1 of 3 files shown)


osmnx/core.py
```diff
@@ -183,14 +183,15 @@ def get_pause_duration(recursive_delay=5, default_duration=10):
     -------
     int
     """
+
     try:
-        response = requests.get('http://overpass-api.de/api/status', headers=get_http_headers())
+        response = requests.get(settings.overpass_endpoint.replace('interpteter', 'status'), headers=get_http_headers())
```
SebMilardo: It looks like there is a typo here: interpteter should be interpreter. Same on line 194

SebMilardo:
I would also suggest some minor changes:

add `.rtrim("/")` at the ends of the paths (to be consistent with what has been done with nominatim), change a little bit how the path is defined (i.e. avoid the `replace`) and at least one unit test in the test file.

I already did this work and I was waiting for a feedback on my issue #328 before submitting a PR. However, If you want you can have a look here [here](https://github.com/SebMilardo/osmnx/commit/a92f7af30b7621db695bdecbec63db41610c7b99) and [here](https://github.com/SebMilardo/osmnx/commit/39649faf535f83cd75b41718ef428cb883700225) and add the changes to your PR

Issue #328, opened 2019-09-02
Feature Request - overpass endpoint


SebMilardo:
Hi

Thank you for the awesome work!
I would like to point osmnx to a local instance of Overpass Turbo. This is also related to #302 
However there is a typo in #302 that I have fixed in my fork and I have also added a test for this new setting. If you are ok with it I can submit a PR.

Kind regards,
Sebastiano

gboeing:
@SebMilardo sounds good, thanks!

SebMilardo:
Pull Request addressing this feature opened at #329

gboeing:
Closed by #329.

PR #329, opened 2019-09-04
Overpass endpoint

SebMilardo:
Very similar to #302 
This PR also adds a Test and `rstrip` at the end of the endpoint


gboeing:
Thanks! Closes #328

PR #339, opened 2019-10-15
ENH: added coarse graining functionality to simplify

### LINKED ISSUE #338: Feature request - coarse graining a network
I would like a functionality to convert a larger street network to one with less amount of nodes: Created by merging nodes that are closer than a threshold. 

Sample usage would be something like:
```python
>>> G = ox.graph_from_address("Berlin, Germany", distance=1000, distance_type='bbox')
>>> G.number_of_nodes(), G.number_of_edges()
(2417, 6163)
>>> H = ox.simplify.coarse_grain(G, radius=100) # collapses all nodes within 100m radius
>>> H.number_of_nodes(), H.number_of_edges() # H is a smaller graph
(800, 2400)
```

debsankha:
This PR resolves #338 that I myself raised. I had the code implemented anyway because I needed this functionality. 

This is my first PR to osmnx. I would like to have your feedback, even in case this PR does not fit the project or significant changes are needed.


gboeing:
@debsankha thanks! How does this PR compare to #249 and #299?

osmnx/simplify.py
```diff
@@ -355,3 +356,58 @@ def clean_intersections(G, tolerance=15, dead_ends=False):
...
+    if not hasattr(buffered_nodes, '__iter__'):
+        # if tolerance is very high, buffered_nodes won't be an iterable of Point's 
+        # but a single point. To take care of this case, we make it into a list.
+        buffered_nodes = [buffered_nodes]
+
+    for node, data in G_proj.nodes(data=True):
+        x, y = data['x'], data['y']
+        lon, lat = data['lon'], data['lat']
+        osm_id = data['osmid']
+        for poly_idx, polygon in enumerate(buffered_nodes):
```
gboeing: Given the nested loops, how does this functionality perform with city-scale graphs? Would it be better performance if we filter by `buffered_nodes` that intersect the current `node`? Why check one at a time instead?

osmnx/simplify.py
```diff
@@ -355,3 +356,58 @@ def clean_intersections(G, tolerance=15, dead_ends=False):
...
+            if polygon.contains(Point(x,y)):
+                poly_centroid = polygon.centroid
+                old2new[node] = dict(label=poly_idx, x=poly_centroid.x, y=poly_centroid.y)
```
gboeing: Are we losing the old node's data dict here? That contains valuable info about the street intersection. Is it possible to retain it here?

samuelduchesne: Looks like it. The graph fails for other useful OSMnx methods such as `graph_to_gdfs()` which in turn is used in a bunch of other methods (eg., plotting).

gboeing:
See #450

PR #240, opened 2019-01-08
Generalize pois queries for diverse tags

chesterharvey:
I modified the `parse_poi_query`, `osm_poi_download`, and `pois_from_place` functions so they accept a `tags` dictionary that allows querying a variety of tags, not just the 'amenity' tag, as previously implemented.

Desired tags are passed to `pois_from_place` with a dictionary like this:
```
tags = {
    'amenity':True,
    'leisure':False,
    'landuse':['retail','commercial'],
    'highway':'bus_stop'
}
gdf = ox.pois_from_place(place='Santa Clara, Santa Clara County, California, USA', tags=tags)
```
A particular key:value pair (e.g., 'highway':'bus_stop') returns elements with that particular pair. A list of values for a certain key (e.g., 'landuse':['retail','commercial']) returns all elements with any of those values. A `True` value (e.g., 'amenity':True) retrieves all elements with that key.

In theory, a `False` value (e.g., 'leisure':False) would filter out any element with that key, but this doesn't seem to work reliably. I don't know if this is an issue with my query construction or a bug with Overpass. To help troubleshoot this, I included an optional boolean parameter (`return_query`) to return the constructed Overpass query alongside the GDF. For example, this is query constructed based on the tag dictionary above:

```
'[out:json][timeout:180];((nwr[~"^(amenity)$"~".*"](37.812276,-122.249374,37.833026,-122.210148);(._;>;););(nwr["leisure"!~".*"](37.812276,-122.249374,37.833026,-122.210148);(._;>;););(nwr[landuse~"retail|commercial"](37.812276,-122.249374,37.833026,-122.210148);(._;>;););(nwr[highway~"bus_stop"](37.812276,-122.249374,37.833026,-122.210148);(._;>;);););out;'
```

PR #342, opened 2019-11-01
Generalize POI to request arbitrary tags

kylebarron:
Similar to the closed PR #240. 

Functions are modified to accept a `tags` argument, which is a dictionary of different key-value pairs. Similar to #240, values of the `dict` can be `True`, `str`, or list of `str`. I.e.
```
tags = {
    'amenity':True,
    'landuse':['retail','commercial'],
    'highway':'bus_stop'
}
```
I didn't try to implement `key=False`, to remove any items of `key` from the search.

In contrast to #240, this leaves the `amenity` parameter in the POI functions, so it doesn't break backwards compatibility. If both `amenities` and `tags['amenity']` are supplied, then they are combined before the Overpass request.

```py
tags = {'tourism': 'motel'}
west = -118.455619
south = 37.330447
east = -118.380432
north = 37.390072
test = create_poi_gdf(north=north, south=south, west=west, east=east, tags=tags)
print(test[['osmid', 'geometry', 'name', 'tourism']].head(5))
```
```
                 osmid                     geometry                     name  \
1925945217  1925945217  POINT (-118.39579 37.36566)          Townhouse Motel   
1925945246  1925945246  POINT (-118.39458 37.36557)       Econo Lodge Bishop   
1925945270  1925945270  POINT (-118.39414 37.36592)        Travelodge Bishop   
1925945363  1925945363  POINT (-118.39573 37.35655)           Super 8 Bishop   
2934354040  2934354040  POINT (-118.39274 37.36029)  Americas Best Value Inn   

           tourism  
1925945217   motel  
1925945246   motel  
1925945270   motel  
1925945363   motel  
2934354040   motel  
```


kylebarron:
The CI fail doesn't seem related to my changes

kylebarron:
I think this CI fail is the same as reported in #347

gboeing:
@kylebarron CI should be fixed in #350

kylebarron:
Great. I can add some tests to fix the coverage fail, but first do you have comments on the changes themselves? For this and #343? I don't want to write the tests before needing to change parts of the PR

osmnx/pois.py
```diff
@@ -76,32 +157,51 @@ def osm_poi_download(polygon=None, amenities=None, north=None, south=None, east=
...
         west, south, east, north = polygon.bounds
 
-        # Parse the Overpass QL query
-        query = parse_poi_query(amenities=amenities, west=west, south=south, east=east, north=north)
-
-    elif not (north is None or south is None or east is None or west is None):
-        # TODO: Add functionality for subdividing search area geometry based on max_query_area_size
-        # Parse Polygon from bbox
-        #polygon = bbox_to_poly(north=north, south=south, east=east, west=west)
+    elif not all(isinstance(x, float) for x in [north, south, east, west]):
```
gboeing: Will this fail if the passed in bounding box contains integer coordinates? They should be valid too.

osmnx/pois.py
```diff
@@ -37,36 +38,116 @@ def parse_poi_query(north, south, east, west, amenities=None, timeout=180, maxsi
...
-        # Parse amenties
-        query_str = query_template.format(north=north, south=south, east=east, west=west,
-                                          timeout=timeout, maxsize=maxsize)
-
-    return query_str
-
-
-def osm_poi_download(polygon=None, amenities=None, north=None, south=None, east=None, west=None,
-                     timeout=180, max_query_area_size=50*1000*50*1000):
+    # Convert input into dict of tags
```
gboeing: Lines 68-98 could use some more comments. First at the top it'd be useful to document more clearly what this chunk of code does and why. Second, throughout this chunk of code it'd be useful to comment more clearly what the line's logic is. It's a little opaque right now.

osmnx/pois.py
```diff
@@ -37,36 +38,116 @@ def parse_poi_query(north, south, east, west, amenities=None, timeout=180, maxsi
...
+            # The `(._; >;);` joins the new result set into the previous result
+            # set: https://wiki.openstreetmap.org/wiki/Overpass_API/Overpass_QL#Item
+            tag_str += '];(._;>;);'
+
+            for t in ['node', 'way', 'relation']:
+                query.append('({t}{tag_str});'.format(t=t, tag_str=tag_str))
+
+    query.append(');')
+    query.append('out;')
+    return ''.join(query)
```
gboeing: One thing that may be useful: could you provide an example returned value as a commented demonstration? What does the resulting string look like at the end?

gboeing:
The code overall looks good. I left a few comments to make it a bit more intuitive to review. I'll check the results of the tests/coverage when you're ready for me to.

osmnx/pois.py
kylebarron: Yes, good catch.

osmnx/pois.py
kylebarron: I think I added a lot of comments, and simplified the code a bit as well, at the risk of being too verbose.

kylebarron:
It might also be helpful to be clearer about which function a user is actually intended to use. I didn't actually realize until right now that there are `pois_from_{point,address,bbox,polygon,place}` functions in `pois.py`, and those are presumably the user-oriented functions. When you read through the [`osmnx.pois` docs](https://osmnx.readthedocs.io/en/stable/osmnx.html#module-osmnx.pois), you see all the inner functions and it's hard to notice that the `pois_from_*` functions exist.

kylebarron:
Given this test https://github.com/gboeing/osmnx/blob/c16f10b26e2066fb474da4dbec51599dcf35660b/tests/test_osmnx.py#L360
it looks like historically if no parameters were provided, it was akin to the current `tags={'amenity':True}`. Right now I require either `amenity` or `tags` to be not `None`, but I could switch to the former if you wanted.

Also just a comment, I'm used to running a formatter like [yapf](https://github.com/google/yapf), [black](https://black.readthedocs.io/en/stable/), and/or [isort](https://github.com/timothycrosley/isort) through my code to keep all the code style consistent, especially when working with other people. There are some portions of the code with really long lines and generally inconsistent style that could be helped by using one of these tools.

gboeing:
Circling back around on outstanding PRs now that the semester is over and I can merge them... @kylebarron are you interested in getting this wrapped up and merged? Looks like it's nearly done.

> Right now I require either amenity or tags to be not None, but I could switch to the former if you wanted.

I think that change would be for the best.

> I'd support prefixing each of the non pois_from_* functions with _, which would remove them from the docs pages, though that would break backwards compatibility.

This needs to be done throughout the package, and will be a bigger refactor. Let's leave it alone for this specific feature for now.

kylebarron:
I think it was very close to finished, but I haven't looked at it in 5 months, and I don't use `osmnx` anymore.

gboeing:
@kylebarron ok, no prob.

gboeing:
superseded by #449

AnoopDusadh0406:
@gboeing sir I want to find out list of hospital and  hostel( name & address) for a given city . I tried pois function and geometries_from_place and use tag =  {'Hospital': True}  but did not find any solution . can you help me for finding solution of these type queries. Thanks advanced.

PR #343, opened 2019-11-01
Use local DEM file for elevations

### LINKED ISSUE #341: Feature Request: New elevation provider
Hello! This is an amazing project, and I'm happy to finally have a project to test out osmnx.

I'm working with OSM data of trails, and adding elevation data to trails is important. Currently, the only elevation provider supported is Google Maps, and this has a couple drawbacks. For one they're currently restricting API usage: 
> Notice: New customers will need to contact sales for access to the Elevation API.

This might mean they only care about large-scale, large-budge projects, and in any case restricts small developers from using it. There are a couple other elevation APIs, but they all have the drawbacks of being web services.

I wrote code for myself to import elevations from a local DEM using `rasterio`. Would you accept a PR for this? 

Using local DEMs are more scalable than APIs. They can be more accurate since most APIs use at most 30m SRTM data, and there exists 10m data for the contiguous US and 1m for some cities from new LIDAR data.

Maybe there could also be a helper function to download predefined types of elevation DEMs, like SRTM, 3DEP for the US, etc. for the area covered by the current Graph.

Drawbacks:

Using `rasterio` would add a dependency on GDAL, which is not insignificant. I'd propose that it not be included in `requirements.txt`, but that a user who's interested in using this function could self install, and it would be imported conditionally.

kylebarron:
I'd already written this code, so it was a quick addition. There's a little bit of duplication in `add_node_elevations_from_dem` from `add_node_elevations`, but I wanted to do the first draft quickly. Not sure whether you would prefer adding a new, separate function, or as a new parameter in `add_node_elevations`. 

#### Outline:

- `add_node_elevations_from_dem` is just a wrapper to send individual points to `query_dem`. 
- Inside `query_dem`, it finds the cell closest to the given latitude/longitde: `x, y = dataset.index(lon, lat)`.
- `rasterio` lets you read a [`window`](https://rasterio.readthedocs.io/en/stable/topics/windowed-rw.html) of the DEM, so that you only load a tiny bit of data into memory. (Haven't tested if this is a problem with extensions other than `.img`).
- It works without issue if you just pass a window to get the single box `x, y`, but especially with rasters with less resolution, interpolating is useful. So alternatively, the user can ask for multiple cells bordering `x, y` and then use scipy's interpolation to find hopefully a more accurate elevation value.

#### To do:

- [x] Right now this fails silently when the provided lat/lon values are outside of the DEM's extent. `dataset.index(lon, lat)` looks like it gives coordinates in the DEM's CRS without raising an error. `dataset.index(0, 0)` gives a response; but then passing those `x`, `y`, values into `dataset.read()` gives an empty numpy array because they don't exist. So when I call `dataset.index(lon, lat)`, I need to check the width and height of the DEM's box, and raise an error there if it's outside.

#### Example:

First download USGS data [here](https://prd-tnm.s3.amazonaws.com/StagedProducts/Elevation/13/IMG/USGS_NED_13_n33w117_IMG.zip) (150MB download, 450MB extracted). Then extract the ZIP file.

```py
import osmnx as ox
west = -116.66587829589844
south = 32.60120475753773
north = 32.794201303793194
east = -116.46263122558594
G = ox.graph_from_bbox(north=north, south=south, west=west, east=east)
dem_path = 'USGS_NED_13_n33w117_IMG.img'

new_g = add_node_elevations_from_dem(G, dem_path)
grades = add_edge_grades(new_g)
```

Closes #341

### DIFF
--- a/osmnx/elevation.py
+++ b/osmnx/elevation.py
@@ -9,16 +9,154 @@
 import math
 import networkx as nx
 import pandas as pd
+import rasterio
 import requests
 import time
+from scipy.interpolate import interp2d
 
 from .core import get_from_cache
 from .core import save_to_cache
 from .utils import log
 
 
-def add_node_elevations(G, api_key, max_locations_per_batch=350,
-                        pause_duration=0.02): # pragma: no cover
+
+def add_node_elevations_from_dem(G, dem_path, num_buffer=1, interp_kind='linear'):
+    """
+    Add elevation for each node in network using a Digital Elevation Model (DEM)
+    and add to node as an attribute. Measurement unit is unchanged from the
+    source DEM; most DEM files are in meters.
+
+    DEM sources include SRTM, which has worldwide 1 arc-second (30m) coverage,
+    and USGS 3DEP, which has 1/3 arc-second (10m) and 1 arc-second (30m)
+    coverage for the continental United States.
+
+    Parameters
+    ----------
+    G : networkx multidigraph
+    dem_path : str
+        path to DEM file with elevation data. Can be any file format readable by
+        GDAL.
+    num_buffer : int
+        number of bordering cells around (lon, lat) to use when interpolating
+    interp_kind : str
+        kind of interpolation. Passed to scipy.interpolate.interp2d. Can be
+        ['linear’, ‘cubic’, ‘quintic']
+
+    Returns
+    -------
+    G : networkx multidigraph
+    """
+
+    # make a pandas series of all the nodes' coordinates as 'lat,lng'
+    # round coorindates to 5 decimal places (approx 1 meter) to be able to fit
+    # in more locations per API call
+    node_points = pd.DataFrame([(node, data['x'], data['y'])
+                                for node, data in G.nodes(data=True)],
+                               columns=['node', 'lon', 'lat'])
+    node_points = node_points.set_index('node')
+
+    results = []
+    for row in node_points.itertuples():
+        value = query_dem(dem_path=dem_path, lon=row.lon, lat=row.lat)
+        results.append(value)
+
+    # sanity check that all our vectors have the same number of elements
+    if not (len(results) == len(G.nodes()) == len(node_points)):
+        raise Exception(
+            'Graph has {} nodes but we received {} results from the elevation API.'
+            .format(len(G.nodes()), len(results)))
+    else:
+        log('Graph has {} nodes and we received {} results from the elevation API.'
+            .format(len(G.nodes()), len(results)))
+
+    # add elevation as an attribute to the nodes
+    node_points['elevation'] = results
+    # round elevation to millimeter
+    node_points['elevation'] = node_points['elevation'].round(3)
+    nx.set_node_attributes(G,
+                           name='elevation',
+                           values=node_points['elevation'].to_dict())
+    log('Added elevation data to all nodes.')
+
+    return G
+
+
+def query_dem(dem_path, lon, lat, num_buffer=1, interp_kind='linear'):
+    """Query elevation data for given point
+
+    Parameters
+    ----------
+    lon : float
+        longitude
+    lat : float
+        latitude
+    num_buffer : int
+        number of bordering cells around (lon, lat) to use when interpolating
+    interp_kind : str
+        kind of interpolation. Passed to scipy.interpolate.interp2d. Can be
+        ['linear’, ‘cubic’, ‘quintic']
+
+    Returns
+    -------
+    value : float
+        elevation in terms of the unit of the DEM (usually meters)
+    """
+    # Read metadata of file
+    dataset = rasterio.open(dem_path)
+
+    # Find x, y of elevation square inside raster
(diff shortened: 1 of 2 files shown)


kylebarron:
A further improvement could be to supply a _list_ of `dem_paths`. I personally am interested in getting elevation for OSM files that cross the 1 degree lat/lon boundaries, and thus would need to use more than one DEM file. To use more than one DEM, you'd either have to combine them ahead of time using GDAL into one file, or look across several DEM's. 

Reading DEM metadata is fast, so looking into a list of DEM's wouldn't be slow. But I'd have to think about what to do for nodes at lat/lon boundaries. Especially if trying to use interpolation around a box of elevation values, you'd have to handle cases where ideally you'd want to load data from two separate DEM files.

kylebarron:
This PR is now extended to accept a list of DEM files. For each point that I'm finding the elevation of, I check which files hold the nearby grid of lat/lon/elev pairs (this depends on the `num_buffer` argument). If more than one file is covered by this grid, I load both, get the data of interest from each file, then just pass it on to `scipy.interpolate.interp2d`. This makes my life simple because the grid from each DEM file could be irregular. And especially in the case where a lat/lon point happens to lie near the corner of a file, where it could be necessary to load four files.


Example:
Download both the files 
[`USGS_NED_13_n33w117_IMG`](https://prd-tnm.s3.amazonaws.com/StagedProducts/Elevation/13/IMG/USGS_NED_13_n33w117_IMG.zip) 
[`USGS_NED_13_n34w117_IMG`](https://prd-tnm.s3.amazonaws.com/StagedProducts/Elevation/13/IMG/USGS_NED_13_n34w117_IMG.zip) 
450MB extracted each

```py
import osmnx as ox
west = -116.66587829589844
south = 32.90120475753773
north = 33.094201303793194
east = -116.46263122558594
G = ox.graph_from_bbox(north=north, south=south, west=west, east=east)
dem_paths = ['USGS_NED_13_n33w117_IMG.img', 'USGS_NED_13_n34w117_IMG.img']
G = add_node_elevations_from_dem(G, dem_paths)
G = add_edge_grades(G)
```

I am getting warnings from `interp2d`, namely:
```
/Users/kylebarron/anaconda/envs/create-database/lib/python3.7/site-packages/scipy/interpolate/_fitpack_impl.py:976: RuntimeWarning: No more knots can be added because the number of B-spline
coefficients already exceeds the number of data points m.
Probable causes: either s or m too small. (fp>s)
	kx,ky=1,1 nx,ny=7,6 m=18 fp=0.000703 s=0.000000
  warnings.warn(RuntimeWarning(_iermess2[ierm][0] + _mess))
```
This SO post looks like it covers the issue, but it's long and haven't read it quite yet: https://stackoverflow.com/questions/3526514/problem-with-2d-interpolation-in-scipy-non-rectangular-grid

I'm guessing that these warnings are only occurring when using multiple grids across DEM files. I've noticed (at least with this data from USGS) that there's usually a very small overlap across files, where cells exist in both, but that the elevation values are slightly different, i.e. same lat/lon in each file but the elevation is like half a meter apart. `interp2d` might be wanting uniquely specified x,y,z values.

kylebarron:
Also, I'm realizing that this code only adds elevations to _nodes_, and with `simplify_graph=True`, some edges might be long LineStrings. Would you be ok with a refactor to add `add_elevations_to_edges`? This would take each edge's geometry, add for each `Point` in the `LineString`, get the elevation, create a `LineString Z` and replace it as the geometry.

This PR is already significant as it is. Should I wait till this is merged and then open a new PR for adding elevations to edges?

gboeing:
I'm wondering if this new DEM code belongs in a separate elevation_dem module, leaving the original elevation module for the Google API codebase. Thoughts?

I agree that edge-elevation work belongs in a separate PR once we deal with this one fully.

> I am getting warnings from interp2d

Have you had luck further pinning down this cause?

kylebarron:
I agree about splitting it from the other elevation code. I was also musing about splitting this code into a separate package, that osmnx then depends on. It doesn't look like there's an existing package that wraps `rasterio` in this way, and I could see people being interested in using the wrapper outside osmnx. Thoughts?

kylebarron:
And no, I haven't debugged the `interp2d` warnings any further.

kylebarron:
Also, since I made the PR, I've learned about GDAL's [Virtual Rasters](https://gdal.org/drivers/raster/vrt.html) (.vrt), which can simplify the code a lot. Essentially I was trying to do extra work to figure out when a point is right at the edge of two raster files. But building VRTs is fast, and then GDAL (and I expect rasterio as well) treat multiple files as a single one, without having to merge the underlying files.

kylebarron:
I did a bit more research on existing packages to make sure I'm not reinventing the wheel, and the only candidate I found is [gis-tools](https://github.com/benjaminpillot/gis-tools) and that's GPL3 licensed. I think I'm going to make a simple MIT-licensed [package](https://github.com/kylebarron/demquery) to query points on a DEM, and then update this PR to use that.

PR #345, opened 2019-11-10
[WIP] Refactors package to separate downloading capabilities from the rest [WIP]

pedrocamargo:
* Refactors to separate downloading capabilities from the rest
* import errors
* circumvents NetworkX lack of algorithm


pedrocamargo:
@gboeing , This is my first stab on it (I took the opportunity to fix that error that was appearing in some other pull requests (test_stats), which you might need to evaluate better. I may have made the wrong fixing there).

In case you agree with where I am taking this, I will start the integration with AequilibraE and then we can see if this undertaking works as expected for everybody.

kylebarron:
A refactor seems useful because `core.py` is so large. +1

pedrocamargo:
@gboeing ?

gboeing:
@pedrocamargo the CI tests were fixed in #350, can you undo your changes to the tests module that worked around it?

gboeing:
Also: reading through the file changes, there's a lot to process in your proposed refactor. Could you succinctly document here in a comment what's changed and why? That'll help both for the review and for future reference.

pedrocamargo:
Document it in code or in here?

gboeing:
Here for now.

pedrocamargo:
@gboeing , the test has been reversed

pedrocamargo:
@gboeing 
As mentioned in the PR, the main objective of the PR is to separate the OSM downloading capabilities from the geo-processing and graph-processing algorithms
The list of changes include:

1. Separating all the custom errors into a file dedicated to them
2. Moving the routines from utils which are applicable to geo-processing  into geo_utils.py
3. Moving all routines for OSM downloading into downloader.py
4. Moving the OSM content handler into its own file (It makes sense having a well-defined class like that into its own file)

With these changes, anybody that wants to use OSMNx only to download OSM data and wrangle the data with different tools won't have to deal with more complicated dependencies, which is usually an issue on Windows.
This way OSMNX becomes a more attractive dependency for projects that don't require that copious amount of dependencies, AequilibraE being an example of such.

gboeing:
@pedrocamargo this has been merged and will be released shortly in v0.11

pedrocamargo:
Thanks, @gboeing !

PR #351, opened 2019-11-28
Add node elevations from DEM files

kylebarron:
Refactor of #343 to use [`demquery`](https://github.com/kylebarron/demquery).

**Overview**:

- I split my code into a new Python package on PyPI. 
- Since learning about [Virtual Rasters](https://gdal.org/drivers/raster/vrt.html), the code to find elevations is an order of magnitude simpler. I don't do any joining of nearby rasters; GDAL does all of that.
- I discovered that `interp2d` gives those warnings only when the grid is not the expected size for the type of interpolation: 3x3 for linear interpolation, 5x5 for cubic interpolation, and 7x7 for quintic interpolation. So I removed the option to manually choose the size of the interpolation region. If `interp_kind` is `linear`, it creates a 3x3 grid.
- Checks for nodata values; the virtual raster can have regions with no data that are inside the file's bounds, if the provided files didn't create an exact rectangle. My package raises an error if it sees those values in the interpolation region.
- I'm working on writing tests in the demquery package. ~~I'm having issues with rasterio working on Travis CI~~ the osmnx `.travis.yml` was quite helpful.
- Do you have plans to deprecate Python 2.7 and 3.4? Python 3.4 is already end of life, and 2.7 is in a month. I made `demquery` 3.5+ compatible, but haven't tested Python 2.7.
- Right now, `demquery` is only on PyPI, and I think that's why the current Travis build is failing.

**Example**:

Download both the files 
[`USGS_NED_13_n33w117_IMG`](https://prd-tnm.s3.amazonaws.com/StagedProducts/Elevation/13/IMG/USGS_NED_13_n33w117_IMG.zip) 
[`USGS_NED_13_n34w117_IMG`](https://prd-tnm.s3.amazonaws.com/StagedProducts/Elevation/13/IMG/USGS_NED_13_n34w117_IMG.zip) 
450MB extracted each

```py
import osmnx as ox
west = -116.66587829589844
south = 32.90120475753773
north = 33.094201303793194
east = -116.46263122558594
G = ox.graph_from_bbox(north=north, south=south, west=west, east=east)
dem_paths = ['USGS_NED_13_n33w117_IMG.img', 'USGS_NED_13_n34w117_IMG.img']
# interp_kind can be None, linear, cubic, or quintic
G = add_node_elevations_from_dem(G, dem_paths, interp_kind=None)
G = add_edge_grades(G)
```

### DIFF
--- a/osmnx/elevation.py
+++ b/osmnx/elevation.py
@@ -6,6 +6,7 @@
 # Web: https://github.com/gboeing/osmnx
 ################################################################################
 
+import demquery
 import math
 import networkx as nx
 import pandas as pd
@@ -91,6 +92,58 @@ def add_node_elevations(G, api_key, max_locations_per_batch=350,
 
     return G
 
+
+def add_node_elevations_from_dem(G, dem_paths, band=1, interp_kind=None):
+    """
+    Get the elevation of each node in the network from DEM files and add it to
+    the node as an attribute.
+
+    Parameters
+    ----------
+    G : networkx multidigraph
+    dem_paths : list of str or pathlib.Path
+        list of paths to DEM files. DEM files can be any format readable by
+        GDAL.
+    band : int
+        band of DEM file to query data from; 1 by default.
+    interp_kind : None or str
+        one of None, 'linear', 'cubic', 'quintic'. None will do no interpolation
+        and choose the value in the DEM closest to the provided point. linear
+        creates a 3x3 grid and runs linear interpolation; cubic creates a 5x5
+        grid and runs cubic interpolation; quintic creates a 7x7 grid and runs
+        quintic interpolation
+
+    Returns
+    -------
+    G : networkx multidigraph
+    """
+
+    nodes = [(node, data['x'], data['y']) for node, data in G.nodes(data=True)]
+    df = pd.DataFrame(
+        nodes, columns=['node_id', 'lon', 'lat']).set_index('node_id')
+    points = df[['lon', 'lat']].values
+    query = demquery.Query(dem_paths, band=band)
+    elevations = query.query_points(points)
+
+    # sanity check that all our vectors have the same number of elements
+    if not (len(elevations) == len(G.nodes()) == len(df)):
+        raise Exception(
+            'Graph has {} nodes but we received {} results from the elevation API.'
+            .format(len(G.nodes()), len(elevations)))
+    else:
+        log(
+            'Graph has {} nodes and we received {} results from the elevation API.'
+            .format(len(G.nodes()), len(elevations)))
+
+    # add elevation as an attribute to the nodes
+    df['elevation'] = elevations
+    df['elevation'] = df['elevation'].round(3)  # round to millimeter
+
+    nx.set_node_attributes(
+        G, name='elevation', values=df['elevation'].to_dict())
+    log('Added elevation data to all nodes.')
+
+    return G
 
 
 def add_edge_grades(G, add_absolute=True): # pragma: no cover
--- a/osmnx/elevation.py
+++ b/osmnx/elevation.py
@@ -6,7 +6,6 @@
 # Web: https://github.com/gboeing/osmnx
 ################################################################################
 
-import demquery
 import math
 import networkx as nx
 import pandas as pd
@@ -16,6 +15,11 @@ import time
 from .core import get_from_cache
 from .core import save_to_cache
 from .utils import log
+
+try:
+    import demquery
+except ImportError:
+    demquery = None
 
 
 def add_node_elevations(G, api_key, max_locations_per_batch=350,
@@ -117,6 +121,9 @@ def add_node_elevations_from_dem(G, dem_paths, band=1, interp_kind=None):
     -------
     G : networkx multidigraph
     """
+    if demquery is None:
+        msg = 'The demquery package must be installed to use this optional feature.'
+        raise ImportError(msg)
 
     nodes = [(node, data['x'], data['y']) for node, data in G.nodes(data=True)]
     df = pd.DataFrame(
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
-demquery>=0.2.0
 descartes>=1.1
 geopandas>=0.5
 matplotlib>=2.2
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+demquery>=0.2.0
 descartes>=1.1
 geopandas>=0.5
 matplotlib>=2.2
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,6 @@
 coverage
 coveralls
+demquery>=0.2.0
 folium
 pytest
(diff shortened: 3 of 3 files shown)


gboeing:
`demquery` needs to be live on conda-forge for the CI to run successfully

gboeing:
> Do you have plans to deprecate Python 2.7 and 3.4? Python 3.4 is already end of life, and 2.7 is in a month. I made demquery 3.5+ compatible, but haven't tested Python 2.7.

OSMnx is dropping official py2.x and py3.4 support in the next release.

osmnx/elevation.py
```diff
@@ -6,6 +6,7 @@
 # Web: https://github.com/gboeing/osmnx
 ################################################################################
 
+import demquery
```
gboeing: This should be handled as an optional import. For example see how scikit-learn components are imported here: https://github.com/gboeing/osmnx/blob/bde270d4a148d8f8911aef59c86a099adc76c24f/osmnx/geo_utils.py#L39-L42
And then checked/handled here: https://github.com/gboeing/osmnx/blob/bde270d4a148d8f8911aef59c86a099adc76c24f/osmnx/geo_utils.py#L427-L428

requirements.txt
```diff
@@ -1,3 +1,4 @@
+demquery>=0.2.0
```
gboeing: Place this dependency in `requirements-dev.txt` instead, as this is an optional dependency for non-core functionality (see previous comment above).

osmnx/elevation.py
kylebarron: Ah makes sense. I wasn't sure how to do optional imports but I agree it should be one

kylebarron:
> > Do you have plans to deprecate Python 2.7 and 3.4? Python 3.4 is already end of life, and 2.7 is in a month. I made demquery 3.5+ compatible, but haven't tested Python 2.7.
> 
> OSMnx is dropping official py2.x and py3.4 support in the next release.

Great. I just asked because as of now [`setup.py`](https://github.com/gboeing/osmnx/blob/master/setup.py) still includes 2.x and 3.4.

kylebarron:
I tried uploading to conda-forge but I'm getting uninformative build errors https://github.com/conda-forge/staged-recipes/pull/10355. I'll have to take another stab at this tomorrow.

kylebarron:
Checks now pass on my [conda-forge recipe](https://github.com/conda-forge/staged-recipes/pull/10355), so ideally it'll be on conda-forge soon.

I'll open a new PR for adding elevations to edges?

kylebarron:
Just toggling to re-run travis

kylebarron:
Ok now the feedstock is rebuilt and the [package page is live](https://anaconda.org/conda-forge/demquery) so I'll toggle it once more.

kylebarron:
Do you want a test on this, so that the coverage percentage doesn't go down? I'm not sure what the easiest way to test is? Find a lightweight DEM and then download it every time?

gboeing:
@kylebarron what's the status on this PR? Were you able to identify a lightweight DEM file and incorporate it into the tests to cover the new functionality?

gboeing:
Closed for inactivity. Can be reopened in the future as needed.

kylebarron:
> Were you able to identify a lightweight DEM file and incorporate it into the tests to cover the new functionality?

I never got a reply to my questions, so I never did anything further. 

Small DEM files are easy to find, e.g. `https://raw.githubusercontent.com/visgl/deck.gl-data/master/website/terrain.png` of SF:


I think adding a test should be pretty simple. I guess the question is how to have a reproducible network of SF in the test. I haven't looked into your test code recently.

janiksudo:
I would like to revive this issue(and help with it) , and add DEM files as another source of elevation next to Google Maps and Airmap.

What needs to be done to conclude your work @kylebarron, and what needs to be done in your opinion @gboeing - I guess tests to cover the new functionality, anything else?

kylebarron:
As of my last commit to this branch, it was working correctly and just needed test cases. I haven't kept up with any changes in `osmnx` since this PR

gboeing:
@janikmu been a while since I looked through this PR, but off the top of my head:

  1. lightweight but thorough tests
  2. double-checking functionality with rest of codebase as there have been some changes in recent months
  3. some sort of ongoing maintenance plan

The first two are straightforward, but I'm unsure of the latter, which is increasingly important as OSMnx accumulates features beyond its core functionality (and which I do not have personal use cases for). For example, this PR adds `demquery` as an optional dependency, but it looks like this package is [no longer](https://github.com/conda-forge/demquery-feedstock/pull/1) being kept up to date on conda-forge as it has a stale auto-PR from January.

This is not meant as a swipe against @kylebarron's `demquery` project in any way: it looks like a useful tool! But, to add a dependency, we also need to have some assurance of near-permanent support for the dependency package itself as well as its conda-forge distribution. The reason is simple: I'm the primary maintainer and nearly sole tech support, and I separately already work 70+ hours a week. When a new feature is contributed, and the contributor eventually drifts away from using OSMnx, I'm on the hook for any additional marginal ongoing maintenance of the codebase and keeping an eye on any dependencies that don't have lots and lots of contributors and ongoing commits.

So the short story is: I'm open to this PR, but I also want you to be aware of these caveats regarding ongoing maintenance and time burdens. I'm open to any comments that may put them at rest.

kylebarron:
Thanks, I must've missed the conda-forge PR before; I previously only published to pip, and conda-forge packaging is much more difficult than publishing to pypi.

In any case, the internals of `demquery` is a simple, well-documented [300-line script](https://github.com/kylebarron/demquery/blob/master/demquery/demquery.py). @janikmu could copy the file if you didn't want to link to an external dependency.

janikmu:
Good comment, @gboeing ! I really did not see it from this side. Maybe it then would be better to not bloat osmnx with more non-core functionality and do elevation from DEM files - > networkx somewhere else. (As elevation really is only another node attribute in the networkx graph, and can be added after the network is created by osmnx).

Somewhere else could mean at the user-level (I, for my use-case can do it following the code from @kylebarron ) or maybe I will even write a module for it if I have the time.

On the other hand, if `demquery` is a no-brainer dependency that needs no maintenance, this is your call @gboeing to decide, if it is outside of core functionality or could be helpful.

gboeing:
Thanks all for the comments. At this point I'm inclined to leave DEM elevation out of OSMnx itself. It looks like it would be relatively straightforward for a user to use demquery to add node elevations after building a graph.

If you do write a short general script adapting the code here to load graph node elevations using demquery, it'd be awesome to include it as an example at https://github.com/gboeing/osmnx-examples for others to reuse. That could be a great option, if you're so inclined.

kylebarron:
> If you do write a short general script adapting the code here to load graph node elevations using demquery, it'd be awesome to include it as an example at [gboeing/osmnx-examples](https://github.com/gboeing/osmnx-examples) for others to reuse. 

I think that's a great suggestion if @janikmu is interested. You can copy the function from this PR into a notebook and provide it as an example for adding elevation nodes.

PR #359, opened 2019-12-18
fix duplicate nodes in OSM XML output (fixes #356)

### LINKED ISSUE #356: Fix duplicate edges in save_graph_osm()
When I originally wrote this function, I didn't realize that the edges dataframe generated by `ox.graph_to_gdfs()` will contain multiple rows with the same `osmid` for any way that is associate with more than two nodes. As such, save_graph_osm() will generate XML like
```
<way id="123456789"/>
  <nd ref="1"/>
  <nd ref="2"/>
  <tag k="highway" v="residential"/>
</way>
<way id="123456789"/>
  <nd ref="2"/>
  <nd ref="3"/>
 <tag k="highway" v="residential"/>
</way>
<way id="123456789"/>
  <nd ref="3"/>
  <nd ref="4"/>
  <tag k="highway" v="residential"/> 
</way>
```
instead of 
```
<way id="123456789"/>
  <nd ref="1"/>
  <nd ref="2"/>
  <nd ref="3"/>
  <nd ref="4"/>
  <tag k="highway" v="residential"/>
</way>
```
The XML will still validate, as was documented in the original PR, but it doesn't technically conform to the OSM XML design schema, nor is it the most efficient way of storing the data. Furthermore I'm just not sure what this might do for any downstream processes (e.g. routers) that make use of this re-formatted file. 

In practice, this should be a straightforward fix, edit [this](https://github.com/gboeing/osmnx/blob/master/osmnx/save_load.py#L217-L224) chunk of code to check whether or not the tree already contains an edge with that id and if so just add the new nodes to it rather than creating a whole new way. However, according to the OSM wiki, OSM XML requires that the `<nd/>` tags be listed "in the correct order" (see the last sentence [here](https://wiki.openstreetmap.org/wiki/Way#Street_as_a_vector), and I'm unsure if its possible to infer that order from the order in which the duplicate `osmid` ways appear in the edges table.. 

So, my questions are two-fold:
1. Is this a big enough problem to worry about fixing?
2. If so, is it possible to recover the correct node order from either the edges table or the original NetworkX graph object?

mxndrwgrdnr:
Addresses issue #356. Bulk of the changes are in utils.py where I put the logic to extract the node order from the edges. The docstrings of that function (`utils.get_unique_nodes_ordered_from_way`) highlight the one major caveat of this approach, namely that if all of the nodes associated with a given way can't be connected based on the edges table, then the algorithm will simply take the first continuous component and throw out the rest. It's unclear to me if the OSM XML schema allows for such a structure, but I couldn't find anywhere where its explicitly forbidden, so I had to account for this edge case. In my testing I haven't found any examples of this yet, but that doesn't mean they aren't out there, and I needed to be careful with the while loop. Thoughts?


osmnx/save_load.py
```diff
@@ -169,11 +169,14 @@ def save_graph_osm(G, node_tags=settings.osm_xml_node_tags,
     if folder is None:
         folder = settings.data_folder
 
+    # get undirected graph so we don't generate duplicate nodes
+    H = get_undirected(G)
+
     # create a copy to convert all the node/edge attribute values to string
-    G_save = G.copy()
+    H_save = H.copy()
```
gboeing: Might be more efficient to say:
```suggestion
    H_save = get_undirected(G).copy()
```
instead of creating that `H` variable if it's unused elsewhere.

osmnx/utils.py
```diff
@@ -198,6 +198,91 @@ def euclidean_dist_vec(y1, x1, y2, x2):
     return distance
 
 
+def get_unique_nodes_ordered_from_way(way_edges_df):
```
gboeing: Does this function get tested during the CI tests? It looks like test coverage drops significantly in this PR.

osmnx/save_load.py
mxndrwgrdnr: done

osmnx/utils.py
mxndrwgrdnr: done

mxndrwgrdnr:
Let's hold off on the merge until we address the latest issue in #356.

tests/test_osmnx.py
```diff
@@ -6,6 +6,7 @@
 
 import matplotlib as mpl
 import warnings
+import pandas as pd
```
gboeing: Can you alphabetize the new import?

osmnx/utils.py
```diff
@@ -198,6 +203,92 @@ def euclidean_dist_vec(y1, x1, y2, x2):
...
```
gboeing: Can you move this new function to geo_utils module? Since the refactor, that module is now holding all the network/spatial utility functions.

mxndrwgrdnr: done.

tests/test_osmnx.py
mxndrwgrdnr: done.

gboeing:
Looks like there's a conflict that still needs to be resolved.

mxndrwgrdnr:
OK looks like the new structure is causing a circular import error. ox.save_load.py now imports from ox.geo_utils.py to initialize the `get_unique_nodes_ordered_from_way()` method, but ox.geo_utils.py  is already importing from from ox.save_load.py (`graph_to_gdfs()`). Any other thoughts about how you'd like the files structured?

gboeing:
Ah, sorry. For now maybe it's best to just leave your helper function in utils.py. And I can just go through everything more thoughtfully next time there's a refactor.

mxndrwgrdnr:
OK cool. Looks like we're all set here then.

gboeing:
Thanks!

PR #364, opened 2020-01-04
Fix improper geometry assignment

martinfleis:
Patch for https://github.com/geopandas/geopandas/issues/1249. In the geopandas master, geometry column assigned after the creation of gdf is not properly recognised and `unary_union` does not work. `set_geometry` fixes the issue.

Hopefully this will be fixed in geopandas before the next release, but there is no harm ensuring that gdf is created correctly in OSMnx.

As a side note, I would recommend having one Travis env with master of geopandas and networkx (+ maybe other) to capture this kind of issues ahead of release. E.g. Geopandas is changing the way it stores CRS [which will require another minor change in OSMnx](https://github.com/geopandas/geopandas/pull/1101#issuecomment-570727439).


gboeing:
Thanks @martinfleis!

PR #368, opened 2020-01-11
Installation instructions enhanced

NJannasch:
Installation is now documented for usage without conda or docker containers.


gboeing:
It's more complicated than that, which is partly why I leave if off the readme. There are other dependencies, some of which have nontrivial installation. And those instructions won't work on all platforms. Hence the more generic instructions: they cover 99% of users and the remaining power users who want to do a pip-only install can double-check their platform-specific requirements for the dependencies, which can vary widely.

PR #369, opened 2020-01-11
rename save osm method and fix colname bug

mxndrwgrdnr:
One last set of changes to the osm output method:

1. Refactored it so it can accept either a graph object or an iterable of nodes/edges geodataframes. Renamed the method accordingly.
2. fixed a bug that caused the edges gdf to have multiple 'id' columns.


mxndrwgrdnr:
sorry I didn't catch this in the last PR @gboeing

Issue #373, opened 2020-01-31
OSM Tags for oneway roads


whereiszoe:
Thank you very much for the awesome package. I am new to GitHub, please bear with me if I raise the issue at the wrong section. 

In the OSM dataset I am working with (i.e., Ho Chi Minh city in Vietnam), the Tag values for oneway are  "F", "T", or "B", rather than "1", "-1", "yes".
I also found a document on interpreting OpenStreetMap Data (https://download.geofabrik.de/osm-data-in-gis-formats-free.pdf), which indicates that:
“F” means that only driving in direction of the line string is allowed. 
“T” means that only the opposite direction is allowed. 
“B” (default value) means that both directions are ok.
 
I wonder if the code should be modified so that it allows for the consideration of such values? Possible changes to the list of "osm_oneway_values" and the condition "if data['oneway'] == '-1':" in "add_paths(G, paths, bidirectional=False)"? Thanks for your answer in advance.

The relevant source code is pasted as below.
 
```python
def add_paths(G, paths, bidirectional=False):
    """
    Add a collection of paths to the graph.
    Parameters
    ----------
    G : networkx multidigraph
    paths : dict
        the paths from OSM
    bidirectional : bool
        if True, create bidirectional edges for one-way streets
    Returns
    -------
    None
    """

    # the list of values OSM uses in its 'oneway' tag to denote True
    osm_oneway_values = ['yes', 'true', '1', '-1']

    for data in paths.values():

        if settings.all_oneway is True:
            add_path(G, data, one_way=True)
        # if this path is tagged as one-way and if it is not a walking network,
        # then we'll add the path in one direction only
        elif ('oneway' in data and data['oneway'] in osm_oneway_values) and not bidirectional:
            if data['oneway'] == '-1':
                # paths with a one-way value of -1 are one-way, but in the
                # reverse direction of the nodes' order, see osm documentation
                data['nodes'] = list(reversed(data['nodes']))
            # add this path (in only one direction) to the graph
            add_path(G, data, one_way=True)

```

gboeing:
Thanks. Would you like to open a PR?

whereiszoe:
Sorry I think I accidentally clicked the "commit directly to the master branch". Please ignore it. This is my first PR on GitHub...

gboeing:
closed by #378

PR #378, opened 2020-02-05
OSM Tags for oneway roads

whereiszoe:
This PR is meant to answer the issue #373 that raised by myself. I implemented some changes in the core.py in case there are other users need this functionality without further checking the one-way attribute values in OSM dataset. Thanks for your feedback.


gboeing:
thanks!

Issue #384, opened 2020-02-11
Feature Request: Retrieve OSM Data for Specific Date


kanaad:
First off, osmnx is incredibe and has made a lot my work a lot easier, so thank you to all who have contributed!

Recently, San Francisco closed Market Street, and now retrieving the the 'drive' network type does not include those edges - since the data I'm using in conjunction with the map is from last year, this presents a problem.

I'm curious as to how possible it is to specify a date when retrieving OSM data, as just another argument in, for example, the graph_from methods.

gboeing:
Thanks for the suggestion! I'm not sure how trivial or complicated this would be. If you know more about it, please share what would be involved or some sample code snippets and we can discuss how it fits with the OSMnx project here.

kanaad:
I should mention that I have very little experience using the Overpass API. However, it seems that there is a data parameter that can be passed into the API, seen [here](https://wiki.openstreetmap.org/wiki/Overpass_API/Overpass_API_by_Example#OSM_data_at_a_certain_date).

Is it possible that a date can be added to the query string passed to `overpass_request`?

pmartincalvo:
Dropping by to add this feature would be very useful. I'm not sure if any work has been done on this, would be willing the lend a helping hand.

gboeing:
If someone wants to create a simple PR for this, I'm open to reviewing it.

pmartincalvo:
Will look into it.

gboeing:
closed in #405

sunayana:
@gboeing : Could you say, when the date is given in the config, is the data extracted from OSM till that date or only on that date. 
Thanks!

gboeing:
Hi @sunayana, per the contributing guidelines, best to ask "how-to" usage questions on StackOverflow as the issue tracker is reserved for fixing bugs and proposing new features.

PR #405, opened 2020-03-20
Allow flexible overpass settings

pmartincalvo:
Until now, overpass QL [settings](https://wiki.openstreetmap.org/wiki/Overpass_API/Overpass_QL#Settings) (the first statement in an overpass query) were mostly hardcoded, with only maxsize and timeout being open to change from the high level functions of OSMnx.

In issue https://github.com/gboeing/osmnx/issues/384, the ability to modify the date for the overpass query was requested as a feature. The date in an overpass query is modified in its settings. When I looked into it, I noticed it would be probably wiser to make settings customizable in a similar fashion to the current `custom_filters`, instead of making an hyperspecialized design just for the date.

The proposed change allows the user to feed a customized settings string. If no string is given, the same default values that are now hardcoded will be used, thus not breaking backwards compatibility.

@gboeing , no tests for this yet. I've tried to understand the testing philosophy of OSMnx from the current tests but I haven't seen any testing related to querying Overpass. Could you kindly let me know your thoughts on this?

### DIFF
--- a/osmnx/core.py
+++ b/osmnx/core.py
@@ -190,6 +190,9 @@ def osm_net_download(polygon=None, north=None, south=None, east=None, west=None,
         grids, ie, 'way["power"~"line"]'
     custom_filter : string
         a custom network filter to be used instead of the network_type presets
+    custom_settings : string
+        a custom settings to be used in the overpass query instead of the default
+        ones
 
     Returns
     -------
--- a/osmnx/core.py
+++ b/osmnx/core.py
@@ -1116,7 +1116,7 @@ def graph_from_point(center_point, distance=1000, distance_type='bbox',
                      truncate_by_edge=False, name='unnamed', timeout=180,
                      memory=None, max_query_area_size=50*1000*50*1000,
                      clean_periphery=True, infrastructure='way["highway"]',
-                     custom_filter=None):
+                     custom_filter=None, custom_settings=None):
     """
     Create a networkx graph from OSM data within some distance of some (lat,
     lon) center point.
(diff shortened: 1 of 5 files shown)


osmnx/core.py
```diff
@@ -221,6 +224,11 @@ def osm_net_download(polygon=None, north=None, south=None, east=None, west=None,
     else:
         maxsize = '[maxsize:{}]'.format(memory)
 
+    # use custom settings if delivered, otherwise just the default ones.
+    overpass_settings = '[out:json][timeout:{timeout}]{maxsize}'.format(timeout=timeout, maxsize=maxsize)
```
gboeing: this should probably be converted to an if/else pattern rather than an if/overwrite pattern

pmartincalvo: Done.

pmartincalvo:
Makes sense. I'll try to chase down all occurrences and apply the same philosophy.

On a side note, right now the default value for the settings is getting hardcoded in the function. If you approve, I would like to move it to the settings file and import it where necessary so there's a single point for modifying if necessary. 

Will wait for your confirmation on that.

gboeing:
> On a side note, right now the default value for the settings is getting hardcoded in the function. If you approve, I would like to move it to the settings file and import it where necessary so there's a single point for modifying if necessary.

I think that would make sense for usability and consistency with the rest of the package.

pmartincalvo:
Footprints and POIs now also allow for custom settings. I have also set the default setting string in settings.py as discussed.

Unless you think the same needs to be applied somewhere else, I would say only testing is pending.

pmartincalvo:
Tests done. Do let me know if any other changes are required.

gboeing:
@pmartincalvo thanks!

gboeing:
closes #384

gboeing:
Note that this `custom_settings` parameter has been deprecated. It is instead now easily accessible via the ox.config() function and settings module as `overpass_settings`. See #492.

PR #413, opened 2020-04-02
save graph to geopackage

### LINKED ISSUE #411: Saving graphs to GeoPackages or PostgreSQL (PostGIS) tables?
hello,

I wonder if it would be possible to export a graph (within the `save_load` module e.g.) directly as:

- two GeoPackages (nodes/edges)
- two PostGIS tables (nodes/edges)

The ESRI Shapefile has some limitations (e.g. field names get truncated to 10 char).

I would suggest to wait for GeoPandas 0.8 to be out for a direct export to PostGIS as it's already [on the track](https://github.com/geopandas/geopandas/pull/822).

Thanks.

gboeing:
fixes #411

### DIFF
--- a/osmnx/save_load.py
+++ b/osmnx/save_load.py
@@ -32,7 +32,7 @@ def save_gdf_shapefile(gdf, filename=None, folder=None):
     gdf : GeoDataFrame
         the gdf to be saved
     filename : string
-        what to call the shapefile (file extensions are added automatically)
+        the name of the shapefiles (not including file extensions)
     folder : string
         where to save the shapefile, if none, then default folder
 
@@ -49,7 +49,7 @@ def save_gdf_shapefile(gdf, filename=None, folder=None):
 
     # give the save folder a filename subfolder to make the full path to the
     # files
-    folder_path = os.path.join(folder, filename)
+    filepath = os.path.join(folder, filename)
 
     # make everything but geometry column a string
     for col in [c for c in gdf.columns if not c == 'geometry']:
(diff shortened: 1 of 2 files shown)


gboeing:
@swiss-knight this has been merged into the master branch now if you wouldn't mind testing it out

swiss-knight:
### Small review

I have quickly tested to write my downloaded graph which is of type:
```
<class 'networkx.classes.multidigraph.MultiDiGraph'>
G_raw is_directed ?: True
```

to a GeoPackage as well as Shapefiles.

Both operations succeeded.  

## Execution time
Execution times (for the saving operation only) are roughly the same (~40s on my old laptop).

## Size
Sizes ; ~76 MB for both Shapefiles' folders and 8.9 MB for the GeoPackage single file.        
I'm always astonished (!) by the difference in sizes, that's cool.    

## Number of features
In both the nodes and the edges tables, number of features written are the same in case of the Shapefile and the GeoPackage.    

## Fields of the attribute table and geographical position of the features
For the nodes, once opened in QGIS they show the same attributes and every feature is at the same geographical position.     
There are extra `x`, `y` and `fid` columns in the attribute table in case of the GeoPackage, which may be normal I think (?).    

For the edges, once opened in QGIS they show the same attributes (with full non 10-char truncated names in case of the GeoPackage \o/ !) and every feature is at the same geographical position.           
There are extra `u`, `v` and `fid` columns in the attribute table in case of the GeoPackage, which again may be normal I think (?).      

### NULL fields
For the edges, some fields are written "_NULL_" in the attribute table of the Shapefile, but not in the GeoPackage attribute table, they are just left empty.
While for the nodes, features of the only column '_highway_' in both the Shapefile and the GeoPackage are written "_NULL_".  It may be a normal behavior again, I don't know, I just noticed that tiny difference.

## Overwriting existing file
I have tested both writing the GeoPackage from scratch, _and_ overwriting the existing file.    
No differences to notice, everything went fine.    
During the write process, I sometime figured out extra files laying around with fancy names (one ending by `journal`), but they are temporary and go away at the end.    


## Writing methods used
I've tested both;    
`ox.save_graph_geopackage(G, filename=os.path.join(gpkg_dir,'raw_graph.gpkg'), encoding='utf-8')`
and
`ox.save_graph_geopackage(G, folder=gpkg_dir, filename='raw_graph.gpkg'), encoding='utf-8')`

with success.

gboeing:
Thanks!

> There are extra x, y and fid columns in the attribute table in case of the GeoPackage, which may be normal I think (?).

Yes, `fid` is the default name of the primary key in the SQLite database: https://gdal.org/drivers/vector/gpkg.html#layer-creation-options

PR #446, opened 2020-05-06
expose timeout and memory parameters in pois and footprints modules

### LINKED ISSUE #445: Timeout Error from footprints_from_place
Hi,

First of all, thank you for this project which appears really helpful.

I have encountered a timeout error when I want to retrieve buildings plan from a place given : 
ReadTimeout: HTTPConnectionPool(host='overpass-api.de', port=80): Read timed out. (read timeout=180)

my code is :

cs = '[out:json][timeout:600]'
buildings2 = ox.footprints_from_place(ville, footprint_type='building', retain_invalid=False, which_result=1, custom_settings=cs)

Are my settings wrong or is it just not possible to fix the timeout ?

Thank you for your answer.

gboeing:
resolves #445


gboeing:
Note that these params have been deprecated. They are now accessible via the ox.config() function and settings module instead.

PR #449, opened 2020-05-11
Generalize POIs module

### LINKED ISSUE #190: Feature enhancement: Include to POI search other Map features as well
Hi @gboeing, one of my students just pointed out that the poi search is missing many useful poi types such as:

- [Shop](https://wiki.openstreetmap.org/wiki/Map_Features#Shop)
- [Craft](https://wiki.openstreetmap.org/wiki/Map_Features#Craft)
- [Emergency](https://wiki.openstreetmap.org/wiki/Map_Features#Emergency)
- [Historic](https://wiki.openstreetmap.org/wiki/Map_Features#Historic)
- [Leisure](https://wiki.openstreetmap.org/wiki/Map_Features#Leisure)

I can work on this one to get these also included but there are a few different ways of building the logic of including these different types for the search. 

The functions could work in a way that:

1. There is e.g. a parameter for the key words that the user wants to search and then the search would be finding the POIs from all these different sources. This would be the most simple one for the user but it lacks the possibility to exclude some sources for POIs.

2. Second option is that we provide parameters for all of these different sources and user defines lists for all the sources s/he wants to search. E.g. `amenities = ['grocery'], shops=['grocery']` etc. This might be a bit unpractical though.

3. The last one (which is the logic I vote for) would be to use flags for different sources. E.g. the search criteria would be passed with parameter `keys = ['grocery']` and then the user could define with boolean parameters the sources s/he wants to search the POIs from, e.g. `amenities = True, shops=True, emergency=False, leisure=True` etc. Other possibility would be to pass a list of values like `sources = ['amenities', 'shops', 'leisure']`. However, the flags might be more stable way of doing this. 

Do you have any thoughts about this?

Cheers,
Henrikki

gboeing:
Generalizes the POIs module to query using a `tags` dict instead of an `amenities` list. Removes the `amenities` parameter from all POI functions. The `tags` dict accepts key:value pairs of the form:

  1. 'tag' : True (to retrieve all items with `tag`)
  2. 'tag' : 'value' (to retrieve all items with `tag` = `value`)
  3. 'tag' : ['value1', 'value2', etc] (to retrieve all items with `tag` equal to either `value1` or `value2` etc.

Resolves #190. Adapts contributions from stale PRs #240 #342 with enhancements from #446.

Usage examples of the new POI querying functionality:

```python
import osmnx as ox

poly = ox.gdf_from_place('Boston, MA, USA', buffer_dist=2000)
gdf = ox.pois_from_polygon(poly['geometry'].iloc[0], tags={'amenity':'university'})

gdf = ox.pois_from_address(address='Piedmont, California, USA',
                           tags={'amenity' : 'school'},
                           custom_settings='[out:json][timeout:180][date:"2019-10-28T19:20:00Z"]')

gdf = ox.pois_from_point(point=(42.344490, -71.070570),
                         distance=500,
                         tags={'amenity' : 'restaurant'},
                         timeout=200,
                         memory=100000)

tags = {'amenity' : True,
        'landuse' : ['retail', 'commercial'],
        'highway' : 'bus_stop'}
gdf = ox.pois_from_place(place='Piedmont, California, USA', tags=tags)
```

### DIFF
--- a/osmnx/pois.py
+++ b/osmnx/pois.py
@@ -45,9 +45,11 @@ def create_poi_query(north, south, east, west, tags,
         server memory allocation size for the query, in bytes. If none, server
         will use its default allocation size
     custom_settings : string
-        custom settings to be used in the overpass query instead of the default
-        ones
-
+        custom settings to be used in the overpass query instead of defaults
+    
+    Returns
+    -------
+    query : string
     """
 
     # pass server memory allocation in bytes for the query to the API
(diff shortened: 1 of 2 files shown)


gboeing:
@carlhiggs @shiqin-liu I'd be curious to hear your thoughts on this PR (and to make you aware of it). It enhances and generalizes the OSMnx pois module to handle all points of interest. It was previously limited to amenities only. This may be useful for querying OSM in the global indicators project?

shiqin-liu:
Excited to see this PR! I think it certainly makes OSM POIs query more flexible. I recalled that what I used before is [node_query](https://github.com/UDST/pandana/blob/a187e6d391c4412863dec6bbee363c29e298f294/pandana/loaders/osm.py#L160) from Pandana as it allows search POIs based on tags within a bbox. These functions are what I am looking for, so thanks for making me aware of this, I think it will certainly be very useful for the project moving forward.

carlhiggs:
This is exciting, thanks for letting me know!  I had previously looked into the use of OSMnx for identifying destinations when I first started working with OSM a few years ago, but ended up using osm2pgsql (having a postgis workflow, this made sense for our national project).  It'll be great to have this as an option when working with OSMnx in the future.  It could be possible to refactor code to make use of it for global indicators; its worth considering as we look to simplify code into a unified package.  Nice one!

PR #450, opened 2020-05-12
Topological intersection consolidation

gboeing:
This PR adds functionality to consolidate graph intersections (nodes) and reconstruct the graph topology. It adapts #339 to resolve outstanding issues. See also #249 and #299.

Many real-world street networks feature complex intersections and traffic circles, resulting in a cluster of graph nodes where there is really just one true intersection, as we would think of it in transportation or urban design. Similarly, divided roads are often represented by separate centerline edges: the intersection of two divided roads thus creates 4 nodes, representing where each edge intersects a perpendicular edge, but these 4 nodes represent a single intersection in the real world. Traffic circles similarly create a cluster of nodes where each street's edge intersects the roundabout.

The existing [clean_intersections](https://osmnx.readthedocs.io/en/stable/osmnx.html#osmnx.simplify.clean_intersections) function cleans up these clusters by buffering their points to an arbitrary distance, merging overlapping buffers, and returning a GeoSeries of their centroids. This PR deprecates this function and replaces it with a new `consolidate_intersections` function that accepts a `rebuild_graph` parameter. If `False`, the function operates like `clean_intersections` did, returning a GeoSeries of merged intersections' centroids. If `True`, it hands off the operation to a new `consolidate_intersections_rebuild_graph` function. This new function consolidates intersections comprising clusters of nodes by merging them (within some buffering tolerance) and returning a rebuilt graph (networkx MultiDiGraph) with consolidated intersections and reconnected edge geometries. It follows these steps:

  1. Buffer nodes to provided tolerance then merge overlapping nodes
  2. Attach (spatial join) each node to its cluster of merged nodes
  3. If a cluster's induced subgraph contains multiple components (i.e., it's not connected) move each component to its own cluster (otherwise you may connect nodes together that are not truly connected, e.g., nearby deadends or surface streets with a bridge)
  4. Create a new empty graph
  5. Create a single new node for each cluster of merged nodes
  6. Create an edge in the new graph for each edge in original graph, from cluster to cluster (always add original self-loops, but otherwise only create the edge if we're not connecting the cluster to itself)
  7. For each cluster with more than 1 node merged in it, extend the incident edge geometries to this new node point. Optionally update graph edge lengths.

This gives you a rebuilt graph with consolidated intersections and reconnected edge geometries, like in the figure below. The consolidated graph nodes/edges are black. The old original graph's nodes/edges are red:

sf-circles

Notice how the traffic circles' many nodes are merged into a new single centroid node, with edge geometries extended to connect to it. Similar consolidation occurs at the intersection of the divided road in the upper-right. Code to create graph visualized above:

```python
import osmnx as ox
G = ox.project_graph(ox.graph_from_place('San Francisco, California, USA', network_type='drive'))
G2 = ox.clean_intersections(G, tolerance=10, rebuild_graph=True, dead_ends=True)
ox.save_graph_geopackage(G2)
```

For best results, the tolerance argument should be adjusted to approximately match street design standards in the specific street network, and you should always use a projected graph to work in meaningful and consistent units like meters. You can also specify if you do not want dead-ends returned in the list of cleaned intersections.

### DIFF
--- a/environments/windows/create-environment.bat
+++ b/environments/windows/create-environment.bat
@@ -15,4 +15,4 @@ CALL conda clean --all --yes
 CALL conda env export > environment.yml
 CALL conda list
 CALL jupyter kernelspec list
-CALL python -c "import osmnx; print(osmnx.__version__)"
+CALL ipython -c "import osmnx; print(osmnx.__version__)"
--- a/environments/windows/environment.yml
+++ b/environments/windows/environment.yml
@@ -10,16 +10,16 @@ dependencies:
   - babel=2.8.0=py_0
   - backcall=0.1.0=py_0
   - beautifulsoup4=4.9.0=py38h32f6830_0
-  - bleach=3.1.4=pyh9f0ad1d_0
+  - bleach=3.1.5=pyh9f0ad1d_0
   - blosc=1.18.1=h6538335_0
   - bokeh=2.0.1=py38h32f6830_0
   - boost-cpp=1.72.0=h0caebb8_0
   - bottleneck=1.3.2=py38h4b8e87e_1
-  - branca=0.4.0=py_0
+  - branca=0.4.1=py_0
   - brotlipy=0.7.0=py38h1e8a9f7_1000
   - bzip2=1.0.8=hfa6e2cd_2
   - ca-certificates=2020.4.5.1=hecc5488_0
-  - cartopy=0.17.0=py38h08d5512_1015
+  - cartopy=0.18.0=py38h4990f70_0
   - certifi=2020.4.5.1=py38h32f6830_0
   - cffi=1.14.0=py38ha419a9e_0
   - cfitsio=3.470=he774522_5
@@ -45,7 +45,7 @@ dependencies:
   - entrypoints=0.3=py38h32f6830_1001
   - expat=2.2.9=he025d50_2
   - fiona=1.8.13=py38h41bf4fa_1
-  - folium=0.10.1=py_0
+  - folium=0.11.0=py_0
   - freetype=2.10.1=ha9979f8_0
   - freexl=1.0.5=hd288d7e_1002
   - future=0.18.2=py38h32f6830_1
@@ -65,9 +65,9 @@ dependencies:
   - imagesize=1.2.0=py_0
   - importlib-metadata=1.6.0=py38h32f6830_0
   - importlib_metadata=1.6.0=0
-  - intel-openmp=2020.0=166
+  - intel-openmp=2020.1=216
   - ipykernel=5.2.1=py38h5ca1d4c_0
-  - ipython=7.13.0=py38h32f6830_2
+  - ipython=7.14.0=py38h32f6830_0
   - ipython_genutils=0.2.0=py_1
   - jedi=0.17.0=py38h32f6830_0
   - jinja2=2.11.2=pyh9f0ad1d_0
@@ -77,10 +77,10 @@ dependencies:
   - jsonschema=3.2.0=py38h32f6830_1
   - jupyter_client=6.1.3=py_0
   - jupyter_core=4.6.3=py38h32f6830_1
-  - jupyterlab=2.1.1=py_0
-  - jupyterlab_server=1.1.1=py_0
+  - jupyterlab=2.1.2=py_0
+  - jupyterlab_server=1.1.3=py_0
   - kealib=1.4.13=h3b59ab9_1
-  - keyring=21.1.1=py38h32f6830_2
+  - keyring=21.2.1=py38h32f6830_0
   - kiwisolver=1.2.0=py38heaebd3c_0
   - krb5=1.17.1=hdd46e55_0
   - libblas=3.8.0=15_mkl
@@ -92,7 +92,7 @@ dependencies:
   - libiconv=1.15=hfa6e2cd_1006
   - libkml=1.3.0=h7e985d0_1011
   - liblapack=3.8.0=15_mkl
-  - libnetcdf=4.7.4=nompi_h256d12c_103
+  - libnetcdf=4.7.4=nompi_h256d12c_104
   - libpng=1.6.37=hfe6a214_1
   - libpq=12.2=hd9aa61d_1
   - libsodium=1.0.17=h2fa13f4_0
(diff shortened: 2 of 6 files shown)


gboeing:
The original version in #339 used a set of nested for loops like this:

```python
old2new = dict()
    for node, data in G.nodes(data=True):
        x, y = data['x'], data['y']
        lon, lat = data['lon'], data['lat']
        osm_id = data['osmid']
        for poly_idx, polygon in enumerate(buffered_nodes):
            if polygon.contains(Point(x,y)):
                poly_centroid = polygon.centroid
                old2new[node] = dict(label=poly_idx, x=poly_centroid.x, y=poly_centroid.y)
                break
```

This isn't scalable to large city graphs. Refactoring to use a fast spatial join like this:
```python
node_points = gdf_nodes[['geometry']]

clusters = gpd.GeoDataFrame(geometry=list(buffered_nodes), crs=node_points.crs)
centroids = clusters.centroid
clusters['x'] = centroids.x
clusters['y'] = centroids.y

gdf = gpd.sjoin(node_points, clusters, how='left', op='within')
```

In testing this refactor on the city of San Francisco, we get a 100-200x speed-up compared to the original nested loops.

gboeing:
@carlhiggs @shiqin-liu I'd be curious to hear your thoughts/suggestions on this PR as we are using the (proposed to be deprecated) `clean_intersections` function in some of the global indicators work.

gboeing:
@ppintosilva @debsankha @samuelduchesne @schoeller @psohouenou all of you have either contributed code or expressed interest in the past regarding this functionality to topologically rebuild a graph after consolidating intersections. This PR brings it all together in a finished form. I've completed preliminary testing. If you have time or interest: any thoughts, suggestions, or testing would be appreciated before merging into master. Thanks!

martinfleis:
This looks great! 

Have you also thought of simplifying network to remove dual centrelines? With consolidated intersections, that should be easy to detect as these would both start and end at the same node (+ some other checks to make sure they're close parallel lines). 

The tricky part would be to generate average geometry, but I can imagine some Delaunay triangulation would help.

gboeing:
> Have you also thought of simplifying network to remove dual centrelines? With consolidated intersections, that should be easy to detect as these would both start and end at the same node (+ some other checks to make sure they're close parallel lines).

Yes that's the next logical step. A rough and dirty approximation to remove duplicate centerlines would be:

```python
# G is a nx.MultiDiGraph
H = nx.DiGraph(ox.consolidate_intersections(G, rebuild_graph=True))
```

This would not average their geometries and it would incorrectly delete true parallel edges elsewhere in the graph. The better option would be something like:

  1. Flag true parallel edges in original graph
  2. Consolidate intersections and rebuild graph topologically
  3. Merge new parallel edges not flagged in step 1 into a single edge with averaged geometry

Given some of our similar research interests, you should shoot me an email if this might be a useful collaboration: I could see a methods paper around edge consolidation. I'm currently planning to write a methods paper this summer on intersection consolidation and how it significantly changes network indicators.

shiqin-liu:
> @carlhiggs @shiqin-liu I'd be curious to hear your thoughts/suggestions on this PR as we are using the (proposed to be deprecated) `clean_intersections` function in some of the global indicators work.

This is an interesting aspect to learn! I will be curious to see how this consolidation will change our indicators as well. For the global project, we use clean_intersection directly with standard 12 tolerance set for all cities to produce intersection density indicator, which contributes to a significant part of our final walkability indicators. As I recalled, we had a brief discussion about whether we should adjust the tolerance for each city based on local topology, but we end up using the standard 12m across cities following the Australian cities standard. I did no follow up on this, maybe @carlhiggs could elaborate on this better. 

This also reminds me of a previous note on the relatively high walkability for Baltimore city (vs. Seattle), this in part is due to a very high intersection density (the others are pop density and accessibility to public transit) within the walkability component. So I guess it will be interesting to test out if this change will make any significant difference to the results.

gboeing:
@shiqin-liu in terms of intersection density, the new `consolidate_intersections` function with `rebuild_graph=True` may yield somewhat (but not very) different intersection counts/densities compared to deprecated `clean_intersections`.

The difference lies in that `rebuild_graph=False` (i.e., equivalent behavior to the deprecated `clean_intersections` function) just merges buffered node points that overlap. But with `rebuild_graph=True`, the `consolidate_intersections` function checks the topology of the overlapping node buffers before merging them.

This prevents *topologically remote but spatially proximate* nodes from being merged. For example:

  - A street intersection may lie directly below a freeway overpass's intersection with an on-ramp. We would not want to merge these together and reconnnect their edges: they are distinct junctions in the system of roads.
  - In a residential neighborhood, a bollarded street may create a dead-end immediately next to an intersection or traffic circle. We would not want to merge this dead-end with the intersection and connect their edges together.

These examples illustrate (two-dimensional) geometric proximity, but topological remoteness. Accordingly, in some situations we may expect higher intersection counts when using the new `consolidate_intersections` function with `rebuild_graph=True` because it is more cautious with merging in these cases. The trade-off is that it has higher time complexity.

shiqin-liu:
@gboeing Thanks for explaining! This makes much more sense to me. It seems that `consolidate_intersections` with `rebuild_graph=True` should be a better option for the global project as it takes into account the local topological complexity. It will be great to test this out when we plan to rerun indicator analysis at some point.

samuelduchesne:
Hi all, sorry for the late reply; I followed this development silently ;). 

In my line of work, we are identifying routes for future district energy systems (that are pipes buried underground) by assigning building energy demands to their closest adjacent streets. Streets with more demand should be focused on. This is why I implemented `nearest_edges` #234 a while back. That's the short story.

Although having topologically accurate street graphs is crucial for street network analysis, they are somewhat too “detailed” for district energy network analysis. The best example is the double centerline edges of divided roads (as mentioned by @gboeing earlier) which presents two possibilities for a building to connect to (on either side of the street). There have been attempts at simplifying such topologies through “skeletonization” of the street network (see the work of Johannes Dorfner [here](https://mediatum.ub.tum.de/doc/1285570/1285570.pdf)).

I am curious to know if the topological intersection consolidation that is discussed here can help us achieve the simplest graph. Does rebuilding the graph (`rebuild_graph`=True) get rid of centerline edges of divided roads?

gboeing:
@samuelduchesne 
 >  Does rebuilding the graph (rebuild_graph=True) get rid of centerline edges of divided roads?

Currently no, but @martinfleis and I are having a side conversation about implementing a feature to consolidate such edges.

carlhiggs:
> > @carlhiggs @shiqin-liu I'd be curious to hear your thoughts/suggestions on this PR as we are using the (proposed to be deprecated) `clean_intersections` function in some of the global indicators work.
> 
> This is an interesting aspect to learn! I will be curious to see how this consolidation will change our indicators as well. For the global project, we use clean_intersection directly with standard 12 tolerance set for all cities to produce intersection density indicator, which contributes to a significant part of our final walkability indicators. As I recalled, we had a brief discussion about whether we should adjust the tolerance for each city based on local topology, but we end up using the standard 12m across cities following the Australian cities standard. I did no follow up on this, maybe @carlhiggs could elaborate on this better.
> 
> This also reminds me of a previous note on the relatively high walkability for Baltimore city (vs. Seattle), this in part is due to a very high intersection density (the others are pop density and accessibility to public transit) within the walkability component. So I guess it will be interesting to test out if this change will make any significant difference to the results.

Apologies for very late response here --- this looks great, and I'll work this into the code for the global project; it will be interesting to compare the new results with old.  Thanks @gboeing and @shiqin-liu!

PR #451, opened 2020-05-14
Refactor package

gboeing:
This PR represents a major refactor to do the following:

  1. Refactor all modules
  2. Refactor tests to make them run 3-4x faster
  3. Remove elapsed time logging and use timestamps in logging instead
  4. Remove make_str function

The make_str was historically used to handle string encoding consistently between python 2.x and python 3.x. Python 2 support was dropped last year, so we can remove make_str from the codebase.


gboeing:
See additional refactoring in #459 and #462

PR #457, opened 2020-05-15
Update how OSMnx handles filepaths when saving graphs to disk + refactor save_load

gboeing:
This PR:

  - updates all graph saving functions to take a filepath argument rather than the old clumsy folder + filename arguments
  - saves shapefiles in a single folder containing both nodes and edges subfiles
  - renames old `save_as_osm` function -> `save_graph_osm` to make it consistent with other saving functions
  - refactors `save_graph_shapefile` function to `use graph_to_gdfs` function
  - refactors `get_undirected` its and related functions out of save_load module and into utils_graph module
  - adds a new api.py module to handle what the package exposes via the init file (still exposes everything, as before)


gboeing:
cc @mxndrwgrdnr this PR and 5ad0dc6 rename old `save_as_osm` function -> `save_graph_xml` to make it consistent with other saving functions.

PR #490, opened 2020-06-04
fix config function behavior

### LINKED ISSUE #489: config function misbehaving
Calling `ox.config()` lets you set the values of the global settings package-wide and resets any un-passed settings to their original default values. Per its docstring:
```
Configure OSMnx by setting the default global settings' values.

Any parameters not passed by the caller are set to their default values.
```

But I noticed there is one odd exception to this for the `useful_tags_node` and `useful_tags_path` settings, which determine the OSM tags to add as graph node/edge attributes respectively. These settings are initially set the same way as the others, but then immediately overwritten with a superset of values:

https://github.com/gboeing/osmnx/blob/59f6a8d61e69cda6e5d9112fbdbaf3d4648e7b23/osmnx/utils.py#L182-L185

This results in weird behavior like this:

```python
import osmnx as ox
print(ox.settings.useful_tags_node)
# ['ref', 'highway']

ox.config(use_cache=True)
print(ox.settings.useful_tags_node)
# ['version', 'timestamp', 'uid', 'id', 'changeset', 'lon', 'user', 'highway', 'ref', 'lat']
```

Using `ox.config` to turn on caching shouldn't change other settings from their default values. We need to prevent this, while offering a straightforward way for users looking to export to XML (who want to add those additional OSM tags/attributes during graph creation).

@mxndrwgrdnr this is likely relevant to your interests.

gboeing:
resolves #489 

Also renames the confusingly-named setting `useful_tags_path` to the more appropriate `useful_tags_way`.

If you're creating a graph to save to .osm XML file later, you can first configure like:

```python
import osmnx as ox
utn = list(set(ox.settings.useful_tags_node + ox.settings.osm_xml_node_attrs + ox.settings.osm_xml_node_tags))
utw = list(set(ox.settings.useful_tags_way + ox.settings.osm_xml_way_attrs + ox.settings.osm_xml_way_tags))
ox.config(use_cache=True, all_oneway=True, useful_tags_node=utn, useful_tags_way=utw)
```

### DIFF
--- a/osmnx/graph.py
+++ b/osmnx/graph.py
@@ -778,7 +778,7 @@ def _convert_path(element):
     path["nodes"] = [group[0] for group in grouped_list]
 
     if "tags" in element:
-        for useful_tag in settings.useful_tags_path:
+        for useful_tag in settings.useful_tags_way:
             if useful_tag in element["tags"]:
                 path[useful_tag] = element["tags"][useful_tag]
     return path
--- a/osmnx/settings.py
+++ b/osmnx/settings.py
@@ -21,7 +21,7 @@ log_filename = "osmnx"
 # useful osm tags - note that load_graphml expects a consistent set of tag names
 # for parsing
 useful_tags_node = ["ref", "highway"]
-useful_tags_path = [
+useful_tags_way = [
     "bridge",
     "tunnel",
     "oneway",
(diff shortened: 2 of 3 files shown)


mxndrwgrdnr:
much cleaner!

PR #493, opened 2020-06-05
Enable simplification of long (rural) roads

pinselimo:
The recursive nature of ```simplification._build_path``` is reduced to only those cases where either a self loop is imminent or the road forks. Path building of long roads is done in a while-loop that substitutes the prior if-statement checking for an endpoint.

So the logics stay the same except for the simple and usual case of a single successor (a road that neither forks nor loops).

#### Memory
In addition, ```_get_paths_to_simplify``` has been turned into a generator in order to reduce the memory footprint.

#### Note
To my understanding the ```RecursionError``` should now be extremely rare. After a recursive call self-loops are filtered by ```successor not in path```. On the other hand a road would need to have an immense amount of forks not classified as endpoints. Of course, the safe way to go is to keep checking for the error.

### DIFF
--- a/osmnx/simplification.py
+++ b/osmnx/simplification.py
@@ -113,10 +113,19 @@ def _build_path(G, node, endpoints, path):
             # if this successor is already in the path, ignore it, otherwise add
             # it to the path
             path.append(successor)
-            if successor not in endpoints:
-                # if this successor is not an endpoint, recursively call
-                # build_path until you find an endpoint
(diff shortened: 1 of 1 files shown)


gboeing:
@pinselimo thanks! Looks good, but the recursive logic is always hard for me to unravel in my mind. I'm going to merge this into a feature branch to do some local comparative testing in the morning.

PR #494, opened 2020-06-05
Enable simplification of long (rural) roads 

### LINKED ISSUE #488: Proposal: Enable simplification of long (rural) roads
Hi!

Downloading a rural, mountainous and rather big area, I've run into the issue that some roads are not simplified properly. This is well documented and happens due to a ```RecursionError``` in ```simplification._build_path```. This error is subsequently caught in line 171 (commit 0ef7002) and logged.

I've had this issue monkey patched so far. Now I forked osmnx and implemented it correctly. It is tested and ready for a PR, However, I opened the issue in order to conform to the contributing guidelines.

#### Further changes for large networks and compatibility:

In a separate commit I've turned ```simplification._get_paths_to_simplify``` into a generator to keep its memory footprint in large networks lower. Also at some point I found that it raises errors with ```Python<3.5``` so I've exchanged ```RecursionError``` for ```RuntimeError```, where the prior is a subtype of the latter, so behaviour remains the same.

#### Testing

All of the above mentioned commits on my branch have been tested according to the contributing guidelines.

cheers,
Simon

gboeing:
Enhancement contributed by @pinselimo in #493. Merged into this feature branch for more testing before merging into master. Resolves #488.

This PR:

  - removes recursion from the simplify_graph path building, allowing it to correctly simplify very long road segments (previously, this exceeded the system recursion depth and left such segments unsimplified)
  - reduces memory footprint by using a generator
  - removes disconnected, self-contained rings by default when simplifying graph (and adds parameter `remove_rings` to optionally turn this off)
  - changes "multiple edges found" log warning to info level instead, as this is not uncommon (happens when OSM has duplicate ways digitized for a single street)

### DIFF
--- a/osmnx/simplification.py
+++ b/osmnx/simplification.py
@@ -227,10 +227,8 @@ def simplify_graph(G, strict=True, remove_rings=True):
     all_nodes_to_remove = []
     all_edges_to_add = []
 
-    # construct a list of all the paths that need to be simplified
-    paths = _get_paths_to_simplify(G, strict=strict)
-
-    for path in paths:
+    # generate each path that needs to be simplified
+    for path in _get_paths_to_simplify(G, strict=strict):
 
         # add the interstitial edges we're removing to a list so we can retain
         # their spatial geometry
--- a/osmnx/simplification.py
+++ b/osmnx/simplification.py
@@ -193,7 +193,7 @@ def _is_simplified(G):
     return "simplified" in G.graph and G.graph["simplified"]
 
 
-def simplify_graph(G, strict=True, rings=False):
+def simplify_graph(G, strict=True, remove_rings=True):
     """
     Simplify a graph's topology by removing interstitial nodes.
 
@@ -211,8 +211,8 @@ def simplify_graph(G, strict=True, rings=False):
         rules but have incident edges with different OSM IDs. Lets you keep
         nodes at elbow two-way intersections, but sometimes individual blocks
         have multiple OSM IDs within them too.
-    rings : bool
-        if False, remove isolated self-contained rings without any endpoints
+    remove_rings : bool
+        if True, remove isolated self-contained rings that have no endpoints
 
     Returns
     -------
@@ -291,7 +291,7 @@ def simplify_graph(G, strict=True, rings=False):
     # finally remove all the interstitial nodes between the new edges
     G.remove_nodes_from(set(all_nodes_to_remove))
 
-    if not rings:
+    if remove_rings:
         # remove any connected components that form a self-contained ring
         # without any endpoints
         wccs = nx.weakly_connected_components(G)
@@ -301,7 +301,7 @@ def simplify_graph(G, strict=True, rings=False):
                 nodes_in_rings.update(wcc)
         G.remove_nodes_from(nodes_in_rings)
 
-    # mark graph as simplified
+    # mark graph as having been simplified
     G.graph["simplified"] = True
 
     msg = (
--- a/osmnx/simplification.py
+++ b/osmnx/simplification.py
@@ -1,7 +1,5 @@
 """Simplify, correct, and consolidate network topology."""
 
-import logging as lg
-
 import geopandas as gpd
 import networkx as nx
 from shapely.geometry import LineString
--- a/osmnx/simplification.py
+++ b/osmnx/simplification.py
@@ -1,5 +1,7 @@
 """Simplify, correct, and consolidate network topology."""
 
+import logging as lg
+
 import geopandas as gpd
 import networkx as nx
 from shapely.geometry import LineString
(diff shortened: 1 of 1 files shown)


gboeing:
@pinselimo do you have a good example of a query that includes a long rural road, for testing?

pinselimo:
Hi Goeff,

you may want to use a custom filter to avoid smaller roads breaking the recursive calls. This is the one that gives me quick results:

~~~python
CUSTOM_FILTER = ('["area"!~"yes"] "highway"!~"cycleway|footway|path|pedestrian|steps|track|corridor|'
                        'proposed|construction|bridleway|abandoned|platform|raceway|service|'
                        'tertiary|unclassified|residential|living_street"]'
                        '["motor_vehicle"!~"no"]["motorcar"!~"no"]["access"!~"private"]'
                        '["service"!~"parking|parking_aisle|driveway|private|emergency_access"]')
~~~

Off the top of my head, I've tried "Lesachtal" and it works. If you want to include areas of both shorter and longer roads try "Carinthia" which includes the prior and three other long roads. This is a bigger area though, took me about a minute or two for the download.

gboeing:
Here are some results from testing simplification between the master branch and the new simplify branch.

I tested with 3 rural, mountainous counties in Northern California + LA county. If I set the system recursion limit to 3000, both code branches perform the same (though I did not do any memory usage profiling). Their runtimes are approximately the same and they produce identical graphs after simplification.

If I set the system recursion limit to 500, branch performance diverges. The simplify branch correctly simplifies the graph topology even in rural areas with 1000+ nodes between true intersections. The master branch (as is known) gives up in such circumstances due to exceeding its max recursion depth, resulting in an only partially simplified graph.

Code for testing:

```python
import osmnx as ox
import sys
import time
ox.config(use_cache=True, log_console=True, log_file=True)

print(sys.getrecursionlimit())
sys.setrecursionlimit(3000) #or 500
print(sys.getrecursionlimit())

# the places to get and the filter to use
places = {'delnorte': 'Del Norte County, CA, USA',
          'trinity': 'Trinity County, CA, USA',
          'lassen': 'Lassen County, CA, USA',
          'la': 'Los Angeles County, CA, USA'}
cf = '["highway"~"motorway|motorway_link|trunk|trunk_link|primary|secondary|tertiary"]'
# also try  '["highway"~"motorway|motorway_link|trunk|trunk_link|primary"]'

# make the graphs
graphs = {}
for label, place in places.items():
    G = ox.graph_from_place(place, custom_filter=cf, simplify=False,
                            truncate_by_edge=True, retain_all=True)
    graphs[label] = G

# simplify each graph using each code branch
def simplify_graphs(graphs, branch):
    for label, G in graphs.items():
        st = time.time()
        G2 = ox.simplify_graph(G)
        et = round(time.time() - st, 3)
        print(branch, label, len(G), len(G2), et)

# checkout master, restart kernel, run
%timeit simplify_graphs(graphs, 'master')

# checkout simplify, restart kernel, run
%timeit simplify_graphs(graphs, 'simplify')
```

I also tested this for San Francisco and New York City with `network_type='drive'` and the two code branches performed identically.

gboeing:
@pinselimo see 845e17c851184c10ca62122c95d876c35475cc14

I have tried to streamline your function further by removing the recursion and simplifying some of its branching logic. I've done some testing, and it yield the same results, but a little simpler and faster.

Can you inspect it (and test it on some real-world locations) and see if you agree?

pinselimo:
@gboeing looks great! At first glance this should really never happen. I will try to verify that later.
Then the if-else block could just check if predecessors is empty, which would be neat.

gboeing:
@pinselimo 

> the if-else block could just check if predecessors is empty

You mean `successors`, right?

gboeing:
@pinselimo ok, my testing has been satisfied at this point. Let me know when yours is done and if it all looked ok. If so, I'll package this into the next release (I'm aiming for Monday afternoon).

gboeing:
I found one small bug that can be reproduced via:

```python
import osmnx as ox
pt = 41.4937980, -87.6881117
G = ox.graph_from_point(pt, dist=500, network_type='drive', retain_all=True)
```

This is rare, but happens if `endpoint_successor` has no successors not already in the path: for example when `endpoint` is [this node](https://www.openstreetmap.org/node/6471322356) and `endpoint_successor` is [this node](https://www.openstreetmap.org/node/2332327023).

This has now been fixed in 24985df

pinselimo:
@gboeing yes I meant successors. I can do extensive testing on monday morning CET. So I can let you know by then.
Interesting bug, I'll also make sure to check it out. (Only have my phone atm)

pinselimo:
I've finished my testing and couldn't find any further problems. Works like a charm :)

gboeing:
Ok, I've tested in on ~8000 study sites over the weekend. Found just one more quirk which can be reproduced using:

```python
import osmnx as ox
point =  41.4134300, -75.6909545
G = ox.graph_from_point(point, network_type='drive')
```

The problem was [this node](https://www.openstreetmap.org/node/96918418), where there is an OSM digitization error: a one-way street turns into a two-way street here, but duplicate incoming one-way edges are present. Thus the node is not considered an endpoint, but you also cannot find new successors because all its other edges (ie, those not already in the path we're building) are incoming only.

This is now fixed and handled in f519d0f.

gboeing:
Ok, all tests pass. I believe it's ready to merge. Thank you!

gboeing:
@pinselimo this has been released in v0.14.1

pinselimo:
@gboeing Thanks! I'm happy to have contributed to this great package.

PR #504, opened 2020-06-09
Remove previously deprecated params, streamline plotting

gboeing:
This PR removes several function parameters that were deprecated with a user warning in a previous release, and replaces some of them. See deprecation details in #502 and #492. It similarly removes the deprecated `plot_shape` function.

It also greatly streamlines the `plot` module, reusing code better among the functions and using kwargs.


gboeing:
Note: OSMnx no longer uses descartes directly, but we need to retain it as a requirement for geopandas to plot polygons.

PR #513, opened 2020-06-26
add airmap as elevation provider

### LINKED ISSUE #512: Add alternative elevation provider
Because Google Maps is not free, and the monthly 200$ budget does not even cover fetching node elevations for one big city, I looked for alternative REST-style APIs that provide elevation data.

Airmap (https://developers.airmap.com/docs/elevation-api) weirdly seems to be free, permanent and without quotas.
I was able to add node elevation for three different network types (walk, bike, drive) for multiple cities without running into limitations. The biggest network was >80.000 nodes and fetching elevation data in batches of 1000 took less than two minutes.

I am happy to prepare a pull request with the minor changes necessary to the elevation module such that one can use something like a `provider=Airmap` function parameter (defaulting to Google Maps?) when calling `add_node_elevations`.

Note though, that the elevation resolution is not as good as with GMaps (elevation is in integer meters). The spatial resolution seems to be similar to SRTM30m datasets. A limitation is that the elevation dataset "is limited to latitudes from 56ºS to 60ºN" (another hint at SRTM30m).
These limitations, however seem like a fair trade-off for everyone that willingly uses the alternative provider.

janiksudo:
This closes #512 

As discussed, `add_node_elevation` parameters remain untouched, everything goes through `settings` and `utils.config`.

### DIFF
--- a/osmnx/elevation.py
+++ b/osmnx/elevation.py
@@ -8,6 +8,7 @@ import pandas as pd
 import requests
 
 from . import downloader
+from . import settings
 from . import utils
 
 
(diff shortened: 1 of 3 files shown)


gboeing:
@janiksudo thanks!

PR #539, opened 2020-07-22
completed refactor of _add_paths

### LINKED ISSUE #538: Refactoring of _add_paths function used to create graph
**Problem being addressed**

The function _add_path was doing multiple tasks making it hard to grasp rapidly. It first evaluated if the path was a one-way given multiple conditions. Then it updated data by reversing it if the path was a one-way and if value of the key in osm path data was either -1 or T. The evaluation of the data being reversed was also within a single condition to avoid hitting a key error raised by the fact that the path data dictionary does not always contain the one-way key for the other conditions. Finally, the function _add_path, used to add a single path, was called after checking every condition if the path was a one-way.

**Proposed solution**

Extracting function _is_path_one_way to evaluate if the path is a one-way, returns a bool. 

Extracting function _is_path_data_reversed to evaluate if the path data is reversed using a try catch to simplify logic, returns a bool.

Reversing path data if both conditions above are true.

Make a single function call in the loop of _add_paths to _add_path passing as the one-way parameter the evaluated condition of _is_path_one_way.

Body of _add_paths after refactor

zacharyDez:
Completed refactor of _add_paths by extracting 2 functions to evaluate if the path is a one way and if the data is reversed. _is_path_one_way and _is_path_data_reversed

Closes #538 

**Changes in PR** 

Extracting function _is_path_one_way to evaluate if the path is a one-way, returns a bool.

Extracting function _is_path_data_reversed to evaluate if the path data is reversed using a try catch to simplify logic, returns a bool.

Reversing path data if both conditions above are true.

Make a single function call in the loop of _add_paths to _add_path passing as the one-way parameter the evaluated condition of _is_path_one_way.

** Code Snippet**

No new functionality added, only a refactor.

code-snippet

### DIFF
--- a/osmnx/graph.py
+++ b/osmnx/graph.py
@@ -588,7 +588,7 @@ def _create_graph(response_jsons, retain_all=False, bidirectional=False):
         G.add_node(node, **data)
 
     # add each osm way (ie, a path of edges) to the graph
-    G = _add_paths(G, paths, bidirectional=bidirectional)
+    _add_paths(G, paths, bidirectional=bidirectional)
 
     # retain only the largest connected component, if caller did not
     # set retain_all=True
@@ -815,8 +815,6 @@ def _add_paths(G, paths, bidirectional=False):
             data["nodes"] = list(reversed(data["nodes"]))
 
         _add_path(G, data, one_way=is_one_way)
-
-    return G
 
 
 class _OSMContentHandler(xml.sax.handler.ContentHandler):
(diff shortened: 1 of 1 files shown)


gboeing:
@zacharyDez this PR has conflicts with the master branch. If you resolve them, I can review then merge.

zacharyDez:
Hi @gboeing, won't get to this in the coming weeks. Looking at the conflict from the github IDE, the conflict is simple to resolve. The code in conflict from master was simply extracted into the method _is_path_one_way. Its exactly the same code within the method. Simply accepting incoming change would suffice. Sorry for the slow reply.

PR #541, opened 2020-07-24
optimize bbox_from_point

### LINKED ISSUE #540: osmnx.utils_geo.bbox_from_point(point, dist=1000, project_utm=False, return_crs=False) too slow for "big" datasets
Hi! 1st of all I have to say this package is amazing and super complete, really useful :)

**"Issue"?**

So I am using it to find Pois in 1km radius from a 55k rows dataset I am working with.
The point is that I wanted to obtain the bounding box from lat/long of all my points (55k). And it took 1h27min with the 
`bbox_from_point(dist=1000, project_utm=False, return_crs=False)` function.

This was my implementation:
`trash_df["bbox"] = trash_df.swifter.apply(lambda column: ox.utils_geo.bbox_from_point((column.latitude, column.longitude), dist=1000), axis=1)`

So I made this fn:

```
def getBbox(lat, lon, radius=1):
    import math
    
    R = 6371

    w = lon - math.degrees(radius/R/math.cos(math.radians(lat)))
    e = lon + math.degrees(radius/R/math.cos(math.radians(lat)))
    s = lat - math.degrees(radius/R)
    n = lat + math.degrees(radius/R)
    
    return (n, s, e, w)
```

And it takes 2seconds and the resulst are equal until 3rd-4th decimal

gboeing:
resolves #540

### DIFF
--- a/osmnx/geocoder.py
+++ b/osmnx/geocoder.py
@@ -73,7 +73,7 @@ def geocode_to_gdf(query, which_result=None, buffer_dist=None):
 
     # if caller passed a list of queries but a scalar which_result value, then
     # turn which_result into a list with same length as query list
-    if isinstance(query, list) and isinstance(which_result, int):
+    if isinstance(query, list) and (isinstance(which_result, int) or which_result is None):
         which_result = [which_result] * len(query)
 
     # turn query and which_result into lists if they're not already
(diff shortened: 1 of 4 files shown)


gboeing:
As a simple test:

```python
import osmnx as ox
dist = 500
point = 37, -117
north, south, east, west = ox.utils_geo.bbox_from_point(point, dist=dist, project_utm=True)
print((north - south) / 2 - dist) #-1.177
print((east - west) / 2 - dist) #0.997
print(ox.utils_geo.bbox_to_poly(north, south, east, west).area) #999636
```

This desired 1 km^2 projected bounding box is within ~1 meter of expected dimensions and has area of 0.9996 km^2.

The new implementation of `ox.utils_geo.bbox_from_point` is several thousand times faster than the old.

PR #542, opened 2020-07-29
New geometries module - Work In Progress

### LINKED ISSUE #478: Are the pois module and footprints module redundant?
Over time, the pois and footprints modules have expanded their functionality and grown more similar to each other. I believe they are approximately redundant now and should be merged together to simplify the codebase.

The following methods are approximately equivalent:

```python
import osmnx as ox
place = 'Emeryville, CA, USA'

# use pois module
po = ox.pois_from_place(place, tags={'building':True})
po.shape #(1117, 73)

# use footprints module
fp = ox.footprints_from_place(place)
fp.shape #(1127, 55)
```

Minor differences would have to be reconciled between these modules (especially the detailed handling of relations in `footprints`) to account for the differences between the resulting GeoDataFrames.

AtelierLibre:
Resolves #478 

Work in progress on a new `geometries` module that combines the capabilities of the existing `pois` and `footprints` modules and introduces conversion of ways to LineStrings or Polygons as appropriate.

Example:

```
# New geometries module
MK = ox.gdf_from_point((52.0430,-0.7590), dist=1000, tags={'highway':True, 'landuse':True, 'building':True})

# Presentation code
MK.to_crs(epsg=27700, inplace=True)

fig,ax = plt.subplots(figsize=(16,16))
MK[MK['landuse'].notna()].plot(ax=ax, column='landuse', cmap='tab20b', legend=True)
MK[MK['highway'].notna()].plot(ax=ax, color='white')
MK[MK['building'].notna()].plot(ax=ax, color='dimgrey', zorder=3)
ax.set_facecolor('darkgrey')
ax.axes.xaxis.set_visible(False)
ax.axes.yaxis.set_visible(False)
```
MK


AtelierLibre:
I had a few issues trying to get the .json read in properly - I hope that's come through okay.

gboeing:
Some quick responses to @AtelierLibre's questions in #478.

> footprints splits large requests into pieces and then processes the separate JSONs received - I don't know how beneficial this is but, assuming it is, wouldn't graph benefit from this as well?

Yes, the `graph` module does this, via the `downloader` module, and `geometries` module should do this too. In fact, the `geometries` module should take advantage of the `downloader` module, including offloading any API interaction code to there.

> so I have preferred to create a single dictionary from the start with a single call to gpd.GeoDataFrame.from_dict(geometries, orient="index") at the end

I think this is the right approach

> I'm having a bit of trouble understanding how the JSON should be included in the package and could use some pointers for that.

Must it be stored as a .json file? What if you create a .py file that contains the same data as a dict instead?

> If there is some other reason for the try/except on the Points and Polygons they can of course go back.

We should plan to do some extended testing when the PR is ready, such as querying geometries across 1,000 different study sites to see if we ever hit anything odd that causes a hiccup.

> I have wondered if the filtering by polygon could be made optional or whether it could be removed if the Overpass query is created more accurately but I haven't investigated this.

This is necessary for consistency. Best strategy here is to follow the existing process of the `graph` module and the `downloader` module.

AtelierLibre:
@gboeing Thanks for the comments

> Yes, the `graph` module does this, via the `downloader` module, and `geometries` module should do this too. In fact, the `geometries` module should take advantage of the `downloader` module, including offloading any API interaction code to there.

Sounds good, I'll take a look at that I'm not familiar with `downloader`

> I think this is the right approach

Thanks, good to get confirmation on that.

> Must it be stored as a .json file? What if you create a .py file that contains the same data as a dict instead?

Yes, this sounds simpler I'll aim for that.

> We should plan to do some extended testing when the PR is ready, such as querying geometries across 1,000 different study sites to see if we ever hit anything odd that causes a hiccup.

Great, I'll be glad to get some help with getting it robustly checked and tested

> This is necessary for consistency. Best strategy here is to follow the existing process of the `graph` module and the `downloader` module.

Okay, I'll review those two

xgerrmann:
@AtelierLibre I have added the functionality gdf_from_bbox, what is the best way to get it into this PR?

xgerrmann:
Im running into some missing geometry here, and I'm currently looking into the cause.
`from osmnx import geometries
import common

city = 'Utrecht'

# Determine bounding box
lat = 52.1
lng = 5.089
print(lat, lng)

width      = common.settings['geo_bounds'][city]['width']
height     = common.settings['geo_bounds'][city]['height']
print(width, height)

width  = 3000
height = width

tags = {'natural':'water'}
#tags = {'waterway':'riverbank'}

gdf = geometries.gdf_from_bbox((lat,lng), (width, height), tags=tags)`

This is what I get:
image

This is what I expect (regarding the water):
image

AtelierLibre:
@xgerrmann I've just had a quick look at this - I think the issue is that some of the waterways are extremely long so their centroids fall outside of the query polygon (the bounding box) and they are removed by [the spatial filter at the end](https://github.com/AtelierLibre/osmnx/blob/502f83a617ca48c39447e812167eca0e4d205399/osmnx/geometries.py#L586). If you delete `.centroid` so that the line looks like `gdf_indices_in_polygon = utils_geo._intersect_index_quadrats(gdf, polygon)` and try it again I think you will get what you expect.

@gboeing I left the `.centroid` in as the `_quadrat_cut_geometry()` function has obviously been written to test for points in a polygon - do you see any problem with using it for other geometries as well?

xgerrmann:
@AtelierLibre Indeed, removing the .centroid does the job. 
Thinking of possible solutions here, I thought of two:
- Also split each geometry with the _quadrat_cut_geometry and check for each sub geometry if it lies within the surrounding polygon. I would say the end result is not really deterministic.
- Split each polygon up into points and then check for each point if it is in the encompassing polygon.
- For each element with a centroid outside the bounding box (most likely limited in number), perform an intersection with the polygon and then replace the original geometry with the intersection.

@gboeing What is the reason why you apply the secondary filter on location? What cases did you encounter where the overpass returned results outside the polygon?

gboeing:
> What is the reason why you apply the secondary filter on location? What cases did you encounter where the overpass returned results outside the polygon?

This truncation is necessary for graph construction. See [here](https://github.com/gboeing/osmnx/blob/v0.15.1/osmnx/graph.py#L429-L432). We truncate once more after this as well for periphery cleaning (otherwise you get inaccurate node degrees as an artifact of artificial boundary effects).

However, I'm starting to wonder if it's really necessary for this current PR (geometry retrieval). I suppose if OSMnx just produces a GeoDataFrame of whatever elements' geometries intersect the polygon according to Overpass, then end user can worry about specifically clipping them as needed on their end after the fact.

AtelierLibre:
> What is the reason why you apply the secondary filter on location? What cases did you encounter where the overpass returned results outside the polygon?

I think that as things stand this happens regularly because if you use `.pois_from_place()` the place is geocoded and retrieved as a polygon (e.g. the outline of a city) from the Nominatim API but the `pois`/`geometries` modules then send the bounding box of the polygon rather than polygon itself to the Overpass API. So you then need to filter again at the end to get the buildings in e.g. Copenhagen rather than the buildings in Copenhagen's bounding box.

I assume that if the polygon (rather than the bounding box) were to be part of the query the filter at the end might not be necessary. Having said that I understand that the footprints module does send the polygon and it doesn't filter at the end but you do not get the same result - see below.

If someone else is able to look at the `downloader` module and API interaction code as mentioned above in relation to splitting the query (and additionally passing the polygon) that would be great because it will take me a while to find time for that at the moment.

```
copenhagen_gdf = ox.gdf_from_place('Copenhagen Municipality,Denmark', tags={'building':True})
copenhagen_footprints = ox.footprints_from_place('Copenhagen Municipality,Denmark')

ax = copenhagen_footprints.plot(figsize=(24,24))
copenhagen_gdf.plot(ax=ax, color='red', alpha=0.6)
```

Copenhagen_gdf_footprints

gboeing:
@AtelierLibre the `footprints` module uses the polygon itself to download data from Overpass. This should be the template to follow for the new `geometries` module. See the [function](https://github.com/gboeing/osmnx/blob/master/osmnx/footprints.py#L18) here for details. To keep things simple, the logic can be generally duplicated from that.

Regarding the broader idea of subdividing query geometries the way the `graph` and `downloader` modules currently do, if that becomes a sticking point to getting this PR wrapped up, let's just note that as a "to do" for future enhancement.

AtelierLibre:
@gboeing okay, thanks. I'm going to try to put some time against it this week/weekend so hopefully I can cover the points above.

AtelierLibre:
@gboeing 
> the footprints module uses the polygon itself to download data from Overpass. This should be the template to follow for the new geometries module. See the function here for details. To keep things simple, the logic can be generally duplicated from that.

Okay, that is what I will try to do.

> Regarding the broader idea of subdividing query geometries the way the graph and downloader modules currently do, if that becomes a sticking point to getting this PR wrapped up, let's just note that as a "to do" for future enhancement.

The subdivision of the geometry and the creation of the `polygon_coord_strs` are very closely linked both in `footprints._osm_footprints_download()` and in `downloader._osm_net_download()` separating them out looks like more work than just copying across directly. Some code duplication could be avoided by extracting

```
    # project to utm, divide polygon up into sub-polygons if area exceeds a
    # max size (in meters), project back to lat-lng, then get a list of
    # polygon(s) exterior coordinates
    geometry_proj, crs_proj = projection.project_geometry(polygon)
    gpcs = utils_geo._consolidate_subdivide_geometry(geometry_proj)
    geometry, _ = projection.project_geometry(gpcs, crs=crs_proj, to_latlong=True)
    polygon_coord_strs = utils_geo._get_polygons_coordinates(geometry)
    utils.log(
        f"Requesting network data within polygon from API in {len(polygon_coord_strs)} request(s)"
    )
```

from `downloader._osm_net_download()` into a separate function e.g. `downloader._make_overpass_polygon_coord_strs()` if you're happy I'll make that change.

Regarding the polygon filter:

Passing the polygon to the Overpass API via `utils_geo._get_polygons_coordinates()` helps because it reduces the area of the query but it won't solve the Copenhagen example on its own because it only passes the exterior rings of polygons/multipolygons. I'm not clear how/if interior rings can be added to the query and I'm not proposing to look at that now. I think keeping the filter at the end is the best solution for the moment.

> Thinking of possible solutions here

@xgerrmann Just out of interest, are there cases where passing the whole geometry to `utils_geo._intersect_index_quadrats()` doesn't work? I haven't dug into the logic of it enough to know, but if it does, it would be great to use it as is for the moment and perhaps have a faster/better solution as part of a future pull request?

gboeing:
> if you're happy I'll make that change (may be best as a separate pull request?)

That sounds good.

> I'm not clear how/if interior rings can be added to the query and I'm not proposing to look at that now. I think keeping the filter at the end is the best solution for the moment.

In general, I'm inclined to think the best plan here is to mimic the `graph` functionality as closely as reasonable. Both the `footprints` and `pois` modules have their own limitations right now. But if the downloading logic and results are comparable between `graph` and `geometries` then it will be relatively straightforward to continue enhancing them in sync in the future. Furthermore, I'll be deprecating the `footprints` and `pois` modules after this PR is merged.

AtelierLibre:
@xgerrmann @gboeing I'm going through trying to make things consistent with the `graph` module. I want to flag up that the proposed `gdf_from_bbox()` isn't significantly different from `gdf_from_point()` other than taking a `(width, height)` tuple rather than a distance. While I appreciate this will change the work by @xgerrmann I think making the parameters consistent with `graph_from_bbox()` e.g. `gdf_from_bbox(north, south, east, west, tags)` might be a better starting postion for this module. If you're okay with this I'll make the change.

gboeing:
> I think making the parameters consistent with graph_from_bbox() e.g. gdf_from_bbox(north, south, east, west, tags) might be a better starting postion for this module. If you're okay with this I'll make the change.

Agreed. Thanks!

AtelierLibre:
@gboeing I think this is getting to the stage where it could do with some thorough testing. I've been doing some myself (and will keep doing more) but it's a bit random and limited in scope so help with that would be much appreciated.

I have generally tried to stay away from the `graph` module but to get `gdf_from_xml()` to work I had to make some changes to the `_OSMContentHandler`. I don't think they should affect `graph_from_xml()` but might be worth keeping an eye on that. On a related note, I think moving `_overpass_json_from_file()` and `_OSMContentHandler` from `graph` to `downloader` would help to make clear that they are shared with `geometries`.

Let me know what you think/what's the best way forward.

gboeing:
Fantastic. Thank you. I'm grateful for all the hard work you've put in. Perhaps we should merge this PR into a feature branch here, when you're comfortable with it as a "first draft"? And then start some coordinated testing?

> On a related note, I think moving _overpass_json_from_file() and _OSMContentHandler from graph to downloader would help to make clear that they are shared with geometries.

I think that's fine.

> I think this is getting to the stage where it could do with some thorough testing. 

Certainly. Let's do something like this. I'll work up a big processing batch to start downloading geometries all over the world, passing in various tag combos. My goal will be to look for exceptions: do we find anything that breaks the code. I'll aim for several thousand requests sampled across urban areas worldwide.

In the meantime, it may be most efficient if your testing focuses on validation... inspecting if individual queries conform to expectations, given the data on OSM. @xgerrmann would you have any appetite for validation-style testing too?

AtelierLibre:
No problem, it's nice to have the opportunity to contribute, and it will be good to check that it all works as expected.

> Perhaps we should merge this PR into a feature branch here, when you're comfortable with it as a "first draft"? And then start some coordinated testing?

That sounds good. I think that makes a lot of sense. I'll keep going with some smaller scale tests for a couple more days and then assuming the outputs are as expected it will be great to scale it up and check the robustness.

AtelierLibre:
I have been working on a few tests which are in notebooks here:

https://github.com/AtelierLibre/osmnx_tests/tree/master/notebooks

The notebooks are a bit rough but the results are standing up so far. I'll do a few more and then we can broaden it out.

gboeing:
Ok. I'm going to merge this into a `geometries` feature branch. Testing and additional development can continue there.

gboeing:
See #549

PR #547, opened 2020-08-08
Create function _make_overpass_polygon_coord_strs()

AtelierLibre:
Separates out creation of the polygon query exterior coordinate strings from `downloader._osm_net_download()` to allow reuse in the proposed `geometries` module. The change should be transparent and have no impact on existing functions.


gboeing:
Thanks! Looks like the CI tests failed. If you check the logs and make the necessary changes, ping me and I can review/merge when it's ready!

AtelierLibre:
Thanks, I think fixing the docstring has resolved it.

PR #549, opened 2020-08-24
add geometries module

### LINKED ISSUE #478: Are the pois module and footprints module redundant?

gboeing:
This PR consolidates testing and revisions to the work done by @AtelierLibre in #542. Closes #478.

It also incorporates the changes in #552 including:

  - renames all the `geometries.gdf_from_x` functions -> `geometries.geometries_from_x`
  - makes polygon_features module private
  - deprecates the `footprints` module/functions with a user warning
  - deprecates the `pois` module/functions with a user warning
  - cleans up comments and docstrings
  - autodocs the `geometries` module

Plus some refactoring/generalizing of code in the `truncate` and `utils_geo` modules to make pre-existing code more general for use in both graph construction as well as geometries GeoDataFrame construction.


gboeing:
@AtelierLibre one relatively frequent error I'm hitting in my tests is the topological error reproduced below. I've (proposed) fixed it in ed1ca01.

```python
import osmnx as ox
from shapely.wkt import loads
tags = {'building': True}
p = 'POLYGON ((-38.25258811245654 -9.389736599198256, -38.23246576205545 -9.389736599198256, -38.2330197813699 -9.39786619752298, -38.20283581838461 -9.39786619752298, -38.20394448689717 -9.414125613420136, -38.24419093881911 -9.414125613420136, -38.243081102365 -9.39786619752298, -38.25314242336009 -9.39786619752298, -38.25258811245654 -9.389736599198256))'
polygon = loads(p)
geoms = ox.geometries.gdf_from_polygon(polygon, tags)
```

```
TopologyException: side location conflict at -38.222845100000001 -9.3991738999999992

---------------------------------------------------------------------------
PredicateError                            Traceback (most recent call last)
~/miniconda3/envs/ox/lib/python3.8/site-packages/shapely/predicates.py in __call__(self, this, other, *args)
     14         try:
---> 15             return self.fn(this._geom, other._geom, *args)
     16         except PredicateError as err:

~/miniconda3/envs/ox/lib/python3.8/site-packages/shapely/geos.py in errcheck_predicate(result, func, argtuple)
    579     if result == 2:
--> 580         raise PredicateError("Failed to evaluate %s" % repr(func))
    581     return result

PredicateError: Failed to evaluate <_FuncPtr object at 0x7f6c17b12dc0>

During handling of the above exception, another exception occurred:

TopologicalError                          Traceback (most recent call last)
<ipython-input-3-0ab905c7bc6b> in <module>
      1 p = 'POLYGON ((-38.25258811245654 -9.389736599198256, -38.23246576205545 -9.389736599198256, -38.2330197813699 -9.39786619752298, -38.20283581838461 -9.39786619752298, -38.20394448689717 -9.414125613420136, -38.24419093881911 -9.414125613420136, -38.243081102365 -9.39786619752298, -38.25314242336009 -9.39786619752298, -38.25258811245654 -9.389736599198256))'
      2 polygon = loads(p)
----> 3 geoms = ox.geometries.gdf_from_polygon(polygon, tags)

~/Dropbox/Documents/School/Prof/Projects/Code/osmnx-repos/osmnx/osmnx/geometries.py in gdf_from_polygon(polygon, tags)
    256 
    257     # create geodataframe from the downloaded data
--> 258     GDF = _create_gdf(response_jsons, polygon, tags)
    259 
    260     utils.log(f"gdf_from_polygon returned geodataframe with {len(GDF)} geometries")

~/Dropbox/Documents/School/Prof/Projects/Code/osmnx-repos/osmnx/osmnx/geometries.py in _create_gdf(response_jsons, polygon, tags)
    409 
    410     # Filter final gdf to requested tags and bounding polygon
--> 411     GDF = _filter_final_gdf(gdf=GDF, polygon=polygon, tags=tags)
    412 
    413     return GDF

~/Dropbox/Documents/School/Prof/Projects/Code/osmnx-repos/osmnx/osmnx/geometries.py in _filter_final_gdf(gdf, polygon, tags)
    857         if polygon:
    858             # get a set of index labels of the geometries that intersect the polygon
--> 859             gdf_indices_in_polygon = utils_geo._intersect_index_quadrats(gdf, polygon)
    860             # create a boolean series, True for geometries whose index is in the set
    861             polygon_filter = gdf.index.isin(gdf_indices_in_polygon)

~/Dropbox/Documents/School/Prof/Projects/Code/osmnx-repos/osmnx/osmnx/utils_geo.py in _intersect_index_quadrats(points, geometry, quadrat_width, min_num)
    381             possible_matches_iloc = sindex.intersection(poly.bounds)
    382             possible_matches = points.iloc[list(possible_matches_iloc)]
--> 383             precise_matches = possible_matches[possible_matches.intersects(poly)]
    384             points_in_geom.update(precise_matches.index)
    385 

~/miniconda3/envs/ox/lib/python3.8/site-packages/geopandas/base.py in intersects(self, other)
    408             intersected.
    409         """
--> 410         return _binary_op("intersects", self, other)
    411 
    412     def overlaps(self, other):

~/miniconda3/envs/ox/lib/python3.8/site-packages/geopandas/base.py in _binary_op(op, this, other, *args, **kwargs)
     65     # type: (str, GeoSeries, GeoSeries, args/kwargs) -> Series[bool/float]
     66     """Binary operation on GeoSeries objects that returns a Series"""
---> 67     data, index = _delegate_binary_method(op, this, other, *args, **kwargs)
     68     return Series(data, index=index)
     69 

~/miniconda3/envs/ox/lib/python3.8/site-packages/geopandas/base.py in _delegate_binary_method(op, this, other, *args, **kwargs)
     49         raise TypeError(type(this), type(other))
     50 
---> 51     data = getattr(a_this, op)(other, *args, **kwargs)
     52     return data, this.index
     53 

~/miniconda3/envs/ox/lib/python3.8/site-packages/geopandas/array.py in intersects(self, other)
    504 
    505     def intersects(self, other):
--> 506         return self._binary_method("intersects", self, other)
    507 
    508     def overlaps(self, other):

~/miniconda3/envs/ox/lib/python3.8/site-packages/geopandas/array.py in _binary_method(op, left, right, **kwargs)
    483             right = right.data
    484 
--> 485         return getattr(vectorized, op)(left.data, right, **kwargs)
    486 
    487     def covers(self, other):

~/miniconda3/envs/ox/lib/python3.8/site-packages/geopandas/_vectorized.py in intersects(data, other)
    667         return _binary_method("intersects", data, other)
    668     else:
--> 669         return _binary_predicate("intersects", data, other)
    670 
    671 

~/miniconda3/envs/ox/lib/python3.8/site-packages/geopandas/_vectorized.py in _binary_predicate(op, left, right, *args, **kwargs)
    329     # empty geometries are handled by shapely (all give False except disjoint)
    330     if isinstance(right, BaseGeometry):
--> 331         data = [
    332             getattr(s, op)(right, *args, **kwargs) if s is not None else False
    333             for s in left

~/miniconda3/envs/ox/lib/python3.8/site-packages/geopandas/_vectorized.py in <listcomp>(.0)
    330     if isinstance(right, BaseGeometry):
    331         data = [
--> 332             getattr(s, op)(right, *args, **kwargs) if s is not None else False
    333             for s in left
    334         ]

~/miniconda3/envs/ox/lib/python3.8/site-packages/shapely/geometry/base.py in intersects(self, other)
    738     def intersects(self, other):
    739         """Returns True if geometries intersect, else False"""
--> 740         return bool(self.impl['intersects'](self, other))
    741 
    742     def overlaps(self, other):

~/miniconda3/envs/ox/lib/python3.8/site-packages/shapely/predicates.py in __call__(self, this, other, *args)
     16         except PredicateError as err:
     17             # Dig deeper into causes of errors.
---> 18             self._check_topology(err, this, other)
     19 
     20 

~/miniconda3/envs/ox/lib/python3.8/site-packages/shapely/topology.py in _check_topology(self, err, *geoms)
     33         for geom in geoms:
     34             if not geom.is_valid:
---> 35                 raise TopologicalError(
     36                     "The operation '%s' could not be performed. "
     37                     "Likely cause is invalidity of the geometry %s" % (

TopologicalError: The operation 'GEOSIntersects_r' could not be performed. Likely cause is invalidity of the geometry <shapely.geometry.multipolygon.MultiPolygon object at 0x7f6b91e93eb0>
```

AtelierLibre:
@gboeing Thanks, I hadn't come across that one. I'm having a look at it. Just quickly though, I think there is an issue with `.buffer(0)` as it seems to convert Points to Empty Polygons which then don't intersect with the query Polygon and so POIs are lost. I'm getting better results with `.convex_hull` as Points stay as Points and LineStrings also seem to go through better.

An alternative approach could be to actually apply the buffer onto any invalid geometries in the GeoDataFrame itself first before doing the intersection. I've always tried to avoid that in order to give the closest representation of the geometry as drawn in OSM (even if it is strictly invalid because of overlapping polygons etc. as in the example that you highlight https://www.openstreetmap.org/relation/8303209) and also because the Shapely docs explicitly state that the geometry that results from `buffer(0)` can be unpredictable even if strictly valid. I'm beginning to wonder if that is the right approach though as end users will then come across these errors themselves when they start trying to operate on GeoDataFrames that have faithfully reproduced geometries that are invalid on OSM.

I'd like to do a few more tests, but would be interested to know what you think about how to handle invalid geometries that have been faithfully reproduced from OSM.

gboeing:
> it seems to convert Points to Empty Polygons

Good point. Might be best to reverse the proposed fix I made and do something more along the lines of what you suggest:

> An alternative approach could be to actually apply the buffer onto any invalid geometries in the GeoDataFrame itself

I generally like that idea... though I am also a bit wary of transforming (truly invalid) raw data in a way that might surprise users. I suppose there's a design decision question here: if OSM has invalid geometries, should OSMnx:

  - throw an exception because it cannot process it
  - raise a warning then try to forcibly fix it with `buffer(0)`
  - silently try to forcibly fix it with `buffer(0)`
  - have a function argument like `fix_invalid=True` that the user can turn on/off

Thoughts?

AtelierLibre:
My thoughts are that invalid geometries in OSM are inevitable so for anyone trying to process many large queries with OSMnx raising exceptions isn't an option. I also think silently fixing it isn't great for the reasons stated above. As an OSM contributor I like that OSMnx can log invalid geometries that it encounters with a link directly to the relevant geometry - it makes it easy to go into OSM and fix them. So I think my preference is to raise a warning, log a link to the geometry and try the forced fix. You did a lot of work to tidy up the API which I think was great, so I'm not keen on the function argument but (and I haven't looked at this) could there be a config setting for `retain_invalid=True`?

gboeing:
Ok, I'd say for now let's go with option 2: log a warning w/link then try to forcibly fix it with buffer(0). We can consider adding a function argument in the future if this becomes a sticking point for many users, but I don't anticipate it will.

Would you like to make that tweak, then open a PR onto this branch?

AtelierLibre:
Sure, yes, okay. I need to sort out my various branches a bit - presumably at this stage I'm okay to delete branches on my side and `git rebase upstream/master` or `git reset --hard upstream/master` to get everything back in sync with your `master` and `geometries` branch?

gboeing:
Yes.

Ok, at this point I've tested ~13,000 urban areas worldwide using `tags = {'building': True}` and the only errors I'm hitting are:

  - the invalid geometry topology errors discussed above
  - memory exhaustion errors (occasionally on the server but usually on my machine) for the largest urban areas with lots of building footprint data coverage. This is as expected.

I'm not doing any validation work on the results here. Just looking for errors.

I'll proceed with an assortment of other `tags` combinations next to see if I can unearth any other errors.

AtelierLibre:
That's great thanks.

And good news about the tests as well. I'll make the changes we discussed and I will keep doing small scale validation against overpass-turbo.eu - that seems like the best benchmark for finding/understanding any inconsistencies.

gboeing:
More test results... I downloaded geometries across ~3000 urban areas worldwide using the following tags:

```python
tags = {'amenity' : True,
        'boundary': 'protected_area',
        'highway' : 'path',
        'landuse' : ['retail', 'commercial'],
        'highway' : 'bus_stop'}
```

The only error I got was the same sporadic `TopologyException: side location conflict` described earlier.

One other comment: I do frequently get `EmptyOverpassResponse` errors (in small cities without any features matching those tag combinations), but this is expected. I wonder if it would be better to handle this specific error in the `geometries` module and return an empty `GeoDataFrame`.

AtelierLibre:
That all sounds good, I'll post some more test results tomorrow - very small scale compared to what you're doing but I'm getting good accuracy which is nice.

Regarding the `EmptyOverpassResponse`, that bit of code is copied verbatim from `create_graph()` so they behave identically at the moment. It would probably be tidier to do what you suggest though, replacing the error with a logged warning and returning an empty `gdf` directly from the `if` clause. I'm happy to do that if you want to go ahead with that.

gboeing:
> I'm happy to do that if you want to go ahead with that.

That'd be great, thanks!

AtelierLibre:
Just to note I have also uploaded a [new notebook of comparisons with overpass-turbo](https://github.com/AtelierLibre/osmnx_tests/blob/master/notebooks/09_further_overpass-turbo_comparisons.ipynb). Underneath the `comparison_report()` cell there is a small dataframe that quantifies the geometries in the GeoDataFrame created by the `geometries` module against a GeoDataFrame created from a geojson downloaded from overpass-turbo.eu for a similar query. A few notes:

- Some differences are a bit opaque as it is not clear to me why overpass-turbo included certain geometries in its results.
- There are very occasional multipolygon relations with members that don't have 'role' tags. These don't meet the tagging guidelines on the OSM wiki but overpass-turbo does return them whereas the geometries module doesn't.
- `geometries` includes geometries that intersect the query polygon at a single point or along a line, overpass-turbo seems to exclude them. Amending this would require modifying `utils_geo._intersect_index_quadrats(gdf, polygon)`. I haven't looked at this and I'm not sure it's strictly necessary. It's nice to have something to benchmark against but I don't think there's a need to replicate overpass-turbo's results exactly and again it crops up very occasionally.

These few small differences aside, I think the results are looking very comparable.

gboeing:
Worth noting that the `footprints` module does not appear to handle [this place-based query](https://stackoverflow.com/questions/63609216/) properly, but the `pois` module and the new `geometries` module *do*.

AtelierLibre:
> Worth noting that the `footprints` module does not appear to handle [this place-based query](https://stackoverflow.com/questions/63609216/) properly, but the `pois` module and the new `geometries` module _do_.

I think this is a benefit of consolidating the overpass query creation/reading of the XML through to returning the response JSONS in the `downloader` module and should mean that in future `geometries` and `graph` stay in sync and both benefit from any work on this part of the process.

gboeing:
@AtelierLibre FYI I'm seeing a warning in the `geometries` module:

```
/osmnx/osmnx/geometries.py:435: UserWarning: GeoSeries.isna() previously returned True for both missing (None) and empty geometries. Now, it only returns True for missing values. Since the calling GeoSeries contains empty geometries, the result has changed compared to previous versions of GeoPandas.

  Given a GeoSeries 's', you can use 's.is_empty | s.isna()' to get back the old behaviour.
```

AtelierLibre:
@gboeing I'm not sure from your message if you're just letting me know or if you think we need to 'fix' this, so at the risk of telling you something you already know:

geopandas used to find both missing (None) and empty geometries (e.g. `Point()`) with `gdf["geometry"].isna()`. It doesn't do that any more so it warns that to find both missing and empty geometries you need to use `gdf["geometry"].is_empty | gdf["geometry"].isna()` - this is what [line 435](https://github.com/gboeing/osmnx/blob/696eff61bcb4e2c97bb7a6b13828d21d025d9822/osmnx/geometries.py#L435) does (I've reordered it in the pull request to match the warning exactly). The issue is that the 'solution' still involves calling `.isna()` on empty geometries so the warning is still triggered. The only way to stop the warning is to actively suppress it as explained in the message.

I've included suppression of the warning in pull request #553. Apologies if this isn't what you were asking for...

MWE triggering the warning:
```
import geopandas as gpd
from shapely.geometry import Point

gdf = gpd.GeoDataFrame({"geometry": [Point()]})
gdf["geometry"].isna()
```

gboeing:
Ok at this point I'm pretty satisfied with this PR and consider it a "release candidate" nearly ready to merge into master.

@AtelierLibre would you mind running one final round of validation tests, to see if any last issues have cropped up?

@xgerrmann if you have any availability, would you be interested in doing some validation tests with this `geometries` branch as well to see if anything does not look ready for primetime?

AtelierLibre:
@gboeing yes, sure, happy to do that. I'll need to spread them out a bit over the next couple of days and agreed, it would be great to get some more eyes on this if possible.

AtelierLibre:
@gboeing I've updated my [test notebooks](https://github.com/AtelierLibre/osmnx_tests) with the new `ox.geometries_from_XXX()` functions and extended them with more study areas. The comparisons with overpass-turbo.eu are still holding up and the visual comparisons with openstreetmap.org are also good. I haven't had any exceptions raised, I have had kernel crashes due to memory issues. I'll keep running checks but, in the tests I've done, I haven't found any major issues.

There are a few things worth being aware of but unless you disagree I don't think they should affect the current pull request.

***XMLs***
The `geometries_from_XML()` function behaves differently from the others because of the XML input:
- XMLs seem to be 'clipped' so some geometries will be incomplete and discarded. In contrast the Overpass API JSONs always seem to contain complete (if sometimes invalid) geometries.
- Unlike with the Overpass API functions there is no input filter. The whole of the XML has to be processed to geometries before any polygon/tag filters are applied. I had thought processing XMLs would allow downloading and processing larger areas - in fact the reverse is true and I found that I couldn't process XMLs >90MB which typically cover relatively small areas. To work with large areas of downloaded OSM data it seems it would be necessary to switch to .pbf files which you can selectively extract data from.

Despite these points the `.geometries_from_XML()` function does cope with the incomplete geometries and gives good results for the geometries that are complete.

***Coastlines***
Reading the [OSM coastlines page](https://wiki.openstreetmap.org/wiki/Tag:natural=coastline) it's clear that coastlines are different from other multipolygon geometries. They are made up of individual ways individually tagged with `natural=coastline`, are not collected into multipolygon relations and do not have a `role`. They instead rely on the direction of the line to determine which side is land/water. At the moment `geometries` just converts each individual way to a linestring. I can imagine there would be a way to construct areas of land/water by chaining the coastline ways together and using them to subdivide the query polygon but this would obviously be an entirely new process to try to incorporate. Sometimes coastlines coincide with `place` tags e.g. `place=island` and so these can be used as a proxy but this is inconsistent. I saw your reply on StackOverflow suggesting using a geocoded admin polygon as a possible stand in.

***Confirmation of invalid polygons***
Just to mention http://tools.geofabrik.de/osmi/ has some useful tools that highlight invalid geometries and can be used as a check for the geometries that `geometries` rejects as invalid.

gboeing:
Thanks @AtelierLibre for documenting those points. I agree with you. I think coastlines are fine to represent as LineStrings or at least can remain a "known quirk". And that XML functionality is fine in my mind: OSMnx will process it as is, clipped and all.

Keep me posted if anything else comes up. And @xgerrmann @capilgram if you do get a chance to test the new `geometries` module, let us know if you spot any issues or errors. I'll plan to merge into master on Monday probably if everything remains all good.

gboeing:
@AtelierLibre thanks for all your hard work on this PR! I really appreciate it. I haven't seen any other validation issues or errors crop up in the past 4 days so I believe this is ready to merge now.

I'm aiming to include this in the release of v0.16.0 this week. When it's in the wild we'll have more eyeballs on it, which may turn up little things to fix here and there.

AtelierLibre:
@gboeing no problem, thanks, it was good to work on it. I hope it has brought a new level of consistency to the geopandas side and that it's seen as an improvement on what was already good in the pois and footprints modules. I also hope it all continues to behave as expected but, as you say, when it starts getting used by others I guess issues will probably crop up and I'll be happy to help with any fixes where I can.

PR #550, opened 2020-08-26
Proposed fix for TopologyException intersecting invalid geometries with the query polygon

AtelierLibre:
Proposed fix for TopologyException encountered when trying to intersect invalid geometries with the query polygon as discussed in #549.

Identifies rows with invalid geometries, logs their ids and applies a `.buffer(0)`

Also makes use of `gdf` or `GDF` more consistent.

### DIFF
--- a/osmnx/geometries.py
+++ b/osmnx/geometries.py
@@ -860,7 +860,7 @@ def _buffer_invalid_geometries(GDF):
     if not GDF.empty:
 
         # create a filter for rows with invalid geometries
-        invalid_geometry_filter = ~GDF['geometry'].is_valid
+        invalid_geometry_filter = ~GDF["geometry"].is_valid
 
         # if there are invalid geometries
         if sum(invalid_geometry_filter) > 0:
@@ -873,7 +873,8 @@ def _buffer_invalid_geometries(GDF):
                 "Some invalid geometries were created and included in the GeoDataFrame. "
                 "A buffer of 0 has been applied to try to make them valid "
                 f"and their unique ids have been logged. {invalid_geometry_ids}",
-                stacklevel=2)
+                stacklevel=2,
+            )
 
             # create a list of their urls and log them
             osm_url = "https://www.openstreetmap.org/"
@@ -881,7 +882,9 @@ def _buffer_invalid_geometries(GDF):
             utils.log(f"Invalid geometries that had .buffer(0) applied: {invalid_geom_urls}")
 
             # apply .buffer(0)
-            GDF.loc[invalid_geometry_filter, 'geometry'] = GDF.loc[invalid_geometry_filter, 'geometry'].buffer(0)
+            GDF.loc[invalid_geometry_filter, "geometry"] = GDF.loc[
+                invalid_geometry_filter, "geometry"
+            ].buffer(0)
 
     return GDF
 
--- a/osmnx/geometries.py
+++ b/osmnx/geometries.py
@@ -50,7 +50,7 @@ def gdf_from_bbox(north, south, east, west, tags):
 
     Returns
     -------
-    GDF : geopandas.GeoDataFrame
+    gdf : geopandas.GeoDataFrame
 
     Notes
     -----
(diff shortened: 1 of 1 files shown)


AtelierLibre:
That became more substantial than I had anticipated. I'm not sure why that issue hasn't come up for me before - now that it is explicitly looking for them, my tests find invalid geometries where they didn't complain before. I'm not sure if the warning as I've implemented it is necessary - maybe the logging would be enough.

osmnx/geometries.py
```diff
@@ -407,8 +409,16 @@ def _create_gdf(response_jsons, polygon, tags):
...
+    GDF = _buffer_invalid_geometries(GDF)
+
     # Filter final gdf to requested tags and bounding polygon
-    GDF = _filter_final_gdf(gdf=GDF, polygon=polygon, tags=tags)
+    GDF = _filter_gdf_by_polygon_and_tags(GDF, polygon=polygon, tags=tags)
+
+    # bug in geopandas <=0.8.1 raises a TypeError if trying to plot empty geometries
+    # but missing geometries (gdf['geometry'] = None) cannot be projected e.g. gdf.to_crs()
+    # don't see any option other than to drop rows with missing/empty geometries
+    GDF = GDF[~(GDF["geometry"].isna() | GDF["geometry"].is_empty)].copy()
```
gboeing, line 421: What version of geopandas fixes this bug? Is it forthcoming in a future release?

osmnx/geometries.py
```diff
@@ -823,17 +833,74 @@ def _subtract_inner_polygons_from_outer_polygons(element, outer_polygons, inner_
...
+        invalid_geometry_filter = ~GDF["geometry"].is_valid
+
+        # if there are invalid geometries
+        if sum(invalid_geometry_filter) > 0:
+
+            # get their unique_ids from the index
+            invalid_geometry_ids = GDF.loc[invalid_geometry_filter].index.to_list()
+
+            # show a warning
+            warnings.warn(
```
gboeing, line 872: I would propose removing use of the `warnings` module and instead using `ox.log(message, level=lg.WARNING)` to keep it consistent with the rest of the codebase.

osmnx/geometries.py
```diff
@@ -823,17 +833,74 @@ def _subtract_inner_polygons_from_outer_polygons(element, outer_polygons, inner_
...
+                "Some invalid geometries were created and included in the GeoDataFrame. "
+                "A buffer of 0 has been applied to try to make them valid "
+                f"and their unique ids have been logged. {invalid_geometry_ids}",
+                stacklevel=2,
+            )
+
+            # create a list of their urls and log them
+            osm_url = "https://www.openstreetmap.org/"
```
gboeing, line 880: I think just supplying the OSMID itself (rather than a full URL to it) is sufficient for users who want to look it up?

osmnx/geometries.py
```diff
@@ -407,8 +409,16 @@ def _create_gdf(response_jsons, polygon, tags):
     # Set default crs
     GDF.crs = settings.default_crs
```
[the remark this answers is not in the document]
AtelierLibre: Sure, I was mirroring the `G` in the `graph` module but I agree it is inconsistent and will replace it.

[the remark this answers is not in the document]
gboeing: Thanks. Yeah, the capital `G` graph always bugs me as it doesn't really conform to Python convention... but it's the standard NetworkX convention for graph objects.

osmnx/geometries.py
AtelierLibre: I raised these two issues last week, they've flagged as targeted for 0.9.

[BUG: Plotting empty geometries generates a Type Error (shapely only) #1572](https://github.com/geopandas/geopandas/issues/1572)
[BUG: to_crs() fails if there's None in geometry #1573](https://github.com/geopandas/geopandas/issues/1573)

osmnx/geometries.py
AtelierLibre: Sure, will do.

osmnx/geometries.py
AtelierLibre: Yes. It is partly a convenience at the moment as it creates a clickable link in the terminal log which takes you directly to that geometry on osm.org. If it's all right I'd like to keep it at the moment and can take it out once all of the testing is out of the way.

gboeing: Certainly.

AtelierLibre:
I think these three commits should cover those issues.

PR #551, opened 2020-08-28
creates empty GeoDataFrame if there are no elements in the overpass response

AtelierLibre:
The main change here is the creation of an empty GeoDataFrame if the response JSONs do not contain any elements. It also removes the use of the EmptyOverpassResponse error. I have added an `else` statement to make the two branches within `_create_gdf()` clear.

It tidies up some of the logging to give a better record of when and how many geometries are being created/removed and moves logging of the final GeoDataFrame into one place at the end of `_create_gdf()`

### DIFF
--- a/osmnx/geometries.py
+++ b/osmnx/geometries.py
@@ -340,8 +340,7 @@ def _create_gdf(response_jsons, polygon, tags):
 
         # log a warning
         utils.log(
-            "Empty GeoDataFrame returned. Check query tags and location.",
-            level=lg.WARNING,
+            "Empty GeoDataFrame returned. Check query tags and location.", level=lg.WARNING,
         )
 
         return gdf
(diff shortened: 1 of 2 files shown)


AtelierLibre:
Just to note I have also uploaded a [new notebook of comparisons with overpass-turbo](https://github.com/AtelierLibre/osmnx_tests/blob/master/notebooks/09_further_overpass-turbo_comparisons.ipynb). Underneath the `comparison_report()` cell there is a small dataframe that quantifies the geometries in the GeoDataFrame created by the `geometries` module against a GeoDataFrame created from a geojson downloaded from overpass-turbo.eu for a similar query. A few notes:

- Some differences are a bit opaque as it is not clear to me why overpass-turbo included certain geometries in its results.
- There are very occasional multipolygon relations with members that don't have 'role' tags. These don't meet the tagging guidelines on the OSM wiki but overpass-turbo does return them whereas the geometries module doesn't.
- `geometries` includes geometries that intersect the query polygon at a single point or along a line, overpass-turbo seems to exclude them. Amending this would require modifying `utils_geo._intersect_index_quadrats(gdf, polygon)`. I haven't looked at this and I'm not sure it's strictly necessary. It's nice to have something to benchmark against but I don't think there's a need to replicate overpass-turbo's results exactly and again it crops up very occasionally.

These few small differences aside, I think the results are looking very comparable.

PR #552, opened 2020-08-28
Updates to geometries branch

gboeing:
I'm collecting my own proposed changes/revisions to the `geometries` branch in this PR, to keep the `geometries` branch itself clean in #549. This PR:

  - renames all the `geometries.gdf_from_x` functions -> `geometries.geometries_from_x`
  - makes polygon_features module private
  - deprecates the `footprints` module/functions with a user warning
  - deprecates the `pois` module/functions with a user warning
  - cleans up comments and docstrings
  - autodocs the `geometries` module

Note: the `geometries_from_x` renaming is intended to keep the naming logic consistent. The purpose of the function is to retrieve geospatial objects' geometries from OSM, so the new naming scheme focuses on that rather than the specific datatype of the returned object (a GeoDataFrame). This is similar to the `graph_from_x` function naming in the graph module (rather than specifically referring to the datatype: MultiDiGraph).


gboeing:
Code for testing deprecation messages in `pois` and `footprints` modules:

```python
import osmnx as ox

point = (34.061757, -118.300361)
address = '3807 Wilshire Blvd, Los Angeles, CA 90010'
place = 'Koreatown, Los Angeles, California'
polygon = ox.utils_geo.bbox_to_poly(*ox.utils_geo.bbox_from_point(point, dist=500))
tags = {'amenity': True}

gdf = ox.footprints_from_point(point)
print(gdf.shape)

gdf = ox.footprints_from_address(address)
print(gdf.shape)

gdf = ox.footprints_from_place(place)
print(gdf.shape)

gdf = ox.footprints_from_polygon(polygon)
print(gdf.shape)

gdf = ox.pois_from_point(point, tags)
print(gdf.shape)

gdf = ox.pois_from_address(address, tags)
print(gdf.shape)

gdf = ox.pois_from_place(place, tags)
print(gdf.shape)

gdf = ox.pois_from_polygon(polygon, tags)
print(gdf.shape)
```

osmnx/geometries.py
```diff
@@ -351,43 +349,46 @@ def _create_gdf(response_jsons, polygon, tags):
...
                         point = _parse_node_to_point(element=element)
                         geometries[unique_id] = point
 
                 elif element["type"] == "way":
                     # Parse all ways to linestrings or polygons
                     linestring_or_polygon = _parse_way_to_linestring_or_polygon(
-                        element=element, coords=coords, polygon_features=polygon_features,
+                        element=element, coords=coords
                     )
                     if linestring_or_polygon:
```
gboeing, line 393: @AtelierLibre What are we checking here? `linestring_or_polygon` is a dict right? Are we checking if it's empty (or, alternatively, if it's null)? If so, let's make that explicit like `len(linestring_or_polygon) > 0`.

osmnx/geometries.py
```diff
@@ -403,9 +404,9 @@ def _create_gdf(response_jsons, polygon, tags):
                     if multipolygon:
```
gboeing, line 404: @AtelierLibre Same comment here about the check as above regarding `linestring_or_polygon`

osmnx/geometries.py
```diff
@@ -615,36 +618,46 @@ def _closed_way_is_linestring_or_polygon(element, polygon_features):
             is_polygon = False
```
gboeing, line 618: @AtelierLibre this is already `False` by default 10 lines above... can we just `pass` here instead?

osmnx/geometries.py
AtelierLibre: @gboeing I started `geometries` by duplicating `pois`. This is a remnant of [line 344](https://github.com/gboeing/osmnx/blob/f05b359c418ff0139f6696d504b766a35775ef7c/osmnx/pois.py#L344). I imagine it was there to catch the possibility that [_parse_polygonal_poi()](https://github.com/gboeing/osmnx/blob/f05b359c418ff0139f6696d504b766a35775ef7c/osmnx/pois.py#L142) wouldn't return anything because it triggered the exception due to an invalid geometry. You're right that the new [_parse_way_to_linestring_or_polygon()](https://github.com/gboeing/osmnx/blob/998613e249e8957510aa6abb091161604fb03104/osmnx/geometries.py#L493) doesn't behave like that and is intended to catch the specific errors, log them and return an emtpy geometry as part of a dictionary that is always created. I think this is redundant and can be removed.

osmnx/geometries.py
AtelierLibre: @gboeing `pois` used to handle relations in a very different way, I suspect I just copied [line 344](https://github.com/gboeing/osmnx/blob/f05b359c418ff0139f6696d504b766a35775ef7c/osmnx/pois.py#L344) down when I was making the parsing of relations consistent with ways so the same comment as above applies.

osmnx/geometries.py
AtelierLibre: Yes, I think so.

AtelierLibre:
@gboeing I've read through the PR and your comment at the top and everything looks okay to me. Hopefully the responses I've given to your questions are clear.

I'm going to respond to your comment in #549 on the place-based query there to keep the conversations clear.

AtelierLibre:
One very small comment is that [the text for the logging the number of requests to the API](https://github.com/gboeing/osmnx/blob/998613e249e8957510aa6abb091161604fb03104/osmnx/downloader.py#L333) should probably be changed to reflect that it is not just getting network data. You may have a preferred wording for that.

In a similar vein, I think it is probably worth revisiting the terminology in ['_intersect_index_quadrats()`](https://github.com/gboeing/osmnx/blob/998613e249e8957510aa6abb091161604fb03104/osmnx/utils_geo.py#L340) to reflect that it is now not just being used to identify points that intersect a (multi)polygon but also linestrings and (multi)polygons. Aside from the possible debate about whether a polygon that shares a single point with the boundary should count as inside it (which I think is moot given that we're not clipping any other geometries to the boundary), switching the intersection test over to that has 'just worked' so I haven't touched it or looked at amending it.

osmnx/geometries.py
gboeing: Resolved in 96a4489

osmnx/geometries.py
gboeing: Resolved in 96a4489

osmnx/geometries.py
gboeing: Resolved in 8ce1dc1

gboeing:
> One very small comment is that the text for the logging the number of requests to the API should probably be changed to reflect that it is not just getting network data. You may have a preferred wording for that.

Great point. Resolved in 54c228b.

>  I think it is probably worth revisiting the terminology in '_intersect_index_quadrats()` to reflect that it is now not just being used to identify points that intersect a (multi)polygon but also linestrings and (multi)polygons.

Ah yeah, good catch. Resolved in 588af33.

Otherwise I think this is all clear now and ready to merge.

PR #577, opened 2020-10-20
speed up graph_from_gdfs

Labulitiolle:
I have not added a unit test as `test_find_nearest` already test this function. 
Regarding the performances for a graph with 97087 nodes and 274361 edges on my 8GB RAM:
- previous `ox.graph_from_gdfs` ~ 32s.
- new `ox.graph_from_gdfs` ~ 9s.

@gboeing Please let me know what you think.

### DIFF
--- a/osmnx/utils_graph.py
+++ b/osmnx/utils_graph.py
@@ -123,17 +123,26 @@ def graph_from_gdfs(gdf_nodes, gdf_edges, graph_attrs=None):
     """
     if graph_attrs is None:
         graph_attrs = {"crs": gdf_edges.crs}
+
     G = nx.MultiDiGraph(**graph_attrs)
 
-    # add the nodes then each node's non-null attributes
-    G.add_nodes_from(gdf_nodes.index)
+    attr_col_headings = [c for c in gdf_edges.columns if c not in ["u", "v"]]
+    attribute_data = zip(*[gdf_edges[col] for col in attr_col_headings])
+
+    multigraph_edge_keys = gdf_edges["key"]
+    attribute_data = zip(attribute_data, multigraph_edge_keys)
+
+    # Generate graph edges
+    for s, t, attrs in zip(gdf_edges["u"], gdf_edges["v"], attribute_data):
+
+        attrs, multigraph_edge_key = attrs
+        key = G.add_edge(s, t, key=multigraph_edge_key)
+
+        G[s][t][key].update(zip(attr_col_headings, attrs))
+
+    # Add nodes attributes
     for col in gdf_nodes.columns:
         nx.set_node_attributes(G, name=col, values=gdf_nodes[col].dropna())
-
-    # add each edge and its non-null attributes
-    for (u, v, k), row in gdf_edges.set_index(["u", "v", "key"]).iterrows():
-        d = {label: val for label, val in row.items() if isinstance(val, list) or pd.notnull(val)}
-        G.add_edge(u, v, k, **d)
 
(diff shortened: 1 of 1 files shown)


gboeing:
@Labulitiolle the PR's name and your comment refer to the `graph_to_gdfs` function but it looks like the `graph_from_gdfs` function is the only thing you're changing?

Labulitiolle:
@gboeing, My bad, indeed I have only changed the `ox.graph_from_gdfs`.
Note that the `ox.graph_to_gdfs` is running in ~8s thus it is not such a big bottleneck for large datasets. 

I will have a look at this `ox.graph_to_gdfs` today to see if it can be Improved.

gboeing:
@Labulitiolle I merged, but then reverted the merge.

I realized that this PR is adding a bunch of null attribute values to the graph edges. OSM attribute data are sparse, so there are lots of nulls in the node/edge GeoDataFrames. But in the MultiDiGraph attribute data, if a node or edge doesn't have a certain attribute, it should just lack that key:value pair (rather than having key:null).

```python
import osmnx as ox

G = ox.graph_from_place('Los Angeles, CA, USA', network_type='drive')
print(len(G))

nodes, edges = ox.graph_to_gdfs(G)
G2 = ox.graph_from_gdfs(nodes, edges, graph_attrs=G.graph)
```

Then look at an edge in G:
```python
list(G.edges(keys=True, data=True))[0]
```
```
(653688,
 21300195,
 0,
 {'osmid': 398770659,
  'lanes': '6',
  'name': 'National Boulevard',
  'highway': 'secondary',
  'oneway': False,
  'length': 88.678,
  'geometry': <shapely.geometry.linestring.LineString at 0x1fa6fcc7af0>})
```

Then look at an edge in G2:
```python
list(G2.edges(keys=True, data=True))[0]
```
```
(653688,
 21300195,
 0,
 {'osmid': 398770659,
  'lanes': '6',
  'name': 'National Boulevard',
  'highway': 'secondary',
  'oneway': False,
  'length': 88.678,
  'geometry': <shapely.geometry.linestring.LineString at 0x1fa6fcc7af0>,
  'tunnel': nan,
  'bridge': nan,
  'ref': nan,
  'maxspeed': nan,
  'width': nan,
  'service': nan,
  'access': nan,
  'junction': nan})
```

Adding a bunch of null attributes to the node/edge attribute data dicts creates memory (and serialization) inefficiencies, even if it speeds up the gdfs -> graph conversion. We need to make sure null attributes are filtered out, as they were originally.

gboeing:
@Labulitiolle see an idea for a fix in this branch: https://github.com/gboeing/osmnx/blob/fix/osmnx/utils_graph.py#L104-L146

Worth doing some profiling to see its runtime.

gboeing:
Runtimes I'm seeing look like:
  - `master` branch: 16.5
  - `fix` branch: 5.8 seconds
  - your branch: 4.1 seconds

So the fix branch isn't quite as fast as your proposal, but it's close and it avoids the inefficiency of null attributes in all the nodes/edges. Let me know what you think and if you have any other suggestions.

Labulitiolle:
Sorry @gboeing , I should have mentioned that I removed the 'not adding nan attributes' property. In the memory/speed tradeoff I tried to maximize speed as I am using your functions on large, filtered geodataframes.

I think your solution on the `fix branch` seems perfectly suited for OSMnx: Keeping low memory usage (important on OSM data) and significantly improving speed 🚀

gboeing:
Proposed solution in #578.

PR #582, opened 2020-11-05
Update utils_graph.py

d3netxer:
I ran into an error when I tried to plot my graph after running betweeness centrality and trying to plot my graph. Before running betweeness centrality I had to convert my MultiDiGraph to a DiGraph, and when I did this, there were no longer any keys. Therefore ox.plot_graph did not work for me, because it calls ox.utils_graph.graph_to_gdfs and within this file it expects the graph to have keys, and if it doesn't it results in this error: TypeError: __call__() got an unexpected keyword argument 'keys'

inclusions:

  - not sure if there is a related issue(s)
  - a description of the changes proposed in the pull request: see above
  - an example code snippet illustrating usage of the new functionality: the function inputs will not change

### DIFF
--- a/osmnx/utils_graph.py
+++ b/osmnx/utils_graph.py
@@ -58,7 +58,12 @@ def graph_to_gdfs(G, nodes=True, edges=True, node_geometry=True, fill_edge_geome
         if len(G.edges) < 1:
             raise ValueError("Graph has no edges, cannot convert to a GeoDataFrame.")
 
-        u, v, k, data = zip(*G.edges(keys=True, data=True))
+        try:
+            # if graph has keys
+            u, v, k, data = zip(*G.edges(keys=True, data=True))
+        except:
+            k = None
+            u, v, data = zip(*G.edges(data=True))
 
         if fill_edge_geometry:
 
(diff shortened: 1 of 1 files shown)


gboeing:
I'm not clear what problem you are identifying. Can you provide a complete minimal reproducible code snippet to demonstrate the problem?

d3netxer:
yes, I cannot plot a DiGraph that I converted from a MultiDiGraph

```
import osmnx as ox

# network from address, including only nodes within 1km along the network from the address
G = ox.graph_from_address(address='350 5th Ave, New York, NY', dist=1000,
                          dist_type='network', network_type='drive')

D = ox.utils_graph.get_digraph(G, weight="length")

fig, ax = ox.plot_graph(D,
                            figsize=(12,12),
                            show=False,
                            close=False,
                            node_color='#8aedfc',
                            node_size=5,
                            edge_color='#e2dede',
                            edge_alpha=0.25,
                            bgcolor='black')
```

I get this TypeError: __call__() got an unexpected keyword argument 'keys'

I was trying to plot the DiGraph then overlay the nodes with high betweeness centrality. After thinking about it more, I realize I don't need to plot the DiGraph, just plotting the original MultiDiGraph is sufficient. But maybe some users would like ox.plot_graph to work with DiGraphs?

gboeing:
Your code snippet doesn't work because the `plot_graph` function takes a `MultiDiGraph` argument not a `DiGraph`, as you can see in its [documentation](https://osmnx.readthedocs.io/en/stable/osmnx.html#osmnx.plot.plot_graph). If you need to plot a `DiGraph`, just upcast it first:

```python
import networkx as nx
import osmnx as ox
ox.config(use_cache=True)
G = ox.graph_from_address('350 5th Ave, New York, NY', network_type='drive')
D = ox.utils_graph.get_digraph(G, weight="length")
fig, ax = ox.plot_graph(nx.MultiDiGraph(D))
```

Issue #586, opened 2020-11-08
Speed-up plot_route_folium


Labulitiolle:
https://github.com/gboeing/osmnx/blob/c1d402601c739108bcdc86523dabd3c0d71dbac4/osmnx/folium.py#L202

Hi @gboeing ,
As I am still using your package with large data sets I have noticed that this function `graph_to_gdfs(G, nodes=False, fill_edge_geometry=True)` builds a gdf wit edges and nodes from the entire graph. As only the nodes and edges from the path are required to plot the route I would suggest using `graph_to_gdfs(G.subgraph(route), nodes=False, fill_edge_geometry=True)`. 

The speed is significantly improved, the computation time for a graph with 90000 edges and a path of 430 nodes is:
- Current version: `23.7146 s`
- Enhanced version: `0.9569 s`

Should I do a PR ?

gboeing:
Good idea. Yes, thanks.

gboeing:
Closed by #587, thanks @Labulitiolle

falcaozane:
AttributeError: module 'osmnx' has no attribute 'plot_route_folium' 

is this function removed or what??

gboeing:
@falcaozane see #1252.

PR #587, opened 2020-11-09
speed-up - plot_route_folium 

Labulitiolle:
The PR addresses issue #586. It will replace `graph_to_gdfs(G, nodes=False, fill_edge_geometry=True)` by   `graph_to_gdfs(G.subgraph(route), nodes=False, fill_edge_geometry=True)`  as only the edges from the path are required to plot the route.

The speed is significantly improved, the computation time for a graph with 90000 edges and a path of 430 nodes is:

Current version: `23.7146 s`
Enhanced version: `0.9569 s`

### DIFF
--- a/osmnx/folium.py
+++ b/osmnx/folium.py
@@ -199,7 +199,7 @@ def plot_route_folium(
         raise ImportError("The folium package must be installed to use this optional feature.")
 
     # create gdf of the route edges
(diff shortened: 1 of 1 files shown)


gboeing:
I have done additional refactoring to the `folium` module in #595 and #596 for further speed and memory improvements.

PR #584, opened 2020-11-07
Custom Type Conversion - load_graphml

d-wasserman:
This PR addresses issue #575, and willl enable edges and nodes imported with load_graphml to be able to do type conversions for attributes beyond the default attributes currently being converted to floats. This enables users who build graphs with additional edge weights or node attributes that can be properly saved and then loaded based on the desired attribute data type. 

Relevant Code Snippets

```
def load_graphml(filepath, node_type=int, node_dtypes=None, edge_dtypes=None):
    """
    Load an OSMnx-saved GraphML file from disk.

    Converts the node/edge attributes to appropriate data types.

    Parameters
    ----------
    filepath : string
        path to the GraphML file
    node_type : type
        convert node ids to this data type
    node_dtypes : dict of attribute name -> data type
        identifies additional is a numpy.dtype or Python type to cast one or more additional node attributes
        defaults to {"elevation":float, "elevation_res":float, "lat":float, "lon":float, "x":float, "y":float} if None
    edge_dtypes : dict of attribute name -> data type
        identifies additional is a numpy.dtype or Python type to cast one or more additional edge attributes. Defaults
        to {"length": float, "grade": float, "grade_abs": float, "bearing": float, "speed_kph": float,
       "travel_time": float} if None
    Returns
    -------
    G : networkx.MultiDiGraph
    """
    # read the graph from disk
    G = nx.MultiDiGraph(nx.read_graphml(filepath, node_type=node_type))

    # convert node/edge attribute data types
    utils.log("Converting node and edge attribute data types")
    G = _convert_node_attr_types(G, node_type, node_dtypes)
    G = _convert_edge_attr_types(G, node_type, edge_dtypes)
```

PR #588, opened 2020-11-11
Customizable data type specification when loading GraphML file

### LINKED ISSUE #575: load graphML - identify attributes to convert to float.
## Problem Statement

Around the [private methods for converting the datatypes of a few attributes on load](https://github.com/gboeing/osmnx/blob/8653183bb966cfa826e92f1c95b11efa51f99572/osmnx/io.py#L215), there are only  a small subset of attributes that are converted. If someone adds attributes or custom edge weights, they come in as objects by default. There is also no way to adjust which attributes have conversions attempted. 

### Solution

It would be great to see two possible solutions:
1. Ideally, data types of attributes converted to graphML could be preserved. This might be hard to achieve in practice. At minimum, a dictionary could be passed as an optional attribute to the load function. 
2. Quick fix, perhaps the set being used to determine attributes to attempt float conversions with could be a parameter. 

A clear and concise description of what you want to change in OSMnx and how you plan to implement it.

### Alternatives Considered

I think the easiest approach outside of this is converting the nodes & edges to GDF, and doing .astype() conversions there. 

### Minimum Viable Solution
Example with nodes. Granted, edges could be similar, but is clearly more complicated. These changes would be passed up to the load graphML method (for nodes/edges separately). Notebook examples have graphML as the preferred option, so I want to find ways to address limitations. 

```
def _convert_node_attr_types(G, node_type, num_attr = {"elevation", "elevation_res", "lat", "lon", "x", "y"}):
    """
    Convert graph nodes' attributes' types from string to numeric.
    Parameters
    ----------
    G : networkx.MultiDiGraph
        input graph
    node_type : type
        convert node ID (osmid) to this type
    num_attr : set
        attributes to attempt a cast to float for
    Returns
    -------
    G : networkx.MultiDiGraph
    """
        # convert node ID to user-requested type
        data["osmid"] = node_type(data["osmid"])

        # convert numeric node attributes from string to float
        for attr in num_attr:
            if attr in data:
                data[attr] = float(data[attr])

    return G
```

gboeing:
Revises #584 to satisfy CI tests.
Resolves #575.
  - adds customizable node/edge attribute data type arguments
  - deprecates old `node_type` argument with warning; to remove in a future release
  - fixes gephi graphml compatibility and streamlines gephi-compatible saving (see https://github.com/networkx/networkx/issues/4347)

### DIFF
--- a/osmnx/io.py
+++ b/osmnx/io.py
@@ -244,8 +244,14 @@ def _convert_node_attr_types(G, node_type, node_dtypes=None):
     G : networkx.MultiDiGraph
     """
     if node_dtypes is None:
-        node_dtypes = {"elevation": float, "elevation_res": float, "lat": float, "lon": float,
-                       "x": float, "y": float}
+        node_dtypes = {
+            "elevation": float,
+            "elevation_res": float,
+            "lat": float,
+            "lon": float,
+            "x": float,
+            "y": float,
+        }
     for _, data in G.nodes(data=True):
         # convert node ID to user-requested type
         data["osmid"] = node_type(data["osmid"])
@@ -279,8 +285,14 @@ def _convert_edge_attr_types(G, node_type, edge_dtypes=None):
     G : networkx.MultiDiGraph
     """
     if edge_dtypes is None:
-        edge_dtypes = {"length": float, "grade": float, "grade_abs": float,
-                       "bearing": float, "speed_kph": float, "travel_time": float}
+        edge_dtypes = {
+            "length": float,
+            "grade": float,
+            "grade_abs": float,
+            "bearing": float,
+            "speed_kph": float,
+            "travel_time": float,
+        }
     # convert numeric, bool, and list edge attributes from string
     # to correct data types
     for _, _, data in G.edges(data=True, keys=False):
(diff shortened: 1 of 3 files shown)


gboeing:
@d-wasserman @Labulitiolle thanks for contributing to/participating in #584. I've revised it here and I wanted to ask one more favor of you: would you mind taking a look at the code here and testing it out with a couple use cases yourselves to see if it behaves as expected or has any bugs? I appreciate it.

d-wasserman:
The code changes make a lot of sense, I was just uncomfortable changing the API. Deprecating node_id is a good idea and we just wrap it into the dictionary. 
I will test it when I find some time. Any time line you have in mind?

gboeing:
> I will test it when I find some time. Any timeline you have in mind?

@d-wasserman thanks. I would love to incorporate this into the next minor version release, tentatively planned for the middle of next week, so any feedback by say Monday would be fantastic.

d-wasserman:
Geoff, just letting you know I did not get to do any tests yet. I did review the most recent commits and appreciate that you added tests for the changes in functionality for graph loading and resolved some other compatibility issues.

PR #606, opened 2020-12-03
Use pathlib instead of os.path

eumiro:
Replace `os.path.*` with `pathlib.Path` to simplify the code and use individual objects for files and directories.

This changes only internal code and private methods. Next steps would be to update the public methods to accepts both `str` and `Path` objects.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -15,14 +15,14 @@ serve to show the default.
 
 If extensions (or modules to document with autodoc) are in another directory,
 add these directories to sys.path here. If the directory is relative to the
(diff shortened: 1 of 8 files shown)


gboeing:
Hi @eumiro thanks for considering contributing to OSMnx. Per the contributing guidelines, if you'd like to propose a bug fix or enhancement, please open an issue first to discuss its fit with the project.

Also, I see that you used a forced push. We cannot merge in a PR with a forced push, because this is a public repository and force pushing destructively overwrites the remote with your local version, unconditionally overwriting any changes made by others in the meantime.

Issue #608, opened 2020-12-04
Use `pathlib.Path` for directory and file operations


eumiro:
Since `osmnx` is now a Python 3.6+ project, there are several features of the language that allow us to write better and more readable code. One of them is [pathlib](https://docs.python.org/3/library/pathlib.html) that can replace most `os.path/glob/fnmatch` calls.

I have started the PR #606 that changes only internal code and private methods. What do you think about it? And how about updating the public interface as well?

gboeing:
Thanks @eumiro. Yes I think consolidating around `pathlib` would be a good idea at this point. #606 has some force-pushed commits so I'm going to close it. Could you open a new PR (make sure you're working off the current master branch)? I think updating the public functions to accept `string` or `Path` objects would make sense too. I'd like to get this in before v1.0 goes through its final tests (starting late next week) before release.

eumiro:
Thank you, @gboeing . Please see #610 .

gboeing:
Thank you. Closed by #610 and #614.

PR #610, opened 2020-12-04
Use pathlib instead of os.path

eumiro:
Described in #608


gboeing:
Thanks. Is this in progress or ready to merge? Would you like to update the public facing functions as well?

eumiro:
I have now updated the public methods as well and all tests are green on my machine. So you can review the changes now.

gboeing:
@eumiro thanks!

PR #614, opened 2020-12-06
Tweaks to #610: use pathlib instead of os.path

gboeing:
Tweaks to #610: use pathlib instead of os.path


gboeing:
@eumiro I have made some revisions to your PR in this branch. Would you please take a look at the new commits and let me know if this all makes sense? Thanks.

eumiro:
LGTM, thank you for the review and fixes!

PR #619, opened 2020-12-09
compact code in geometries

eumiro:
As discussed in #617, this compacts the code in `osmnx/geometries.py`, such as removing unnecessary code in conditions and list comprehensions.

### DIFF
--- a/osmnx/geometries.py
+++ b/osmnx/geometries.py
@@ -334,12 +334,8 @@ def _create_gdf(response_jsons, polygon, tags):
     gdf : geopandas.GeoDataFrame
         GeoDataFrame of geometries and their associated tags
     """
-    # make sure we got data back from the server request(s)
-    element_count = sum(len(rj["elements"]) for rj in response_jsons)
-    utils.log(f"{element_count} elements in the JSON responses (includes every node).")
-
     # if there are no elements in the responses
-    if not element_count:
+    if not any(len(rj["elements"]) for rj in response_jsons):
 
         # create an empty GeoDataFrame
         gdf = gpd.GeoDataFrame()
@@ -353,8 +349,8 @@ def _create_gdf(response_jsons, polygon, tags):
 
     # else if there were elements in the response
     else:
-        # log creation
-        utils.log("Converting elements to geometries")
+        count = sum(len(rj["elements"]) for rj in response_jsons)
+        utils.log("Converting {count} elements in JSON responses to geometries")
 
         # Dictionaries to hold nodes and complete geometries
         coords = dict()
--- a/osmnx/geometries.py
+++ b/osmnx/geometries.py
@@ -350,7 +350,7 @@ def _create_gdf(response_jsons, polygon, tags):
     # else if there were elements in the response
     else:
         count = sum(len(rj["elements"]) for rj in response_jsons)
-        utils.log("Converting {count} elements in JSON responses to geometries")
+        utils.log(f"Converting {count} elements in JSON responses to geometries")
 
         # Dictionaries to hold nodes and complete geometries
         coords = dict()
--- a/osmnx/geometries.py
+++ b/osmnx/geometries.py
@@ -335,11 +335,11 @@ def _create_gdf(response_jsons, polygon, tags):
         GeoDataFrame of geometries and their associated tags
     """
     # make sure we got data back from the server request(s)
-    element_count = sum([len(rj["elements"]) for rj in response_jsons])
+    element_count = sum(len(rj["elements"]) for rj in response_jsons)
     utils.log(f"{element_count} elements in the JSON responses (includes every node).")
 
     # if there are no elements in the responses
-    if element_count < 1:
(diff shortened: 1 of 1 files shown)


gboeing:
@eumiro thanks for this proposed enhancement. I'm going to request a review from @AtelierLibre as he was the primary author of this module and this PR would greatly benefit from his expert eye.

AtelierLibre:
@gboeing @eumiro I did scan through the changes before and they looked like quite specific optimisations that you might both be better placed to assess than me but I'll have a look and come back to you.

osmnx/geometries.py
```diff
@@ -335,11 +335,11 @@ def _create_gdf(response_jsons, polygon, tags):
         GeoDataFrame of geometries and their associated tags
     """
     # make sure we got data back from the server request(s)
-    element_count = sum([len(rj["elements"]) for rj in response_jsons])
+    element_count = sum(len(rj["elements"]) for rj in response_jsons)
     utils.log(f"{element_count} elements in the JSON responses (includes every node).")
 
     # if there are no elements in the responses
-    if element_count < 1:
+    if not element_count:
```
gboeing, line 343: Don't use implicit false with an integer value.

I wonder if this check can be simpler, like here: https://github.com/gboeing/osmnx/blob/551364776a06d9448ac0a379352182e91dd4817f/osmnx/graph.py#L534

osmnx/geometries.py
```diff
@@ -632,43 +632,40 @@ def _is_closed_way_a_polygon(element, polygon_features=_polygon_features):
             # identify common keys in element's tags and polygon_features dict
             intersecting_keys = element_tags.keys() & polygon_features.keys()
 
-            # if common keys are found
```
gboeing, line 635: @AtelierLibre the changes in this function would particularly benefit from your eye to verify if the logic remains consistent with your original design.

gboeing (changes requested):

osmnx/geometries.py
eumiro: Definitely this would be better because it would just stop as soon as something is found. What about the `utils.log` line 339 that wants to get the number though?

gboeing: It looks like that log call could be moved to line 357 and merged with the exiting log call there. Something like `utils.log("Converting {count} elements in JSON responses to geometries")`

eumiro: Sounds good. I have pushed another commit, thank you for the hint.

The whole if-else block could be inverted (to have a positive condition) and anyway it is quite long and does not fit on a screen. But that would be a larger refactoring.

osmnx/geometries.py
AtelierLibre: @gboeing unless I'm missing something, doesn't this change  just remove the `if len(intersecting_keys) > 0:` check (which I can see is probably redundant) and then outdent the rest of the code?

osmnx/geometries.py
AtelierLibre: Thanks, the summing of the list comprehension did bother me and so the `if not any()` looks like a neat alternative. Merging the log call makes sense as well.

AtelierLibre:
@eumiro @gboeing Thanks both. Just so I'm clear, the principal reasons for the changes are:

- to remove the sum of the list comprehension in [line 338](https://github.com/gboeing/osmnx/pull/619/commits/ff5b39f664b8a8271afa6b278a005f77d0fd250c#diff-d3df5a773bc937a2104af059716174b4c8b93a2d6d83d2f262c37a8d7bcf9a69L338) and to replace it with a `not any()` and `sum()` of a generator later on
- to check for empty lists/dicts with `if not lst:` rather than `if len(lst) == 0:`
- to remove a redundant `if` statement in [line 636](https://github.com/gboeing/osmnx/pull/619/commits/ff5b39f664b8a8271afa6b278a005f77d0fd250c#diff-d3df5a773bc937a2104af059716174b4c8b93a2d6d83d2f262c37a8d7bcf9a69L636)
- to check whether the invalid_geometries_filter series has values in it with `.any()` instead of `sum(invalid_geometries_filter) > 0`

If these changes improve the code then, unless I have missed something, it seems okay to me.

gboeing (approved):

gboeing:
@eumiro tests are failing (could you fix that?) but otherwise this looks good.

osmnx/geometries.py
```diff
@@ -353,8 +349,8 @@ def _create_gdf(response_jsons, polygon, tags):
 
     # else if there were elements in the response
     else:
-        # log creation
-        utils.log("Converting elements to geometries")
+        count = sum(len(rj["elements"]) for rj in response_jsons)
+        utils.log("Converting {count} elements in JSON responses to geometries")
```
gboeing, line 353: Making this a proper f-string will probably fix the tests:
```suggestion
        utils.log(f"Converting {count} elements in JSON responses to geometries")
```

PR #626, opened 2020-12-22
Add query by OSM ID functionality

### LINKED ISSUE #620: Query by OSM ID
When extracting data from nominatim, sometimes a user knows the exact osmid of an entity. The current state of osmnx.downloader.nominatim_request allows for search, reverse, or lookup, but not for details. When I try to feed the response to osmnx.geocode._geocode_query_to_gdf, It does not work, due to the different structure of the response. 

I believe that adding the details request_type to nominatim_request, and adding the parsing of its output in _geocode_query_to_gdf make a good solution.

gboeing:
Resolves #620. Simple demonstration: 

```python
# query by place name, same as before
query = ["Copenhagen", "Berlin", "Central Park, New York"]
gdf = ox.geocode_to_gdf(query)

# new ability to query by OSM ID
query = ["R2192363", "N240109189", "W427818536"]
gdf = ox.geocode_to_gdf(query, by_osmid=True)
```

By passing `by_osmid=True`, the `geocode_to_gdf` function will treat the `query` argument as an OSM ID (or list of OSM IDs) for [Nominatim lookup](https://nominatim.org/release-docs/develop/api/Lookup/) rather than text search. Note that each integer OSM ID must be represented as a string with an N, W, or R prepended to identify its type (node, way, or relation) in accordance with the Nominatim format.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 1.0.0 (TBD)
 
   - set use_cache=True by default
+  - add ability to query a place by OSM ID in geocoder.geocode_to_gdf function
   - add optional setting for download/cache-only mode
   - replace md5 with sha1 for cache filename hashing
   - replace streets_per_node graph attribute with equivalent street_count node attribute
--- a/osmnx/downloader.py
+++ b/osmnx/downloader.py
@@ -498,18 +498,20 @@ def _osm_geometry_download(polygon, tags):
     return response_jsons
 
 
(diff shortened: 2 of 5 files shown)


gboeing:
Demonstration: 

```python
# query by place name, as always
query = ["Copenhagen", "Berlin", "Central Park, New York"]
gdf = ox.geocode_to_gdf(query)

# new ability to query by OSM ID
query = ["R2192363", "N240109189", "W427818536"]
gdf = ox.geocode_to_gdf(query, osmid=True)
```

PR #635, opened 2021-01-03
Remove Python2 numeric relics

eumiro:
Remove code used in Python2 that is no more necessary in Python3, such as:

- `a / b` for two integers returns a float, so it is no more necessary to convert one or both to floats before the division
- `math.floor`, `math.ceil`, and `round` return integers

And a few numeric fixes or visual improvements.

### DIFF
--- a/osmnx/downloader.py
+++ b/osmnx/downloader.py
@@ -595,7 +595,7 @@ def nominatim_request(params, request_type="search", pause=1, error_pause=60):
         sc = response.status_code
 
         # log the response size and domain
-        size_kb = len(response.content) / 1000.0
+        size_kb = len(response.content) / 1000
         domain = re.findall(r"(?s)//(.*?)/", url)[0]
         utils.log(f"Downloaded {size_kb:,.1f}KB from {domain}")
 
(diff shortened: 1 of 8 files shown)


osmnx/downloader.py
```diff
@@ -595,7 +595,7 @@ def nominatim_request(params, request_type="search", pause=1, error_pause=60):
         sc = response.status_code
 
         # log the response size and domain
-        size_kb = len(response.content) / 1000.0
+        size_kb = len(response.content) / 1024
```
gboeing, line 598: Dividing by 1024 puts the size in units of kibibytes rather than kilobytes. I'm going to push a commit to keep it in kilobytes.

osmnx/downloader.py
```diff
@@ -663,7 +663,7 @@ def overpass_request(data, pause=None, error_pause=60):
...
```
gboeing, line 666: Same kibibytes comment as above.

osmnx/downloader.py
gboeing: See: https://en.wikipedia.org/wiki/Byte#Multiple-byte_units

osmnx/downloader.py
gboeing: Resolved by cb41881

osmnx/downloader.py
gboeing: Resolved by cb41881

eumiro: Right, sorry. I am using kibibytes, but keep calling them kilobytes. Thank you for pointing that out.

Issue #648, opened 2021-01-30
function truncate_graph_dist returns an unexpected result


guibar:
I found a problem with this function whereby some unexpected nodes make it into the returned graph.
My debugger pointed me to the issue straight away. Starting on line 40 of truncate.py, we have:

```python
distances = nx.shortest_path_length(G, source=source_node, weight=weight)
distant_nodes = {k: v for k, v in distances.items() if v > max_dist}
G = G.copy()
G.remove_nodes_from(distant_nodes)
```

The problem is that nodes which are unreachable from the source (in my case, nodes which are on an incoming edge on the border of my zone of interest) make it into the copy of G because they are not in distant_nodes as a result of the shortest_path_length call not including them in the distances dict.

**Environment information**
  - I am on Linux, python = 3.8.5
  - osmnx= 1.0.1

Here is a reproducible example:

```python
import osmnx

d2_box = (1.932, 2.204, 49.323, 49.46)
G = osmnx.graph.graph_from_bbox(d2_box[3], d2_box[2], d2_box[1], d2_box[0],
                                network_type='drive', simplify=True, retain_all=False,
                                truncate_by_edge=True, clean_periphery=True)
GG = osmnx.truncate.truncate_graph_dist(G,7846489021,max_dist=20)
list(GG.nodes())

```
returns 
[5538515552, 265534270] which are nowhere near the source node.

gboeing:
Thanks! Do you have a proposed fix you'd like to contribute as a PR?

guibar:
Hi Geoff,

Thanks for all your work btw!
Yes, I can propose something although python is not my language of
expertise.
I am doing it right now.

Guillaume

gboeing:
closed by #651

PR #651, opened 2021-01-31
proposed fix to issue_648 by which unreachable nodes are mistakenly l…

guibar:
This is a simple fix to this issue:
https://github.com/gboeing/osmnx/issues/648
I think the proposed fix is self explanatory.

### DIFF
--- a/osmnx/truncate.py
+++ b/osmnx/truncate.py
@@ -40,13 +40,12 @@ def truncate_graph_dist(G, source_node, max_dist=1000, weight="length", retain_a
     distances = nx.shortest_path_length(G, source=source_node, weight=weight)
 
     # then identify every node further than max_dist away
-    distant_nodes = {k: v for k, v in distances.items() if v > max_dist}
+    distant_nodes = {k for k, v in distances.items() if v > max_dist}
+    unreachable_nodes = G.nodes - distances.keys()
 
     # make a copy to not mutate original graph object caller passed in
     G = G.copy()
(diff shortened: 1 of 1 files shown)


osmnx/truncate.py
```diff
@@ -45,6 +45,8 @@ def truncate_graph_dist(G, source_node, max_dist=1000, weight="length", retain_a
     # make a copy to not mutate original graph object caller passed in
     G = G.copy()
     G.remove_nodes_from(distant_nodes)
+    # nodes unreachable from source_node are missing from distances and should be removed too
+    G.remove_nodes_from(G.nodes - distances)
```
gboeing, line 49: Could you change it to call `remove_nodes_from` only once, like this:

```python
# then identify every node further than max_dist away
distant_nodes = {k for k, v in distances.items() if v > max_dist}
unreachable_nodes = G.nodes - distances.keys()

# make a copy to not mutate original graph object caller passed in
G = G.copy()
G.remove_nodes_from(distant_nodes | unreachable_nodes)
```

gboeing (changes requested):
Thanks @guibar. This looks good. I made one request to make this a bit more efficient. When that's done I can merge it in.

osmnx/truncate.py
gboeing: @guibar actually, no worries, I just pushed this change myself.

gboeing (approved):

PR #661, opened 2021-02-24
Modifying the default filters

### LINKED ISSUE #659: network_type "drive" leaves edges with highway=bus_guideway
After a call such as:

osmnx.graph.graph_from_bbox(bbox[0], bbox[1], bbox[2], bbox[3], network_type='drive', simplify=True, retain_all=False, truncate_by_edge=True, clean_periphery=True)

I end up with a graph which has some edges with tag "highway=bus_guideway" which is a valid tag but not for driving:

https://wiki.openstreetmap.org/wiki/Tag:highway%3Dbus_guideway

I think it is just a matter to exclude that tag in the drive filter. 
I have edited the filter to be:

filters["drive"] = (
        f'["highway"]["area"!~"yes"]{settings.default_access}'
        f'["highway"!~"abandoned|bridleway|construction|corridor|cycleway|elevator|'
        f"escalator|footway|path|pedestrian|planned|platform|proposed|raceway|service|"
        f'steps|track|bus_guideway"]'
        f'["motor_vehicle"!~"no"]["motorcar"!~"no"]'
        f'["service"!~"alley|driveway|emergency_access|parking|parking_aisle|private"]'
    )

and the edges are not there anymore.

guibar:
Edges tagged with bus_guideway are not excluded from the driving, walking and cycling filters. 
Changed the filter values to exclude them.

Fixes #659

### DIFF
--- a/osmnx/downloader.py
+++ b/osmnx/downloader.py
@@ -43,9 +43,9 @@ def _get_osm_filter(network_type):
     # are tagged as providing certain services
     filters["drive"] = (
         f'["highway"]["area"!~"yes"]{settings.default_access}'
-        f'["highway"!~"abandoned|bridleway|construction|corridor|cycleway|elevator|'
+        f'["highway"!~"abandoned|bridleway|bus_guideway|construction|corridor|cycleway|elevator|'
         f"escalator|footway|path|pedestrian|planned|platform|proposed|raceway|service|"
-        f'steps|track|bus_guideway"]'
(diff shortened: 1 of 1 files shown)


gboeing (changes requested):
Could you maintain the alphabetical ordering of the tag values in the filter strings?

gboeing (approved):

gboeing:
Thanks!

PR #670, opened 2021-03-08
handle boolean string conversion in io module

### LINKED ISSUE #665: Function _convert_node_attr_types(G, dtypes=None) in io.py doesn't work with boolean attributes
If I add a boolean field on nodes and save it to graphml, I don't get it back when I read from the file with:

`osmnx.io.load_graphml(graphml_path, node_dtypes=node_dtypes)`

even though I have added by field in node_dtypes with the bool type. 

I can see that my fields are saved correctly as 'True' or 'False' in the graphml file. 

But when reading from the file, it fails because the following code in _convert_node_attr_types:

```
        for attr in dtypes:
            if attr in data:
                data[attr] = dtypes[attr](data[attr])
```
doesn't work with bool i.e. bool('False') doesn't return False but True as can be seen here:

https://stackoverflow.com/questions/715417/converting-from-a-string-to-boolean-in-python

So all my field are set to True. 

We can discuss a way to fix this if you want. Although they are going to be hacky given that the really good solution would be for bool('False') to be False.

gboeing:
resolves #665 

This PR fixes the handling of boolean type conversion in the `load_graphml` function and exposes a new `graph_dtypes` function parameter. This fix is necessary because Python will otherwise parse the string "False" to the boolean value True: that is, `bool("False") == True`.

This fix also streamlines the former handling of boolean values by automating it and making it consistent and configurable. This does raise a caveat, however, for anyone using the `all_oneway=True` configuration to subsequently save XML. If you configured `all_oneway=True` and saved GraphML, you must specify `edge_dtypes` if you want to reload the GraphML file (because it will contain `oneway` values as strings, rather than the expected/default bools):

```python
import osmnx as ox
ox.config(all_oneway=True)
G = ox.graph_from_place('Piedmont, CA, USA', network_type='drive')
ox.save_graphml(G)

# this will raise a ValueError
G2 = ox.load_graphml('data/graph.graphml')

# this will work
G2 = ox.load_graphml('data/graph.graphml', edge_dtypes={'oneway': str})
```


gboeing:
@guibar would you mind taking a look at this PR and seeing if it resolves your use case in #665? Thanks!

guibar:
Hi @gboeing,
It works fine on my use case and by looking at it, looks good too.

PR #675, opened 2021-03-14
refactor distance module

gboeing:
This PR:
  - [x] refactors the distance module
  - [x] adds a new, faster, streamlined `nearest_nodes` function
  - [x] adds a new, faster, streamlined `nearest_edges` function
  - [x] deprecates the old `get_nearest_node`, `get_nearest_nodes`, `get_nearest_edge`, and `get_nearest_edges` in favor of the new superior functions
  - [x] tests that the new functions yield the same results as the old functions did
  - [x] adds `utils_geo.interpolate_points` function and deprecates `redistribute_vertices` in favor of it

### DIFF
--- a/osmnx/bearing.py
+++ b/osmnx/bearing.py
@@ -5,6 +5,12 @@ import math
 import networkx as nx
 import numpy as np
 
+# scipy is an optional dependency for entropy calculation
+try:
+    import scipy
+except ImportError:
+    scipy = None
+
 
 def get_bearing(origin_point, destination_point):
     """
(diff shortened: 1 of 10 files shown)


gboeing:
For now, the new `nearest_edges` function requires `rtree.index.Index` not `pygeos.STRtree` as the spatial index, to use its `nearest` function for nearest-neighbor querying.

https://github.com/gboeing/osmnx/blob/93accb492b50876decf3ecde08e7b69e56942619/osmnx/distance.py#L192-L196

However, it looks pygeos is adding a `STRtree` nearest neighbor function, with plans to implement it in geopandas: https://github.com/pygeos/pygeos/pull/272#issuecomment-786966311 When that happens, we can use it directly here.

PR #678, opened 2021-03-15
better handling of node geometry when converting gdfs to graph

### LINKED ISSUE #677: Issue with projection and geopandas interface
I am building a network using `graph_from_gdfs`, after creating `nodes` and `edges` geodataframes myself. The nodes and edges are in a projected coordinate system (i.e. not raw latitude-longitude). However, when I create the new graph, then extract the nodes and edges using `graph_to_gdfs`, it looks like the nodes have been "un-projected" (back to latitude-longitude), but the edges have not - they remain in the same coordinate system. Furthermore, the created geodataframe is inconsistent - the coordinates are clearly latitudes and longitudes, but the associated projection system indicates otherwise, e.g.

Screen Shot 2021-03-14 at 3 19 47 PM

_but_

Screen Shot 2021-03-14 at 3 20 04 PM

Minimum working example:

```python
import osmnx as ox
import geopandas as gp
# get an example network
network = ox.graph_from_place("Sausalito, CA, USA", network_type='drive', simplify=False)
# extract corresponding geodataframes
nodes, edges = ox.graph_to_gdfs(network)
# project geodataframes to "flat" local coordinates
edges = edges.to_crs("epsg:7131")
nodes = nodes.to_crs("epsg:7131")
# recreate network from projected data
network = ox.graph_from_gdfs(nodes, edges)
# extract geodataframes again. The nodes have been unprojected, but the edges have not
# additionally, the nodes geodataframe is now invalid (crs does not match coordinates)
nodes, edges = ox.graph_to_gdfs(network)
```

gboeing:
resolves #677

### DIFF
--- a/osmnx/utils_graph.py
+++ b/osmnx/utils_graph.py
@@ -1,6 +1,7 @@
 """Graph utility functions."""
 
 import itertools
+import warnings
 from collections import Counter
 
 import geopandas as gpd
@@ -36,7 +37,9 @@ def graph_to_gdfs(G, nodes=True, edges=True, node_geometry=True, fill_edge_geome
     Returns
     -------
     geopandas.GeoDataFrame or tuple
-        gdf_nodes or gdf_edges or tuple of (gdf_nodes, gdf_edges)
+        gdf_nodes or gdf_edges or tuple of (gdf_nodes, gdf_edges). gdf_nodes
+        is indexed by osmid and gdf_edges is multi-indexed by u, v, key
+        following normal MultiDiGraph structure.
     """
     crs = G.graph["crs"]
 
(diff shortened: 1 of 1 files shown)


osmnx/utils_graph.py
```diff
@@ -129,6 +136,20 @@ def graph_from_gdfs(gdf_nodes, gdf_edges, graph_attrs=None):
...
     G : networkx.MultiDiGraph
     """
+    if not ("x" in gdf_nodes.columns and "y" in gdf_nodes.columns):
+        raise ValueError("gdf_nodes must contain x and y columns")
+
+    # if gdf_nodes has a geometry attribute set, drop that column (as we use x
+    # and y for geometry information) and warn the user if the geometry coords
+    # differ from those in the x and y columns
+    if hasattr(gdf_nodes, "geometry"):
+        if (gdf_nodes.geometry.x != gdf_nodes["x"]).all() or (
```
adelarue, line 146: If for some reason the geometry column contains something other than `Point` objects, access to `geometry.x` will raise
```python
ValueError: x attribute access only provided for Point geometries
```
Do you want to add your own check that every node geometry is indeed a Point, or maybe update the documentation to say that the geometry column in the nodes data frame will be discarded, and further will throw an error if it contains anything other than single points?

gboeing: Thanks. I see that ValueError as an example of "the geometry values do not match the x and y columns' values". Rather than handle it separately (or just let it error), I've revised the PR in ae3e4b2 to show the warning if it occurs. I think this clearer warning + documentation should help users engaging in an edge case adjust appropriately.  The key thing here would be prevent "silent" weird behavior, as you documented in #677: as long as there's a warning (or at least an error) instead of silently corrupting the data structures.

adelarue: Thanks! Looks good to me 👍

PR #683, opened 2021-03-23
Adjust formatting of inline code block.

maxerickson:
**Read these instructions carefully**

Before you proceed, review the contributing guidelines in this repo's CONTRIBUTING.md file, especially the sections on project coding standards and tests.

In this pull request, please include:

  - a reference to related issue(s)
  - a description of the changes proposed in the pull request
  - an example code snippet illustrating usage of the new functionality

No related issue, someone on OSM US Slack asked about a query not working and it turned out that a unicode doublequote copied from the documentation was the issue.

### DIFF
--- a/osmnx/utils.py
+++ b/osmnx/utils.py
@@ -187,7 +187,7 @@ def config(
     overpass_settings : string
         Settings string for overpass queries. For example, to query historical
         OSM data as of a certain date:
-        ``[out:json][timeout:90][date:"2019-10-28T19:20:00Z"]``.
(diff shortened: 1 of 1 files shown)


gboeing:
Hi @maxerickson. Thanks for your suggestion. Could you clarify for me what you're fixing here though? The original code appears to be correct. The single backticks denote inline python code, and the single quote marks within that inline code denote a string literal. Changing it as you've proposed would not make sense in this context: you've replace the single quotes with additional backticks, so it no longer denotes a string literal as it should.

maxerickson:
In that case it should likely be double backtick+single quote, rst uses double backticks for inline code blocks.

PR #690, opened 2021-04-01
add graph node elevations from local DEM raster file(s)

gboeing:
Inspired by @kylebarron's work in #351, I'm exploring how to incorporate functionality to add graph node elevations from local DEM raster file(s). This PR aims to:

  - add a new `add_node_elevations_raster` function as a lightweight solution for DEM querying
  - offer optional multiprocessing
  - introduce only optional dependencies (rasterio and gdal)
  - replace the old `add_node_elevations` function with a new `add_node_elevations_google` function (and deprecate it and the old "airmap" provider)

cc @janikmu who expressed interest in this feature in the past.

In preliminary testing, multiprocessing offers >1 order of magnitude speed-up when working with large graphs (>200,000 nodes). My initial testing has this working with the complete ASTER v3, SRTM 30, and SRTM 90 v4.1 DEMs, totalling ~60,000 individual files and ~800 GB. As a frame of reference, using those rasters it can add elevations to every node in a mid-sized city like Oakland CA in ~1 second and in a large graph like LA county (unsimplified, so node count >2,200,000) in ~90 seconds.


gboeing:
Tests currently failing due to https://github.com/networkx/networkx/issues/4718

janikmu:
First of all, thanks for the Implementation of local raster files, but I wonder why you simultaneously removed/deprecated the alternative Airmap provider alongside Google?

gboeing:
@janikmu a few reasons... 1) the key need of having a free alternative to google is now satisfied with the raster option; 2) streamline the codebase and maximize its maintainability; 3) simplify the code, which is currently a bit convoluted as the function contains lots of branching logic to handle different providers. Now everything will be simpler and a free alternative will be maintained going forward.

PR #699, opened 2021-04-28
Resolve Overpass API requests to a specific IP address

### LINKED ISSUE #698: Overpass status endpoint provides incorrect results
I've been noticing mysteriously long pauses required by Overpass's status endpoint recently. There may be an explanation now... it's possible that the Overpass status endpoint provides "incorrect" results due to load balancing/redirection among multiple subdomains now. In #697 @mmd-osm [says](https://github.com/gboeing/osmnx/issues/697#issuecomment-827793957):

> /api/status may show incorrect results for overpass-api.de, as there are really two independent servers behind this URL. However, /api/status always reflects the status of a single server only. There's currently no workaround for it, unless you're specifically targeting one of the two machines.

and 

> It's a bit unfortunate that /api/status hasn't been designed in a way to handle multiple servers in a consistent way. Due to this design issue, it's hard to recommend some best practice. At this time, I wouldn't hardcode any of the subdomains, as they might change over time: new servers might get added, or existing ones decommissioned and replaced by bigger machines.
> 
> I don't know if it's feasible for osmnx to decouple the "overpass-api.de" name server lookup from sending the actual request to the server. As an example, curl offers the command line option --resolve to influence name resolving, thereby directing the request to one specific IP addresses. By using the same IP address for both queries and /api/status, you should get a consistent picture.
> 
> The clear downside of this approach is that you would essentially need to re-implement DNS round robin, to make sure a request uses another server in case of an issue.

So, my understanding is that the domain overpass-api.de just redirects to one of its subdomains (currently z.overpass-api.de and lz4.overpass-api.de). So if we check the status endpoint of overpass-api.de, we may see results for subdomain z, but when we submit the query itself it gets redirected to subdomain lz4. This could result in violating the slot management timing for that server, and potentially a long delay before the next query is allowed. This may explain the particularly long delay times between queries I've seen in recent months.

I believe we can do the decoupling suggested by @mmd-osm in Python with the `socket` module.

```python
import socket
socket.gethostbyname("localhost")
# returns 127.0.0.1
```

For example if we run:
```python
domains = ["overpass-api.de", "z.overpass-api.de", "lz4.overpass-api.de"]
ips = [socket.gethostbyname(d) for d in domains]
for d, ip in zip(domains, ips):
    print(d, "\t", ip)
```

we see:
```
overpass-api.de 	 178.63.11.215
z.overpass-api.de 	 178.63.11.215
lz4.overpass-api.de 	 178.63.48.217
```
but if I change my computer's IP address and run it again, we see:
```
overpass-api.de 	 178.63.48.217
z.overpass-api.de 	 178.63.11.215
lz4.overpass-api.de 	 178.63.48.217
```

So, it looks like `socket.gethostbyname(domain)` does return the IP address of one of its subdomains when you pass in "overpass-api.de". We could then submit all Overpass API requests directly to the IP address resolved by `socket.gethostbyname(domain)`. Something like:

```python
import requests
overpass_endpoint = "http://overpass-api.de/api"
start = overpass_endpoint.find('//') + 2
end = overpass_endpoint[start:].find('/')
domain = overpass_endpoint[start:start+end]
ip = socket.gethostbyname(domain)
overpass_endpoint = overpass_endpoint.replace(domain, ip)
print(domain, ip, overpass_endpoint)
print(requests.get(overpass_endpoint + "/status").text)
```
which prints:
```
overpass-api.de 178.63.48.217 http://178.63.48.217/api
Connected as: 2333923515
Current time: 2021-04-28T18:09:53Z
Rate limit: 2
2 slots available now.
Currently running queries (pid, space limit, time limit, start time):
```

This *seems* to work. I'll try to put together a PR. @mmd-osm let me know if I'm making any naive logical errors here.

gboeing:
Fixes #698 when making Overpass requests by resolving a URL to its IP address then forcing `socket.getaddrinfo` to use that IP address instead of the host under the hood of the `requests` library. Works for http and https requests.

### DIFF
--- a/osmnx/downloader.py
+++ b/osmnx/downloader.py
@@ -4,10 +4,12 @@ import datetime as dt
 import json
 import logging as lg
 import re
+import socket
 import time
 from collections import OrderedDict
 from hashlib import sha1
 from pathlib import Path
+from urllib.parse import urlparse
 
 import numpy as np
 import requests
@@ -245,7 +247,26 @@ def _get_http_headers(user_agent=None, referer=None, accept_language=None):
     return headers
 
 
-def _get_pause(recursive_delay=5, default_duration=60):
+def _resolve_ip(url):
+    """
+    Returns the URL with its domain resolved to an IP address.
+
+    Parameters
+    ----------
+    url : string
+        URL to resolve the domain of
+
+    Returns
+    -------
+    url : string
+        the URL with its domain replaced by its IP address
+    """
+    domain = urlparse(url).netloc
+    ip = socket.gethostbyname(domain)
+    return url.replace(domain, ip)
+
+
+def _get_pause(base_endpoint, recursive_delay=5, default_duration=60):
     """
     Get a pause duration from the Overpass API status endpoint.
 
@@ -254,6 +275,8 @@ def _get_pause(recursive_delay=5, default_duration=60):
 
     Parameters
     ----------
+    base_endpoint : string
+        base Overpass API endpoint (without "/status" at the end)
     recursive_delay : int
         how long to wait between recursive calls if the server is currently
         running a query
@@ -265,7 +288,7 @@ def _get_pause(recursive_delay=5, default_duration=60):
     pause : int
     """
     try:
-        url = settings.overpass_endpoint.rstrip("/") + "/status"
+        url = base_endpoint.rstrip("/") + "/status"
         response = requests.get(url, headers=_get_http_headers())
         status = response.text.split("\n")[3]
         status_first_token = status.split(" ")[0]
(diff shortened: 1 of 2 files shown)


jslorrma:
> Fixes #698 when making Overpass requests by resolving a URL to its IP address then forcing `socket.getaddrinfo` to use that IP address instead of the host under the hood of the `requests` library. Works for http and https requests.

This pull request breaks any download in case proxies are or have to be used, for e.g. if one is using osmnx from within a corporate network.  `socket.gethostbyname` can only resolve the URL's domain to an IP address for hosts within the corporate network. Trying to resolve an external URL raises the following error: `gaierror: [Errno 11001] getaddrinfo failed`

PR #701, opened 2021-04-29
added route_linewidth additional parameter #700

### LINKED ISSUE #700: add route_linewidth as string or list parameter in osmnx.plot.plot_graph_routes()
It would be useful to pass a list of linewidths in the  plot_graph_routes api so that we can render dynamically not just the color but also the size of each route. i.e. for plotting origins-destinations data this allow to have different route tickness based on flow size of each route

groundrace:
add in plot_graph_routes the possibility to pass route_linewidth as list of different linewidth one for each route to be plotted
this fix #700

### DIFF
--- a/osmnx/plot.py
+++ b/osmnx/plot.py
@@ -362,7 +362,7 @@ def plot_graph_routes(G, routes, route_colors="r", route_linewidth=4, **pgr_kwar
         G,
         route=routes[0],
         route_color=route_colors[0],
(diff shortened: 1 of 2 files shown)


gboeing:
@groundrace the tests are failing. See [here](https://github.com/gboeing/osmnx/tree/main/tests) for guidance.

gboeing:
Thanks @groundrace

PR #704, opened 2021-05-09
allow disabling slot management pausing

### LINKED ISSUE #697: When runnning a custom overpass api with no slot management, the check in downloader._get_pause for "Currently" ends up in an infinite loop.
When runnning a custom overpass api with no slot management, the check in downloader._get_pause for "Currently" ends up in an infinite loop.

overpass/api/status will return "Currently running queries (pid, space limit, time limit, start time):" But with no slots beneath it. Therefore checking for "Currently" isn't sufficient. 

I helped myself by hardcoding the pause-parameter in _osm_network_download to 0, but that is not really a solution...

gboeing:
fixes #697 

This PR allows the user to configure whether OSMnx should check the overpass server status endpoint for how long to pause before making the next request. That check is necessary if the server uses slot management, but the setting can (should) be disabled if you are running your own overpass instance.


gboeing:
@jannefleischer would you like to test this proposed fix to your issue #697? I appreciate it!

PR #713, opened 2021-05-25
fix typo in documentation

salcc:
There is a typo in the documentation of the add_edge_bearings function from the bearing module:

> geodesic line from **from** the origin node


gboeing:
@salcc thanks!

Issue #717, opened 2021-06-08
Improve speed and travel time calculation (when simplified)


EwoutH:
**Is your feature proposal related to a problem?**

Currently speed and travel time aren't that accurate, especially if a graph is simplified before the data is added.

**Describe the solution you'd like to propose**

A subset of the alternatives below.

**Describe alternatives you've considered**

- [ ] Don't discard identical travel time values when simplifying, but keep them all to be able to calculate a sum. [Doable](https://github.com/gboeing/osmnx/compare/main...EwoutH:patch-2).
- [ ] Add variables for the total travel time and average speed and calculate those before simplifying. [Also doable](https://github.com/gboeing/osmnx/compare/main...EwoutH:patch-1).
- [ ] When simplified, don't use only the first `max_speed` value but take a weighted average based on road length. See [speed.py#L59](https://github.com/gboeing/osmnx/blob/5e3afe671fcfa4d2f54b0905c825ebfd6468631b/osmnx/speed.py#L59).

**Additional context**

I'm currently working on building a graph of the road network of Helsinki, and encountered these problems.

gboeing:
Thanks @EwoutH. I generally like the idea of your suggestions here. The current logic in OSMnx is essentially: if you want precise speed/travel time handling, use an unsimplified graph. It may be possible to provide greater accuracy without too much complexity in the codebase. I see you've got one PR open. Would you like to contribute this enhancement? What specific subset of the alternatives do you propose?

gboeing:
One thing that would be useful would be to flesh out the workflow of the use case you're envisioning. What's the use case and its order of operations? I'm guessing you're doing something like creating an unsimplified graph, adding speeds and travel times, and then simplifying it.

gboeing:
Hi @EwoutH I'm checking to see if you're still actively developing these enhancements you had proposed, or if this is worth closing for now.

PR #718, opened 2021-06-10
simplification.py: Keep all travel_time values

EwoutH:
Currently when multiple edges are merged together, all identical attribute values except length are discarded. While this is logical for most attributes like street names and road types, it doesn't make much sense for the travel time.

This commit adds an exception for the travel_time attribute that makes sure identical values aren't discarded. By keeping all travel time values, the total travel time can be calculated later.

Part of #717.

### DIFF
--- a/osmnx/simplification.py
+++ b/osmnx/simplification.py
@@ -266,7 +266,8 @@ def simplify_graph(G, strict=True, remove_rings=True):
                 # if there's only 1 unique value in this attribute list,
                 # consolidate it to the single value (the zero-th)
                 edge_attributes[key] = edge_attributes[key][0]
-            elif key != "length":
+            # also keep all travel_time values, to be able to sum them later
+            elif key not in ["length", "travel_time"]:
                 # otherwise, if there are multiple values, keep one of each value
                 edge_attributes[key] = list(set(edge_attributes[key]))
 
--- a/osmnx/simplification.py
+++ b/osmnx/simplification.py
@@ -267,7 +267,7 @@ def simplify_graph(G, strict=True, remove_rings=True):
                 # consolidate it to the single value (the zero-th)
                 edge_attributes[key] = edge_attributes[key][0]
             # also keep all travel_time values, to be able to sum them later
-            elif key not in ["length", "travel_time"]:
+            elif key not in {"length", "travel_time"}:
                 # otherwise, if there are multiple values, keep one of each value
                 edge_attributes[key] = list(set(edge_attributes[key]))
 


gboeing:
It looks like you're currently retaining `travel_time` but not summing it later, as happens with `length` [here](https://github.com/EwoutH/osmnx/blob/travel-time/osmnx/simplification.py#L278).

EwoutH:
> It looks like you're currently retaining `travel_time` but not summing it later, as happens with `length` [here](https://github.com/EwoutH/osmnx/blob/travel-time/osmnx/simplification.py#L278).

Yes that is correct, it keeps it as a list that can be summed later. We could also sum it in the function itself and return a single value, do you have a preference?

osmnx/simplification.py
```diff
@@ -266,7 +266,8 @@ def simplify_graph(G, strict=True, remove_rings=True):
                 # if there's only 1 unique value in this attribute list,
                 # consolidate it to the single value (the zero-th)
                 edge_attributes[key] = edge_attributes[key][0]
-            elif key != "length":
+            # also keep all travel_time values, to be able to sum them later
+            elif key not in ["length", "travel_time"]:
```
gboeing, line 270:
```suggestion
            elif key not in {"length", "travel_time"}:
```

gboeing:
> We could also sum it in the function itself and return a single value

Yes for consistency it should probably be handled the same way `length` is handled now.

EwoutH:
Using a set instead of a list is a good idea, thanks!

> Yes for consistency it should probably be handled the same way `length` is handled now.

I agree. Almost added a quick commit, but then thought we can't assume the `travel_time` attribute is always present and it got a bit messy. If you have a preferred implementation feel free to add it, I will take a further look on Monday.

gboeing:
I would suspect that a simple check could address that, like:

```python
if attr_name in edge_attributes:
    edge_attributes[attr_name] = sum(edge_attributes[attr_name])
```

EwoutH:
Had a moment to work on it which resulted in an implementation I'm quite satisfied with. It's in [this commit](https://github.com/EwoutH/osmnx/commit/1b329e169435375c6211eabfea0864e59b7f0b7b?diff=split).

For each key present, it first checks it's only a single value, then checks if it needs to be summed, and if both false it returns just the unique set as a list. By checking if the `key` is in set `summed_attributes` and not the other way around, it avoids the possibility of an attribute not being preset leading to an error.

Leave any remarks and if you like this way of implementing it, l will open a new pull request (to avoid squashing this branch and deleting history).

gboeing:
I left you a couple comments there.

EwoutH:
Thanks! I doubted which order the `if` and `elif` statements best could be arranged, but missed that the length of the set of the list and not the list itself was used.

I opened PR #721 with this implementation including both your suggestions.

PR #721, opened 2021-06-14
simplification.py: Refactor list to value/set and summation

EwoutH:
This commit refactors the part of the simplification loop that converts lists to lists with unique values and single values. It allows a set of certain edge attributes to added up instead, currently including `length` and `travel_time`.

Issue: #717
Succeeds this PR: #718

### DIFF
--- a/osmnx/simplification.py
+++ b/osmnx/simplification.py
@@ -225,7 +225,7 @@ def simplify_graph(G, strict=True, remove_rings=True):
         raise Exception("This graph has already been simplified, cannot simplify it again.")
 
     utils.log("Begin topologically simplifying the graph...")
-    
+
     # define edge segment attributes to sum upon edge simplification
     attrs_to_sum = {"length", "travel_time"}
 
--- a/osmnx/simplification.py
+++ b/osmnx/simplification.py
@@ -272,7 +272,7 @@ def simplify_graph(G, strict=True, remove_rings=True):
             elif len(set(edge_attributes[key])) == 1:
                 edge_attributes[key] = edge_attributes[key][0]
             # otherwise, if there are multiple values, keep one of each value
-            else
+            else:
                 edge_attributes[key] = list(set(edge_attributes[key]))
 
         # construct the geometry and sum the lengths of the segments
--- a/osmnx/simplification.py
+++ b/osmnx/simplification.py
@@ -260,21 +260,25 @@ def simplify_graph(G, strict=True, remove_rings=True):
                     # containing the one value
                     edge_attributes[key] = [edge[key]]
 
+        # define attributes that need to be added up on simplification
+        summed_attributes = {"length", "travel_time"}
+
         for key in edge_attributes:
-            # don't touch the length attribute, we'll sum it at the end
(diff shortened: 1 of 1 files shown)


gboeing:
Tests are failing.

EwoutH:
Missed a colon after `else`, fixed it and now the [CI passes](https://github.com/EwoutH/osmnx/actions/runs/938593227) on my branch. Should have catched that earlier, sorry.

gboeing:
I did a little code clean up and variable renaming in 08361cc as I realized some of the code clarity could be improved.

gboeing (approved):

EwoutH:
I would like to do some more work on improving simplification. One of my goals is that we can get a graph without any lists in the dataframes. Which can be done in two ways:
- Combine the list of attributes to a single one
- Don't merge the segments

Therefore, I would like to propose two new input arguments:
1. `merge_all_attributes` (default `False`). This boolean enforces to merge all attributes to a single value. My current approach would be to take the most occurring value, based on the length of the road segments.
So for example, if I have 3 segments each of length 100, and the first two have 2 lanes and the last one 1, it will assume 2 lanes for the simplified edge. However, it the last edge has a length of >200, it will assume 1 lane for the simplified segment.
2. `exclude_attributes` (default `None`). This can be a string of list of strings, like `["highway", "maxspeed"]`, and if these values differ, it will not simplify those edges.

Combining these two, you could say for example:
```Python
simple_graph = ox.simplify_graph(graph, simplify(merge_all_attributes=True, exclude_attributes=["highway", "maxspeed"])
```
This will ensure that all edges have only a single value for all attributes, but the highway and maxspeed attributes are always of the original type.

When this is done, we could make more specific smart aggregations where they make sense, like done in this PR by summing length and travel_time.
```Python
# define edge segment attributes to sum upon edge simplification
attrs_to_sum = {"length", "travel_time"}
```

gboeing:
Regarding:

> 2. exclude_attributes (default None). This can be a string of list of strings, like ["highway", "maxspeed"], and if these values differ, it will not simplify those edges.

See #1117.

gboeing:
Thanks for suggesting this. Let's discuss first. I'll have some time to consider this in detail around 2 weeks from now. There may be a way to make this useful to a general audience, but I am currently (though superficially) skeptical that collapsing multiple values to a single value (or summing) would work as a universal solution. At a minimum, it would have ramifications elsewhere in the package.

EwoutH:
I thought about it some more. It might be useful to have a more generic `merge_attributes` input argument. Then you could specify which attributes you want to merge, and how you want to do that. For example: `{'lanes': 'most_frequent', 'maxspeed': 'mean'}`. Aside from a few built-in methods (like most_frequent, mean, min), we could also allow users to use custom functions. 

Furthermore, it might still be interesting to allow weighting over edge length, so that longer edges have a larger influence than shorter ones. Maybe that could be an boolean input like `length_as_weight`.

I thought about making it a separate function, but one problem you encounter is that you already lose information on simplification, like the length being squashed to a single number. So doing it while simplifying seems the most logical option.

(it might be good to open a new issue to continue this discussion)

PR #725, opened 2021-06-18
allow user-defined aggregation function when imputing missing edge speeds

### LINKED ISSUE #722: Median instead of mean when guessing highway speeds
While analyzing highway speeds, I noticed that osmnx relies on existing values for speed to fill in missing speeds per highway type. But it is using `mean`, while I think `median` would be more beneficial, correct and round, as it will predict actual speed in a given country better.

### Impl:
I am thinking about changing this line only:
https://github.com/gboeing/osmnx/blob/9dbee4980fe7a43279b0bfa823b6b40317705c30/osmnx/speed.py#L87

(and accompanying comments). I tried running tests after changing this line and they are passing, but I still think this is behavioral change and can break some assumption on client code, not sure what are your thoughts (is it worth it). This is main reason I didn't submit PR, but opened issue first.

Another possible problem with this approach - few data points can lead to worse results (think of 2 speeds of 30kph and 50kph would produce 30, instead of maybe, more correct 40kph), but I think osmnx users generally work with larger amount of data than 2 or 3 rows and that this is not a problem.

### Example

I did this with streets from Belgrade, Serbia (funny fact - Belgrade is mentioned in @gboeing's page [here](https://geoffboeing.com/2016/11/osmnx-python-street-networks/))

This is with `mean`:

slika

And with `median`:
slika

Here are raw values from these pictures for better searching and copy/pasting:

| type | speed (mean) |
| ---------------| ---------|
| living_street | 21.25 |
| motorway | 95.19767441860465 |
| motorway_link | 48.58974358974359 |
| primary | 52.78801843317972 |
| primary_link | 46.59574468085106 |
| residential | 37.29013539651837 |
| secondary | 44.09898305084746 |
| secondary_link | 42.916666666666664 |
| tertiary | 42.27095516569201 |
| tertiary_link | 49.25925925925926 |
| trunk | 65.625 |
| trunk_link | 50.0 |
| unclassified | 42.64705882352941 |

| type | speed (median) |
| ---------------| ---------|
| living_street | 10.0 |
| motorway | 91.0 |
| motorway_link | 50.0 |
| primary | 50.0 |
| primary_link | 50.0 |
| residential | 40.0 |
| secondary | 40.0 |
| secondary_link | 50.0 |
| tertiary | 40.0 |
| tertiary_link | 50.0 |
| trunk | 60.0 |
| trunk_link | 60.0 |
| unclassified | 45.0 |

gboeing:
resolves #722

### DIFF
--- a/osmnx/speed.py
+++ b/osmnx/speed.py
@@ -13,15 +13,16 @@ def add_edge_speeds(G, hwy_speeds=None, fallback=None, precision=1, agg=np.mean)
     """
     Add edge speeds (km per hour) to graph as new `speed_kph` edge attributes.
 
-    Imputes free-flow travel speeds for all edges based on mean `maxspeed`
-    value of edges, per highway type. For highway types in graph that have no
-    `maxspeed` value on any edge, function assigns the mean of all `maxspeed`
-    values in graph.
+    By default, this imputes free-flow travel speeds for all edges via the
+    mean `maxspeed` value of the edges of each highway type. For highway types
+    in the graph that have no `maxspeed` value on any edge, it assigns the
+    mean of all `maxspeed` values in graph.
 
-    This mean-imputation can obviously be imprecise, and the user can override
-    it by passing in `hwy_speeds` and/or `fallback` arguments that correspond
-    to local speed limit standards. The user can also specify a different
-    aggregation function to impute missing values from the observed values.
+    This default mean-imputation can obviously be imprecise, and the user can
+    override it by passing in `hwy_speeds` and/or `fallback` arguments that
+    correspond to local speed limit standards. The user can also specify a
+    different aggregation function (such as the median) to impute missing
+    values from the observed values.
 
     If edge `maxspeed` attribute has "mph" in it, value will automatically be
     converted from miles per hour to km per hour. Any other speed units should
@@ -47,7 +48,9 @@ def add_edge_speeds(G, hwy_speeds=None, fallback=None, precision=1, agg=np.mean)
     precision : int
         decimal precision to round speed_kph
     agg : function
-        aggregation function to impute missing values from observed values
+        aggregation function to impute missing values from observed values.
+        the default is numpy.mean, but you might also consider for example
+        numpy.median, numpy.nanmedian, or your own custom function
 
     Returns
     -------
--- a/osmnx/speed.py
+++ b/osmnx/speed.py
@@ -9,7 +9,7 @@ import pandas as pd
 from . import utils_graph
 
 
-def add_edge_speeds(G, hwy_speeds=None, fallback=None, precision=1):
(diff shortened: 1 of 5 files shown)


osmnx/speed.py
```diff
@@ -45,6 +46,8 @@ def add_edge_speeds(G, hwy_speeds=None, fallback=None, precision=1):
         values on any edge
     precision : int
         decimal precision to round speed_kph
+    agg : function
+        aggregation function to impute missing values from observed values
```
stalker314314, line 50: Consider explicitly mentioning `np.median` as example of different input. Maybe "aggregation function to impute missing values from observed values. Use np.median if you are trying to impute official values on highways" or something along those lines.

Rationale: np.mean and np.median are only agg function that comes to my mind to be useful here. If so, other one should maybe have honorary mention:)

osmnx/speed.py
```diff
@@ -84,12 +87,12 @@ def add_edge_speeds(G, hwy_speeds=None, fallback=None, precision=1):
     # highway type
     for hwy, group in edges.groupby("highway"):
         if hwy not in hwy_speed_avg:
-            hwy_speed_avg.loc[hwy] = group["speed_kph"].mean()
+            hwy_speed_avg.loc[hwy] = agg(group["speed_kph"])
```
stalker314314, line 90: I am not getting proper behavior here. My highways are filled with nans (I guess as anywhere in the world). But when I use `np.median`, I get result as nan! It seems it is not the same if one use `np.median(series)` or `series.median()`. I am no expert on numpy, but this is what I am observing. Here is simple repro: https://gist.github.com/stalker314314/45ae826371a5b88cbdc76492deeee508

I will try to explore more in coming days!

stalker314314: Seems like `Series.median()` function is skipping nans by default (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.median.html?highlight=median#pandas.Series.median), while `numpy.median` does not:) Who could tell:)

Solution that come to my mind could be to either:
1. use `np.nanmedian` (but I would never thought of this unless doc tell me that), or
2. remove nans before calling agg function (but this might limit usability of agg - what if caller want to aggregate with nans!?)

stalker314314: And BTW, using `np.nanmedian` does the job and your PR is perfectly correct (if you consider that as a good enough)

gboeing: Thanks for the feedback. Worth noting that as this argument is a function, the user can pass any function they want. They're not constrained to pre-built numpy or pandas aggregations as they can write their own custom function with any special logic they'd like. But I'll add a note to the docstring pointing the user toward `np.median` and `np.nanmedian` as a couple options they might consider.

gboeing: Resolved in f645225

osmnx/speed.py
gboeing: Resolved in f645225

Issue #17, opened 2016-12-12
Certificate verify failed


tetekos:
Hello all,

I am new here thus excuse me if I am asking something simple.

I have been trying to get osmnx going with that:
import osmnx as ox
city = ox.gdf_from_place('Berkeley, CA')
ox.plot_shape(ox.project_gdf(city))

But I am getting this output error all the time:
**[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590)**

Please could you suggest any solutions cause the ones I have seen so far are not working.

Thanks a lot.

gboeing:
What is your OS, architecture, Python version, OSMnx version, etc? Are your Python packages up-to-date, especially `requests` and `openssl`?

gboeing:
@tetekos checking back once again on this open issue... see previous comment.

gboeing:
Closing this issue now. If this problem crops up again, please add additional details and we can chase it down.

MonoPaul:
Getting that same issue here - client is using a custom SSL certificate.
I can connect using requests directly by either pointing to the local cert, or by passing the parameter cert=False to the get request.
Any chance you can add a parameter that gets passed to general https requests ?

MonoPaul:
Ah - issue occurs on Win 7 & Jupyter running on Debian Mint, on both python 2.7.13 & 3.5.
I am using OSMnx v 0.5, and requests & openssl are all up to date :-)

gboeing:
@MonoPaul can you provide me a working example of code to reproduce this, plus the error's full traceback message?

MonoPaul:
The full traceback'll have to wait till Monday - my home SSL works just fine. The code I used was simply a regular request : 
import osmnx as ox
G = ox.graph_from_place('Manhattan, New York, USA')
I'm guessing that works under normal circumstances - what I'd be requesting would be more of a feature request than a bug report : an option to either pass a specific SSL cert or disable SSL (ie passing the cert=('/path/to/cert') in the requests.get)
Let me know if you'd rather have me branch instead ;-)

gboeing:
Thanks for that info!

[...]

PR #738, opened 2021-08-02
Use Session for HTTP requests

hokieg3n1us:
This proposed change would resolve any issues similar to [issue #17 ](https://github.com/gboeing/osmnx/issues/17).

Description: Include a dictionary of configuration for a requests Session that is instantiated by the downloader, including keys for auth, cert, and verify. This allows a user to modify that session by providing a tuple of username and password for basic authentication, or provide client and server certificates. It should also improve performance when making repeated requests to the same host by re-using the underlying urllib3 connection.

Example Usage:

```
import osmnx as ox

session_config = {}
session_config['auth'] = ('username', 'password')
session_config['cert'] = '/path/to/cert'
session_config['verify'] = '/path/to/CA/'

osmnx.config(session_config=session_config)
```


gboeing:
Thanks @hokieg3n1us let me think about this. Two questions for you:

  1. Would `requests.Session` have any conflicts/redundancies with the new explicit IP address resolution fixes released in v1.1.1 (#699)?
  2. Could you add more detail about why/how a user would use this in the [docstring](https://github.com/gboeing/osmnx/pull/738/files#diff-d1d494915afc629c524967682a123e6bf3d316a6f7b2b6c7ac780f698d055ef0R208-R209)?

hokieg3n1us:
@gboeing 

1. You're correct that it should be redundant. The requests.Session will create a persistent urllib3 connection that will be re-used for requests to the same host/port, so I backed out the _config_dns function. 
2. I updated the docstring. Maybe we should link to Requests documentation for [Advanced Usage](https://docs.python-requests.org/en/master/user/advanced/) to describe the customization that can be done on the session.

hokieg3n1us:
@gboeing

I diagnosed the intermittent failures (inactivity on the underlying TCP connection causing the server to close), and refactored significantly to still gain the benefits of using a requests Session, while providing some of the more important configuration options to a user. More of the underlying Session parameters can be exposed as necessary.

osmnx/downloader.py
```diff
@@ -670,14 +679,14 @@ def overpass_request(data, pause=None, error_pause=60):
     else:
         # if this URL is not already in the cache, pause, then request it
         if pause is None:
-            this_pause = _get_pause(base_endpoint)
-        utils.log(f"Pausing {this_pause} seconds before making HTTP POST request")
-        time.sleep(this_pause)
+            this_pause = _get_pause(session, base_endpoint)
+            utils.log(f"Pausing {this_pause} seconds before making HTTP POST request")
+            time.sleep(this_pause)
```
gboeing, line 684: These need to be moved back out of the `if` block.

osmnx/utils.py
```diff
@@ -205,10 +205,13 @@ def config(
...
-        session used by downloader for making HTTP requests to external APIs
-        override to provide proxies, client/server SSL certificates, or
-        basic authentication using username and password
+    session_config : dict
+        dictionary that supports overriding configuration for requests Session
+        to be used to during external API requests. currently supports auth, cert,
+        and verify as keys. auth is a tuple of (username, password). cert is the
+        path to a single file (containing the private key and certificate) or a tuple
+        of both files' paths. verify can be the path to a CA_BUNDLE file, or directory
+        of trusted CAs.
```
gboeing, line 214: It might be easier and more future-proof to just point the user toward the requests documentation on requests.Session.auth, cert, and path since we're just passing these values through to those requests.Session attributes.

osmnx/downloader.py
```diff
@@ -275,7 +278,7 @@ def _get_pause(base_endpoint, recursive_delay=5, default_duration=60):
 
     try:
         url = base_endpoint.rstrip("/") + "/status"
-        response = settings.session.get(url, headers=_get_http_headers())
+        response = session.get(url, headers=_get_http_headers())
```
gboeing, line 281: `_create_session` is not called in the `_get_pause` function, but it is called in the other functions that make HTTP requests to the Overpass server?

osmnx/downloader.py
```diff
@@ -451,6 +454,8 @@ def _osm_network_download(polygon, network_type, custom_filter):
     else:
         osm_filter = _get_osm_filter(network_type)
 
+    session = _create_session()
```
gboeing, line 457: I notice that `_create_session` is called 3 times (i.e. separately in the different "download" functions). Is there a reason to do it this way, rather than creating a session one time, for example whenever `ox.config` is called, and then just using that session throughout?

gboeing (changes requested):
I've left just a few comments. My main overall question would be: have you tested this to confirm consistent IP address resolution in line with what #699 provided? If so, could you document those test methods/results here? Thanks!

osmnx/downloader.py
hokieg3n1us: That produces an out of scope variable reference to get_pause similar to the recently fixed issue for the response status code.

hokieg3n1us:
I am closing this pull request because I cannot reproduce the intermittent issues causing the remote server to close the connection, and that's difficult to debug properly without more information about OSM's service layer. It's also difficult to elegantly construct a session, without refactoring the downloader.py to be instantiated classes that live/die during graph construction since you can't create a single Session to be used for the entirety of osmnx, because the client can't hold open an inactive TCP connection.

I plan to identify a route of cleanly passing cert, auth, and verify directly to the requests.post/requests.get.

gboeing:
New PR for development at #745.

Issue #741, opened 2021-08-08
Improve project_gdf performance


cheginit:
**Is your feature proposal related to a problem?**

No.

**Describe the solution you'd like to propose**

For UTM projection in [`project_gdf`](https://github.com/gboeing/osmnx/blob/cd404ae2f153a5918a924b033e0e30700f156f99/osmnx/projection.py#L109) function, `unary_union` is being used which is a computationally expensive operation. Since for finding UTM there's no need for exact location of the centroid, I propose to use [`representative_point`](https://geopandas.org/docs/reference/api/geopandas.GeoSeries.representative_point.html#geopandas.GeoSeries.representative_point) function of GeoPandas to speed up the process. This is the function description:
> Returns a GeoSeries of (cheaply computed) points that are guaranteed to be within each geometry.

**Additional context**

This is the change that I am proposing:

Replace:
```python
avg_lng = gdf["geometry"].unary_union.centroid.x
```

with:
```python
avg_lng = gdf["geometry"].representative_point().x.mean()
```

An example for comparing their performance:

```python
import osmnx as ox

street = ox.graph_from_bbox(west=-96.0255, south=29.421, east=-94.782, north=30.512)
street = ox.graph_to_gdfs(
    street,
    nodes=False,
    node_geometry=False,
    fill_edge_geometry=True,
)
```

![image](https://user-images.githubusercontent.com/13016644/128642173-947ad2d7-bb4b-446b-8e5b-5b1baf08b783.png)

We can clearly see the performance gain from this change.

gboeing:
This is a clever suggestion. Yes, I think `representative_point` is good enough for auto-determining a UTM zone. Would you like to open a simple PR to contribute this performance enhancement?

cheginit:
Sure, would be happy to.

cheginit:
Also, I found [this](https://github.com/Turbo87/utm/blob/40eb34c86895bf3a5f97b5819b9da4b164151d3c/utm/conversion.py#L301-L323) for computing zone number, do you want to follow that logic or prefer to keep it as is?

gboeing:
> do you want to follow that logic or prefer to keep it as is?

Let's keep it simple and consistent for now. This might be worth revisiting in the future.

gboeing:
Closed by #743.

PR #743, opened 2021-08-09
ENH: Improve performance of computing UTM zone number

cheginit:
Following the discussion in #741, this PR uses an approximate method for computing longitude of a GeoDataFrame to obtain UTM zone number.

### DIFF
--- a/osmnx/projection.py
+++ b/osmnx/projection.py
@@ -106,7 +106,7 @@ def project_gdf(gdf, to_crs=None, to_latlong=False):
             raise ValueError("Geometry must be unprojected to calculate UTM zone")
 
         # calculate approximate longitude of centroid of union of all geometries in gdf
-        avg_lng = gdf["geometry"].to_crs("epsg:4326").representative_point().x.mean()
+        avg_lng = gdf["geometry"].representative_point().x.mean()
 
         # calculate UTM zone from avg longitude to define CRS to project to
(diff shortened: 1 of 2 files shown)


osmnx/projection.py
```diff
@@ -105,8 +105,8 @@ def project_gdf(gdf, to_crs=None, to_latlong=False):
         if is_projected(gdf.crs):  # pragma: no cover
             raise ValueError("Geometry must be unprojected to calculate UTM zone")
 
-        # calculate longitude of centroid of union of all geometries in gdf
-        avg_lng = gdf["geometry"].unary_union.centroid.x
+        # calculate approximate longitude of centroid of union of all geometries in gdf
+        avg_lng = gdf["geometry"].to_crs("epsg:4326").representative_point().x.mean()
```
gboeing, line 109: The `.to_crs("epsg:4326")` should be unnecessary as the lines above verify that the geometries are unprojected. The function should be able to handle geometries in any unprojected CRS without first requiring reprojection to epsg:4326.

gboeing (changes requested):

osmnx/projection.py
cheginit: I see. I thought it needs to be in 4326.

gboeing (approved):

gboeing:
Thanks!

PR #745, opened 2021-08-12
Expose parameters to requests methods

hokieg3n1us:
This proposed change would resolve any issues similar to issue #17 .

Description: Include a dictionary of configuration for parameters to requests methods. This allows configuring HTTP requests for environments where the nominatum or overpass api requires 2-way SSL or basic authentication. More details on what keyword arguments are available and how they're configured is available in the [requests documentation for advanced usage](https://docs.python-requests.org/en/master/user/advanced/). 

Example Usage:

import osmnx as ox

requests_config = {}
requests_config ['auth'] = ('username', 'password')
requests_config ['cert'] = '/path/to/cert'
requests_config ['verify'] = '/path/to/CA/'

osmnx.config(requests_config=requests_config )

### DIFF
--- a/osmnx/downloader.py
+++ b/osmnx/downloader.py
@@ -319,7 +319,7 @@ def _get_pause(base_endpoint, recursive_delay=5, default_duration=60):
 
     try:
         url = base_endpoint.rstrip("/") + "/status"
-        response = requests.get(url, headers=_get_http_headers())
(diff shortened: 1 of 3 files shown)


hokieg3n1us:
I think request_kwargs makes a lot of sense.  I can make this change once I'm back from travel.  No access to dev environment until Sunday.

PR #765, opened 2021-10-30
CI: Add Python 3.10 to job matrix

EwoutH:
CI: Add Python 3.10 to job matrix.

In draft status, currently unable to resolve dependencies.


gboeing:
No need to add tests for 3.10 yet. It usually takes a couple of months before the geospatial stack is ready for primetime on the newest python version. We should look into adding 3.10 compatibility around the end of the year.

EwoutH:
I agree, the ecosystem takes more time than I thought.

PR #773, opened 2021-11-04
ENH: Address Shapely and GeoPandas deprecations

cheginit:
Shapely and GeoPandas have deprecated some options that will be removed in their future version. This PR fixes those warnings. The deprecation functionalities are:

* Shapely has deprecated iterating over multi-part geometries which is documented [here](https://shapely.readthedocs.io/en/stable/migration.html).
* GeoPandas has deprecated op argument in favor of predicate in sjoin (geopandas/geopandas#1626).


gboeing:
Thanks @cheginit I'll be able to review this week.

gboeing (approved):

Issue #751, opened 2021-08-30
Pull request #699 breaks any data download with proxies - gaierror: [Errno 11001] getaddrinfo failed


jslorrma:
**Problem description**
Pull request #699 breaks any data download in case proxies are used or have to be used, e.g. if one is using osmnx from within a corporate network.  

Reason is that `socket.gethostbyname` can only resolve the URL's domain to an IP address for hosts within the corporate network. Trying to resolve an external URL raises the following error: `gaierror: [Errno 11001] getaddrinfo failed`

**Provide a complete minimal reproducible example**

```python
import osmnx as ox

lami, lama, lomi, loma = (18.505, 18.550, 73.83, 73.870)
G=ox.graph_from_bbox(lami, lama, lomi, loma, network_type="drive")
```

that gives the following error:

```python
~\.conda\envs\ox\lib\site-packages\osmnx\graph.py in graph_from_bbox(north, south, east, west, network_type, simplify, retain_all, truncate_by_edge, clean_periphery, custom_filter)
     86         truncate_by_edge=truncate_by_edge,
     87         clean_periphery=clean_periphery,
---> 88         custom_filter=custom_filter,
     89     )
     90 

~\.conda\envs\ox\lib\site-packages\osmnx\graph.py in graph_from_polygon(polygon, network_type, simplify, retain_all, truncate_by_edge, clean_periphery, custom_filter)
    418 
    419         # download the network data from OSM within buffered polygon
--> 420         response_jsons = downloader._osm_network_download(poly_buff, network_type, custom_filter)
    421 
    422         # create buffered graph from the downloaded data

~\.conda\envs\ox\lib\site-packages\osmnx\downloader.py in _osm_network_download(polygon, network_type, custom_filter)
    506     for polygon_coord_str in polygon_coord_strs:
    507         query_str = f"{overpass_settings};(way{osm_filter}(poly:'{polygon_coord_str}');>;);out;"
--> 508         response_json = overpass_request(data={"data": query_str})
    509         response_jsons.append(response_json)
    510     utils.log(

~\.conda\envs\ox\lib\site-packages\osmnx\downloader.py in overpass_request(data, pause, error_pause)
    699 
    700     # resolve url to same IP even if there is server round-robin redirecting
--> 701     _config_dns(base_endpoint)
    702 
    703     # define the Overpass API URL, then construct a GET-style URL as a string to

~\.conda\envs\ox\lib\site-packages\osmnx\downloader.py in _config_dns(url)
[...]

PR #774, opened 2021-11-04
Resolves the URL's domain to public IP address using Google's public DNS

jslorrma:
Current implementation to get the public IP from a URL's domain raises the following error: gaierror: [Errno 11001] getaddrinfo failed under certain conditions, e.g. if one is using osmnx from within a corporate network. (See #751)

This PR proposes a fix to use a public DNS therefore - Google's public DNS here.

The proposal changes `downloader.py` adding the new function definition `_gethostbyname` and patching function `_config_dns` to recieve IP via `_gethostbyname` in case of a `socket.gaierror` exception.

### DIFF
--- a/osmnx/downloader.py
+++ b/osmnx/downloader.py
@@ -250,6 +250,58 @@ def _get_http_headers(user_agent=None, referer=None, accept_language=None):
     return headers
 
 
+def _gethostbyname(host: str) -> list:
+    """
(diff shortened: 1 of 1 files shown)


gboeing:
Thanks @jslorrma I'll be able to review this week.

gboeing (approved):

PR #792, opened 2022-01-13
Plot undirected graph route correctly

amorfinv:
Plotting the route in an undirected graph fails because the geometry of the edge is always in a certain direction.

For example:

```python
import numpy as np
import osmnx as ox
from matplotlib import pyplot as plt

G = ox.graph_from_place(place, network_type="drive")
G_undirected = ox.get_undirected(G)

# print route
orig = list(G_undirected)[0]
dest = list(G_undirected[120]
route = ox.shortest_path(G_undirected, orig, dest, weight="length")
fig, ax = ox.plot_graph_route(G_undirected, route, route_color="y", route_linewidth=6, node_size=0)
```

This results in the following image:

funky_route

However, with the proposed changes the route is plotted like this:

output

I was thinking of creating a different `for` loop for `nx.MultiGraph` objects but the proposed `if` statement ([here](https://github.com/amorfinv/osmnx/blob/0d3b62da24c6d3af4ad91587156ce56fde2d932e/osmnx/plot.py#L308)) seems a bit cleaner to me. It would only check the second condition if the the graph is an `nx.MultiGraph`.

``` python
if not G.is_directed() and G.nodes[u]["x"] != data["geometry"].coords[0][0]:
```
Let me know if you think this would be useful or if you have any suggestions.

-Andres

### DIFF
--- a/osmnx/plot.py
+++ b/osmnx/plot.py
@@ -305,7 +305,7 @@ def plot_graph_route(
             # if geometry attribute exists, add all its coords to list
             xs, ys = data["geometry"].xy
             # When graph is nx.MultiGraph check if geometry of edge is in correct order
-            if not G.is_directed() and G.nodes[u]["x"] != data['geometry'].coords[0][0]:
+            if not G.is_directed() and G.nodes[u]["x"] != data["geometry"].coords[0][0]:
                 xs = xs[::-1]
                 ys = ys[::-1]
             x.extend(xs)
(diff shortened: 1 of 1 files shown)


amorfinv:
I made a new [branch](https://github.com/amorfinv/osmnx/tree/plot_fix) with just one commit that includes all the tests as that seems a bit cleaner.  I can close this request and make a new one with that branch if you prefer.

gboeing:
Yes, would you close this one and open a PR with that new branch? I'll take a look at what you did there. Thanks.

PR #794, opened 2022-01-18
geopandas 0.11 compatibility fix

martinfleis:
GeoPandas merged a a fix of an old bug and when you drop a geometry column from a GeoDataFrame, it now correctly returns a pandas.DataFrame instead of an invalid GeoDataFrame (https://github.com/geopandas/geopandas/pull/2060).

I've noticed that OSMnx now depends on this behaviour and tests fail when run against GeoPandas main. This PR issues a patch.


gboeing (approved):

gboeing:
Thank you.

PR #801, opened 2022-02-19
plot undirected graph route correctly

amorfinv:
Plotting the route in an undirected graph fails because the geometry of the edge is always in a certain direction.

For example:

```python
import numpy as np
import osmnx as ox
from matplotlib import pyplot as plt

G = ox.graph_from_place(place, network_type="drive")
G_undirected = ox.get_undirected(G)

# print route
orig = list(G_undirected)[0]
dest = list(G_undirected[120]
route = ox.shortest_path(G_undirected, orig, dest, weight="length")
fig, ax = ox.plot_graph_route(G_undirected, route, route_color="y", route_linewidth=6, node_size=0)
```

This results in the following image:

funky_route

However, with the proposed changes the route is plotted like this:

output

I was thinking of creating a different `for` loop for `nx.MultiGraph` objects but the proposed `if` statement ([here](https://github.com/amorfinv/osmnx/blob/ab85b50e0ae37acf7d130f88bcb53fa8c8bf498c/osmnx/plot.py#L308)) seems a bit cleaner to me. It would only check the second condition if the the graph is an `nx.MultiGraph`.

``` python
if not G.is_directed() and G.nodes[u]["x"] != data["geometry"].coords[0][0]:
```
Let me know if you think this would be useful or if you have any suggestions.

-Andres

### DIFF
--- a/osmnx/plot.py
+++ b/osmnx/plot.py
@@ -258,7 +258,7 @@ def plot_graph_route(
 
     Parameters
     ----------
-    G : networkx.MultiDiGraph
+    G : networkx.MultiDiGraph or networkx.MultiGraph
         input graph
     route : list
         route as a list of node IDs
@@ -304,6 +304,9 @@ def plot_graph_route(
         if "geometry" in data:
             # if geometry attribute exists, add all its coords to list
             xs, ys = data["geometry"].xy
+            # When graph is nx.MultiGraph check if geometry of edge is in correct order
+            if not G.is_directed() and G.nodes[u]["x"] != data["geometry"].coords[0][0]:
+                xs, ys = xs[::-1], ys[::-1]
             x.extend(xs)
             y.extend(ys)
         else:


gboeing:
Supersedes #792

gboeing:
Thanks @amorfinv. Can you help me understand your use case? You're modeling a driving network (which is directed) as an undirected graph. Can you explain the motivation?

If you want to model a "to-be" alternative network, a better method could be to add a *vu* edge from node *v* to *u* whenever edge *uv* exists but its reciprocal is missing. I say this because OSMnx's functions are designed to work with MultiDiGraphs (as stated in its docstrings). It would make the API less-consistent to make this change here for this function, but not for all the others that cannot handle MultiGraphs (e.g., `plot_route_folium`, among many others).

Another solution would be to enforce MultiDiGraph argument types in some of these functions, where OSMnx is overly permissive in not explicitly checking (though the required type is stated in the docstring).

amorfinv:
Sure. I am currently researching urban airspace. I assume that in some places drones may need to fly between buildings so they will be above the existing road network. So I stack a bunch of networks on top of each other to build my airspace and alternate the directionality of each with height. I simplify the problem by planning the horizontal route with an undirected graph.

Adding a _vu_ edge would also work but I just ended up implementing it in the way I proposed on this PR as it seemed easier. (this is where I implemented it -> [link](https://github.com/Metropolis-2/bluesky-rogue-scenariomaker))

I thought it would be useful for other cases where directionality may not matter but I can't think of any (maybe pedestrian networks?). However, I do see how this can make the API less consistent if it is only applied to this function. I can go through the other plotting functions and make sure they work with MultiGraphs if that is the direction you want to go in. Otherwise, I saw that networkx has some nice [decorators](https://github.com/networkx/networkx/blob/main/networkx/utils/decorators.py) for checking the graph type so that could be an elegant option.

gboeing:
Thanks for fleshing out the explanation. If you want a bidirectional driving network, the OSMnx way of doing this would be to append the "drive" network type to the `bidirectional_network_types` list in the `settings` module, like:

```python
import osmnx as ox
ox.settings.bidirectional_network_types.append('drive')

G = ox.graph_from_place(place, network_type="drive")
orig = list(G)[0]
dest = list(G)[120]
route = ox.shortest_path(G, orig, dest, weight="length")
fig, ax = ox.plot_graph_route(G, route, route_color="y", route_linewidth=6, node_size=0)
```

Now the driving network is modeled as a fully bidirectional graph. This allows us to consistently use MultiDiGraph objects throughout the package, but easily model networks as fully bidirectional whenever desired. Does this satisfy your use case?

amorfinv:
This works great. Thanks alot!. I will close the PR.

PR #805, opened 2022-03-05
edge params in plot_footprints

### LINKED ISSUE #795: Optional edge params for plot_footprints

`plot_footprints()` does not have the functionality to customize edge color and edge width. The function uses `gdf.plot()` which has optional `edgecolor` and `linewidth` parameters. These parameters are by default set to `"none"` and `0` respectively within `plot_footprints()` function, but they are not accessible to change. 

Add `edgecolor` and `edgewidth` parameters to `plot_footprints()`. Their default values can be `"none"` and `0`, which will leave the default behavior of `plot_footprints()` the exact same as before this update. Within the function, pass `edgecolor` to `gdf.plot()`'s `edgecolor`, and `edgewidth` to `gdf.plot()`'s `linewidth`.

**Additional context**

This is the update to the function. The changes are very minimal.
```
def plot_footprints(
    gdf,
    ax=None,
    figsize=(8, 8),
    color="orange",
    edgecolor="none",
    edgewidth=0,
    alpha=None,
    bgcolor="#111111",
    bbox=None,
    save=False,
    show=True,
    close=False,
    filepath=None,
    dpi=600,
):
    """
    Plot a GeoDataFrame of geospatial entities' footprints.

    Parameters
    ----------
    # omitting other param descriptions for brevity in feature proposal.
    edgecolor : string
        color of the edge of the footprints
    edgewidth : float
        width of the edge of the footprints
    
    Returns
    -------
    fig, ax : tuple
        matplotlib figure, axis
    """
    if ax is None:
        fig, ax = plt.subplots(figsize=figsize, facecolor=bgcolor, frameon=False)
        ax.set_facecolor(bgcolor)
    else:
        fig = ax.figure

    # retain only Polygons and MultiPolygons, then plot
    gdf = gdf[gdf["geometry"].type.isin({"Polygon", "MultiPolygon"})]
    ax = gdf.plot(ax=ax, facecolor=color, edgecolor=edgecolor, linewidth=edgewidth, alpha=alpha)

    # determine figure extents
    if bbox is None:
        west, south, east, north = gdf.total_bounds
    else:
        north, south, east, west = bbox

    # configure axis appearance, save/show figure as specified, and return
    ax = _config_ax(ax, gdf.crs, (north, south, east, west), 0)
    fig, ax = _save_and_show(fig, ax, save, show, close, filepath, dpi)
    return fig, ax
```

And now to show the new function in use. First, I'll grab some building footprints.
```
building_footprints = ox.geometries_from_address(
    '4 S Market St, Boston, MA 02109', 
    tags = {"building": True}, 
    dist = 250)
```

Now, I'll plot those footprints with the default parameters.
```
ox.plot_footprints(building_footprints, color = 'lightgrey', bgcolor='white')
```
boston_noedge

Finally, here is the function using the optional new parameters.
```
ox.plot_footprints(building_footprints, color = 'lightgrey', edgecolor = 'grey', edgewidth = 1, bgcolor = 'white')
```
boston_edge

I think that this provides a lot of aesthetic customization, but can also help more clearly convey edges when many footprints are clustered together. I've found it's particularly nice for buildings, water, green spaces, and city limits.

Thanks,
Will

WilliamThyer:
Fix #795

`plot_footprints()` does not have the functionality to customize edge color and edge width. The function uses `gdf.plot()` which has optional `edgecolor` and `linewidth` parameters. These parameters are by default set to `"none"` and `0` respectively within `plot_footprints()` function, but they are not accessible to change.

Fix this by adding `edgecolor` and `edgewidth` parameters to `plot_footprints()`. Their default values are `"none"` and `0`, which will leave the default behavior of `plot_footprints()` the exact same as before this update. Within the function, pass `edgecolor` to `gdf.plot()`'s `edgecolor`, and `edgewidth` to `gdf.plot()`'s `linewidth`.

### DIFF
--- a/osmnx/plot.py
+++ b/osmnx/plot.py
@@ -560,8 +560,8 @@ def plot_footprints(
     ax=None,
     figsize=(8, 8),
     color="orange",
-    edgecolor="none",
-    edgewidth=0,
+    edge_color="none",
+    edge_linewidth=0,
     alpha=None,
     bgcolor="#111111",
     bbox=None,
@@ -584,9 +584,9 @@ def plot_footprints(
         if ax is None, create new figure with size (width, height)
     color : string
         color of the footprints
-    edgecolor : string
+    edge_color : string
         color of the edge of the footprints
-    edgewidth : float
+    edge_linewidth : float
         width of the edge of the footprints
     alpha : float
         opacity of the footprints
@@ -620,7 +620,7 @@ def plot_footprints(
 
     # retain only Polygons and MultiPolygons, then plot
     gdf = gdf[gdf["geometry"].type.isin({"Polygon", "MultiPolygon"})]
-    ax = gdf.plot(ax=ax, facecolor=color, edgecolor=edgecolor, linewidth=edgewidth, alpha=alpha)
+    ax = gdf.plot(ax=ax, facecolor=color, edgecolor=edge_color, linewidth=edge_linewidth, alpha=alpha)
 
     # determine figure extents
     if bbox is None:
--- a/osmnx/plot.py
+++ b/osmnx/plot.py
@@ -560,6 +560,8 @@ def plot_footprints(
     ax=None,
     figsize=(8, 8),
     color="orange",
+    edgecolor="none",
+    edgewidth=0,
     alpha=None,
     bgcolor="#111111",
     bbox=None,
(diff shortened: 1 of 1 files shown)


gboeing (changes requested):
We should probably keep the parameter names consistent with the other functions in the module. See `plot_graph`'s function signature: https://github.com/gboeing/osmnx/blob/main/osmnx/plot.py#L125-L144

gboeing (approved):

Issue #810, opened 2022-03-15
single incorrect result for nearest edge


jGaboardi:
**Problem description**
  - What did you do?
    - Attempted to find the nearest edge to a point. 
  - What did you expect to happen?
    - Return the nearest edge. 
  - And what actually happened instead?
    - Returns the incorrect edge only for a *specific point* (same edge every time).
    - This happens at multiple scales, but returns the correct nearest edge when the "troublemaker"(?) edge is not included. The following includes links to more detailed examples:
      - correct result with network instance [600 meters](https://gist.github.com/jGaboardi/f21109b0d48d1b9af75262592f5a11f4) from point
      - incorrect result with network instance [1200 meters](https://gist.github.com/jGaboardi/48d08a0d7b1db55412668ad15b9e125e) & [2400 meters](https://gist.github.com/jGaboardi/e4b0b189dff89850fe1a812e228e7ae6) from point
      - incorrect result with network instance of all Knox County, TN

**Environment information**
  - What operating system are you using?
    - macOS Big Sur v11.6.5 (20G527)
  - What Python version are you using?
    - 3.10.2
  - What OSMnx version are you using?
    - 1.1.2
  - Provide a complete list of your environment's packages and their versions (for example, run `conda list` or `pip list` then paste the output between the two "details" tags below)

<details>

```
# Name                    Version                   Build  Channel
actlike                   0.0.9                     dev_0    <develop>
affine                    2.3.0                      py_0    conda-forge
appdirs                   1.4.4              pyh9f0ad1d_0    conda-forge
appnope                   0.1.2           py310h2ec42d9_2    conda-forge
argon2-cffi               21.3.0             pyhd8ed1ab_0    conda-forge
argon2-cffi-bindings      21.2.0          py310he24745e_1    conda-forge
asttokens                 2.0.5              pyhd8ed1ab_0    conda-forge
attrs                     21.4.0             pyhd8ed1ab_0    conda-forge
backcall                  0.2.0              pyh9f0ad1d_0    conda-forge
backports                 1.0                        py_2    conda-forge
backports.functools_lru_cache 1.6.4              pyhd8ed1ab_0    conda-forge
beautifulsoup4            4.10.0             pyha770c72_0    conda-forge
black                     22.1.0             pyhd8ed1ab_0    conda-forge
bleach                    4.1.0              pyhd8ed1ab_0    conda-forge
blosc                     1.21.0               he49afe7_0    conda-forge
[...]

PR #813, opened 2022-03-17
addressing #810 - breaking ties from rtree

jGaboardi:
This PR addresses #810, which shows that within `rtree` a nearest edge can be selected arbitrarily when bounding boxes overlap. Adding the `num_results=10` to `rtree.nearest()` [here](https://github.com/gboeing/osmnx/blob/605fa2f75c0d3e886b087964dc137d7491af4113/osmnx/distance.py#L288) resolves this infrequent situation by selecting the network edge that is actually nearest.


gboeing:
Thanks @jGaboardi

Issue #784, opened 2021-11-21
Handle attributes that are based on OSM way direction


teddyort:
**Is your feature proposal related to a problem?**

Some OSM attributes are based implicitly on the knowledge of the direction of the underlying OSM way. For example, [lanes:forward](https://wiki.openstreetmap.org/wiki/Key:lanes#Lanes_in_different_directions) and lanes:backward both are described with respect to the order of the nodes in the way.

Problem: When osmnx generates a graph, it adds edges in both directions and copies the attributes to both edges. Therefore, if direction specific attributes are present, it becomes ambiguous as to which direction is "forward" and which is "backward".

**Describe the solution you'd like to propose**

I think a clean solution would be to add a tag to every edge _reversed={True, False}_ as a boolean indicating whether the edge is in the original direction or reversed.

I would be happy to provide a PR for this, if it would be helpful.

**Describe alternatives you've considered**

Another option would be to make a comprehensive list of direction dependent tags and handle them during graph creation. For example, the "forward" direction in attributes can be assigned to the forward edge, and the backward direction attributes to the backward edge. However, while this could be useful for some cases, I would worry that we couldn't forsee all the possible tags and usecases and so would prefer the proposed solution which provides the necessary information (that an edge was reversed) and leaves the processing to the end-user.

**Additional context**
As an example of the problem see https://www.openstreetmap.org/way/426604957. There the _lanes=3_, and _lanes:forward=2_ and _lanes:backward=1_. These attributes are copied "as is" to both the forward and backward directed edges in the MultiDiGraph which makes it impossible to detect which of the edges has 2 lanes, and which as 1. Additionally, this edge also has an attribute _turn:lanes:forward=left|through;right_ which describes the turning lanes of the edge approaching the intersection. Again, there would be know way to identify from the graph which of these edges the property applies to.

gboeing:
Let me know if you'd like to contribute this as a PR.

teddyort:
Yes, I'll be happy to. I should be able to submit it shortly.

gboeing:
Closed by #814

PR #814, opened 2022-03-19
Add a tag to indicate when an edge is reversed from the original OSM

teddyort:
This PR adds a tag to each edge indicating if it was created by reversing the original OSM node order. This allows using order-dependent attributes (such as lane directions) as described in https://github.com/gboeing/osmnx/issues/784.

To use this, check if an edge was reversed after downloading an OSM graph:
```
G = ox.graph_from_bbox(37.79, 37.78, -122.41, -122.43, network_type='drive')
print(G[65285109][5352470018][0]['reversed'])
```

### DIFF
--- a/osmnx/graph.py
+++ b/osmnx/graph.py
@@ -772,12 +772,15 @@ def _add_paths(G, paths, bidirectional=False):
         if not settings.all_oneway:
             path["oneway"] = is_one_way
 
-        # zip path nodes to get (u, v) tuples like [(0,1), (1,2), (2,3)]. if
-        # the path is NOT one-way, reverse direction of each edge and add this
-        # path going the opposite direction too
+        # zip path nodes to get (u, v) tuples like [(0,1), (1,2), (2,3)].
         edges = list(zip(nodes[:-1], nodes[1:]))
-        if not is_one_way:
-            edges.extend([(v, u) for u, v in edges])
 
         # add all the edge tuples and give them the path's tag:value attrs
+        path['reversed'] = False
         G.add_edges_from(edges, **path)
+
+        # if the path is NOT one-way, reverse direction of each edge and add 
+        # this path going the opposite direction too
+        if not is_one_way:
+            path['reversed'] = True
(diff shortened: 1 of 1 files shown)


teddyort:
Thanks for fixing this for me @gboeing. I was meaning to do it for a while but was having some issues with my setup. Really appreciate all you do to make osmnx available for everyone!

zhongyinggan:
@teddyort @gboeing Thank you so much for your inputs. I've learned a lot. I am currently using OSMnx for routing, and I would like to know the the direction of freeway. For example, is it I-110 N or I-110 South.

The following are my codes:

import osmnx as ox
import networkx as nx


G = ox.graph_from_place("Los Angeles County, California, USA", network_type="drive")

origin = (33.821642, -118.281422)
destination = (33.876752, -118.055168)
origin_node = ox.distance.nearest_nodes(G, origin[1], origin[0])
destination_node = ox.distance.nearest_nodes(G, destination[1], destination[0])
route = ox.shortest_path(G, origin_node, destination_node, weight="travel_time")
fig, ax = ox.plot_graph_route(G, route, route_color="c", node_size=0)

G.edges[(route[8], route[9], 0)]


The following is the output:
{'osmid': [604115924, 604115927],
 'oneway': True,
 'lanes': '4',
 'ref': 'I 110',
 'name': 'Harbor Freeway',
 'highway': 'motorway',
 'reversed': False,
 'length': 1232.397,
 'geometry': <LINESTRING (-118.287 33.831, -118.287 33.836, -118.287 33.837, -118.287 33....>}

The value for 'reversed' is False, which means the direction is the same as the default OpenStreetMap direction. However, I was wondering how I could find out the default OpenStreetMap direction, in this case, is it North or South. Otherwise, the value of 'reversed' would be meaningless to me.

Thank you, and any comments will be greatly appreciated.

gboeing:
Hi @zhongyinggan thanks for using OSMnx! Please ask usage questions on StackOverflow and someone in the community will be able to help you.

Issue #776, opened 2021-11-11
Add APIs for loading/saving graphml from/to IO-like objects


xSAVIKx:
**Is your feature proposal related to a problem?**

Loading graphs in serverless environments doubles the amount of RAM required to load a graph.

The reason for that is that with the current APIs one is forced to store the graphml file in the file system prior to loading it with e.g. `load_graphml`. And in serverless environments file system is usually mounted directly into the memory.
It may be even worse if you first need to store it in some IO stream cause it's not possible to write directly into a file for some reason (you end up with 3x more memory needed)

**Describe the solution you'd like to propose**

The `networkx` already provides an API to parse a graph from a string (`parse_graphml`), so going to use it and have a similar one in the `osmnx.io` package that does the `load_graphml` logic but for a graph from a string.

**Describe alternatives you've considered**

Re-implement the `load_graphml` logic but outside of the library. Use `parse_graphml` directly and then use copy-pasted logic from the library.

**Additional context**

```python

graph = b"my-graph-data"
with open("/tmp/my-graph", "wb") as graph_file:
    graph_file.write(graph)

G = ox.io.load_graphml("/tmp/my-graph")
os.remove("/tmp/my-graph")

# Do stuff with `G`
```

xSAVIKx:
It's hard to say, but this should be the case for any serverless environment while you don't have a normal file system there. And even if you do have a file system, why one may want to write the content to disk first and then read it again? I mean, if I am downloading the file from a remote location it just doesn't make a lot of sense to do the job twice.

gboeing:
Thanks. If you'd like to open a PR I can take a look.

gboeing:
Let me know if you'd like to contribute this as a PR.

xSAVIKx:
I'd love to as soon as the war in Ukraine is gonna be won. I'm from Kharkiv (very close to the Russian border) and had to leave my hometown due to massive shelling and airstrikes.

benafischer94:
I think the use case for this will continue grow.  I was working through a similar use case of needing the data but trying to reduce number of copies across containers/functions consuming it, as well as the issue of downloading, writing, then reading again.
I'll take a look and see if I can work through the PR for this one. 

@xSAVIKx , hope you can return home safe soon.

Vermouth1103:
Is it possible to load shp/geojson format graph with osmnx? As I can see, there isn't any supporting methods in the docs.

gboeing:
[...]

PR #815, opened 2022-03-22
Addressing #776 - Add APIs for loading/saving graphml from/to IO-like objects

benafischer94:
This PR addresses #776 by adding a new function to the IO api `load_graphml_stream` for creating a GraphML instance from a bytes string or string object. This allows reading GraphML data without having to first write to disc. Useful for serverless instances/those pulling the data regularly form a 3rd party source. Included support for both bytes string and string objects to simplify implementation. If the user has the string in memory and already converted to string for other purposes, no need to convert back to bytes string.

Included updated test with a sample bytes string GML data and added to _api.

Usage is similiar to `load_graphml` with a bytes string and optional encoding instead of filepath

```python
graph = b"my-graph-data"
G = ox.load_graphml_stream(graph,encoding="utf-8")
# Do stuff with G
```


benafischer94:
@gboeing  Looks like they're passing now on my instance, I missed re-running black after adding some codecov skips for error handling. So I believe it should pass the actions now. Thanks!

benafischer94:
@gboeing looking at what's not covered at this point, it appears in line with the rest of the project tests. I've added a bit more exception handling to this function than surrounding. For the error handling I've added most are marked no cover. I'm not sure if this is the right approach or if intentional failure tests are preferred.

Line 251 in io.py is a copy-paste from load_graphml
```python
if graph_dtypes is not None:
    default_graph_dtypes.update(graph_dtypes) # pragma: no cover
if node_dtypes is not None:
    default_node_dtypes.update(node_dtypes)
if edge_dtypes is not None:
    default_edge_dtypes.update(edge_dtypes)
```
Which previously didn't have the None type checking in tests.

Open to feedback with regards to how you would like to see this handled.

PR #823, opened 2022-05-20
Enhancements to #815

gboeing:
Enhancements to #815

### DIFF
--- a/osmnx/_api.py
+++ b/osmnx/_api.py
@@ -32,7 +32,6 @@ from .graph import graph_from_point
 from .graph import graph_from_polygon
 from .graph import graph_from_xml
 from .io import load_graphml
-from .io import load_graphml_stream
 from .io import save_graph_geopackage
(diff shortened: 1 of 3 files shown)


gboeing:
@benafischer94 I have simplified this functionality somewhat. Essentially, we're handing off stream reading and decoding to the user and consolidating the GraphML loading functionality into a single file to avoid so much duplicated logic. Now the user can pass either a GraphML string or a GraphML filepath to the `load_graphml` function. From the updated docs:

> Note that you must pass one and only one of `filepath` or `graphml_str`. If passing `graphml_str`, you may need to decode the bytes read from your file before converting to string to pass to this function.

And the usage looks like this:
```python
file_bytes = open("tests/input_data/short.graphml", "rb").read()
data = str(file_bytes.decode())
G = ox.load_graphml(graphml_str=data, node_dtypes=nd, edge_dtypes=ed)
```

PR #836, opened 2022-06-10
Use defaultdict instead of manual dict manpulation

RobertLucey:
Drops lines, also slightly faster.

### DIFF
--- a/osmnx/simplification.py
+++ b/osmnx/simplification.py
@@ -1,5 +1,6 @@
 """Simplify, correct, and consolidate network topology."""
 
+from collections import defaultdict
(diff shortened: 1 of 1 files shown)


gboeing:
Thanks for using OSMnx, and thanks for offering this contribution. Note the tests are not passing

Looking through it, I'm not sure yet if we'll merge it in. In my testing it looks like the speed-up is negligible and I feel like the code is less readable. `defaultdict` may be a useful language feature here for terse code and a small performance increase but I want to make sure the simplification algorithm remains as readable as can be (without meaningfully impacting performance).

PR #837, opened 2022-06-14
Fix Pandas' FutureDeprecation warning

cheginit:
This PR just casts the indexer to a `list` from `set` since Pandas throws:

```bash
osmnx/simplification.py:503: FutureWarning: Passing a set as an indexer is deprecated and will raise in a future version. Use a list instead.
  gdf.loc[wcc, "cluster"] = f"{cluster_label}-{suffix}"
```


gboeing:
Thanks.

PR #842, opened 2022-07-05
Proposed fix for CI failure

nickodell:
flake8-bugbear complains of "B023 Function definition does not bind loop variable 'u'." ([link to action log](https://github.com/gboeing/osmnx/runs/7175001370?check_suite_focus=true#step:5:14).) See [the upstream issue](https://github.com/PyCQA/flake8-bugbear/issues/263) for why the warning was added and what it's checking for. 

This is not really a bug, because the lambda only lasts as long as a single iteration of the iterator. This PR fixes the warning by passing the edge as a parameter to the key lambda, which avoids binding the loop var.

Also, I refactored by moving _get_shortest_edges to own function. Too hard to wrap my head around the code otherwise.

Tested changes by using the 11-plot-routes-folium-web-map.ipynb example.

You may also want to just ignore the bugbear error as a false positive. If that's your preference, feel free to close this PR.

### DIFF
--- a/osmnx/elevation.py
+++ b/osmnx/elevation.py
@@ -42,7 +42,8 @@ def _query_raster(nodes, filepath, band):
     """
     # must open raster file here: cannot pickle it to pass in multiprocessing
     with rasterio.open(filepath) as raster:
-        values = np.array(tuple(raster.sample(nodes.values, band)), dtype=float).squeeze()
+        nodes_generator = (tuple(x) for x in nodes.values)
+        values = np.array(tuple(raster.sample(nodes_generator, band)), dtype=float).squeeze()
         values[values == raster.nodata] = np.nan
         return zip(nodes.index, values)
 
--- a/osmnx/folium.py
+++ b/osmnx/folium.py
@@ -92,7 +92,7 @@ def plot_route_folium(
     """
     # create gdf of the route edges in order
     node_pairs = zip(route[:-1], route[1:])
-    uvk = ((u, v, min(G[u][v], key=lambda k: G[u][v][k]["length"])) for u, v in node_pairs)
+    uvk = ((u, v, min(G[u][v].items(), key=lambda k: k[1]["length"])[0]) for u, v in node_pairs)
     gdf_edges = utils_graph.graph_to_gdfs(G.subgraph(route), nodes=False).loc[uvk]
     return _plot_folium(gdf_edges, route_map, popup_attribute, tiles, zoom, fit_bounds, **kwargs)
 


gboeing:
> This is not really a bug, because the lambda only lasts as long as a single iteration of the iterator. This PR fixes the warning by passing the edge as a parameter to the key lambda, which avoids binding the loop var.

Thanks @nickodell this looks like a good fix, and I'd rather fix it than ignore it just for the sake of keeping the code "by the book."

> Also, I refactored by moving _get_shortest_edges to own function.

This may be a useful idea but I think we need to consider it further first. For the sake of keeping this PR clean, can we just limit it to a minimal-diff CI fix for now?

nickodell:
>For the sake of keeping this PR clean, can we just limit it to a minimal-diff CI fix for now?

@gboeing Just pushed a commit which reduces it to a one line change. Now I'm getting a test failure that I'm still tracking down.

nickodell:
I tracked down the cause of the error. It seems to be a change to rasterio in raster.sample() in 1.3.0.

Observations:

* The error happens in the `test_elevation` test.
* The issue does not occur if I downgrade rasterio to 1.2.10.
* The issue does occur if I install rasterio 1.3.0.
* The error message is the following:

   ```
   ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()   
   ```

 * The exception happens on this line in rasterio:

   ```
   pts = zip(*filter(None, pts))
   ```

   This line takes the list of points, drops any points which are falsey, then converts a list like `[(x, y), (x, y), (x, y)]` into lists of each coordinate, like `[[x, x, x], [y, y, y]]`. When it checks if the point is falsey, numpy throws an exception, because an array of two points doesn't have a true/false value.

Here's a potential fix for this:

```
diff --git a/osmnx/elevation.py b/osmnx/elevation.py
index 79aec49..820ca8f 100644
--- a/osmnx/elevation.py
+++ b/osmnx/elevation.py
@@ -42,7 +42,8 @@ def _query_raster(nodes, filepath, band):
     """
     # must open raster file here: cannot pickle it to pass in multiprocessing
     with rasterio.open(filepath) as raster:
-        values = np.array(tuple(raster.sample(nodes.values, band)), dtype=float).squeeze()
+        nodes_list = [tuple(x) for x in nodes.values]
+        values = np.array(tuple(raster.sample(nodes_list, band)), dtype=float).squeeze()
         values[values == raster.nodata] = np.nan
         return zip(nodes.index, values)
```

The downside is that you'd have a speed penalty from converting between a numpy array and a list of tuples.

gboeing:
> When it checks if the point is falsey, numpy throws an exception, because an array of two points doesn't have a true/false value.

Thanks for tracking that down. I wonder if it's worth opening an issue upstream at rasterio as it seems odd to suddenly not be able to pass a numpy array into the sample method.

On this line:

```python
nodes_list = [tuple(x) for x in nodes.values]
```

Can we convert the list comprehension to generator comprehension to make it more efficient?

gboeing:
The generator looks good. If you decide to open an upstream issue, we can patch OSMnx for now and revisit again later if rasterio regains numpy compatibility.

gboeing:
Thanks!

Issue #855, opened 2022-08-22
ARM Docker Image


tomharvey:
**Is your feature proposal related to a problem?**
Yes

When I use the docker image on my ARM based machine it ... is ... very ... slow.

**Describe the solution you'd like to propose**
Build the image for ARM architecture

Maybe you can build cross platform from your workstation and continue to manually build-push the images to docker hub?

Might require some changes to the build pipeline - which I see in https://github.com/gboeing/osmnx/pulls you're already looking into - whereby a remote pipeline is used that is capable of building for both.

**Describe alternatives you've considered**

I've built an ARM version myself and I use that. So, I'm happy, but maybe there are others out there? I'm on ARM so checking out the repo and running `docker build` was enough.

**Additional context**

I sent you a https://twitter.com/alushtom/status/1557747027286564864 in case this isn't the best place.

Something like the below would build cross platofrm

```
docker buildx build --platform 'linux/amd64,linux/arm64,linux/arm/v7,linux/arm/v8' -t gboeing/osmnx .'
```

Some more info on that command https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

gboeing:
I'm open to building for amd64 and arm64. However, when I looked into this in the past, I got an error running something simple like this on my linux machine:
```
>>> docker buildx build --platform linux/arm64 .
.buildkit_qemu_emulator: /bin/bash: Invalid ELF image for this architecture
```
As I don't have a personal use for this, I dropped it at that. If you want to open a PR with a simple change to get multi-platform working, I'm open to reviewing it.

tomharvey:
I suspect that neither of us has the time, inclination or incentive to debug QEMU on your workstation.

Given that, how open are to moving the build/push to a remote machine? GitHub actions would seem like the path of least resistance of you were open to that kind of setup. Then, any changes to master could automatically build and push a new image.

If that’s worth exploring I can set that up on a fork for your review.

There are additional benefits in removing your workstation from the work flow too.

gboeing:
> Then, any new tags could automatically build and push a new image.

This could work. The reason I haven't done so in the past is that it takes an hour or more (it varies) for the new version to show up in conda's search index, which is necessary to build the docker image using the canonical OSMnx installation method. Hence, I just grab lunch then run a bash script when I get back.

tomharvey:
Lol. Blasted reality.

[...]

PR #856, opened 2022-08-23
Multi platform GitHub action

tomharvey:
Following up with a proof of concept on https://github.com/gboeing/osmnx/issues/855

Uses GitHub Actions to build the docker images on a remote machine. This remote machine is configured to allow builds of an ARM based image (using QEMU)

Some more "upstream reality" scuppered the original plan for a single cross-platform image. [The Jupyter project creates separate images](https://github.com/jupyter/docker-stacks/issues/1737) for arm and x86 and since we need to use the relevant one of these as the base image, we also need to create separate images.

So, this workflow will build `osmnx:latest` and `osmnx:aarch64-latest` which is in line with the naming convention used on the Jupyter project.

### Work remaining

Before I go further, I wanted to post a draft PR and see what you think. You should be able to see the build logs of the images in https://github.com/tomharvey/osmnx/actions but I think if you approve the new workflow it'll run on your account so you can see it build in there. Note that the aarch64 build takes considerably longer as it's run under QEMU emulation

- [ ] We're currently only building `latest` tags, add the ability to build for tag version (such as `v1.2`)
- [ ] Build on manual trigger instead of code push
- [ ] Actually push to dockerhub. Suggest that I initially push to `tomharvey/osmnx:{aarch64-}latest` so we can QA from there.
- [ ] What to do with the conda env? `conda env export -n base > /home/jovyan/work/environment.yml`

The current build.sh dumps the conda env to your workstation. What's that used for? This could upload that file to the build artefacts on GitHub if you needed to keep that.

### Potential future improvements

None of the below are strictly necessary, but let me know if you think otherwise.

##### 1. Add some testing.
Just as in the build step you've got `ipython -c "import osmnx; print('OSMnx version', osmnx.__version__)"` we could add a test step to ask the built image to run this before pushing to dockerhub.

##### 2. Build more images for more feature packed containers.
As we have a mechanism to choose the base image tag, we can expand on this to change the base image, and build an osmnx image from the [base/minimal/datascience/pyspark notebook images as well](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html).

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -6,37 +6,14 @@ env:
 on: [push]
 jobs:
   amd64-base:
-    runs-on: ubuntu-latest
-    steps:
-      -
-        name: Checkout
-        uses: actions/checkout@v3
-      -
-        name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v2
-      -
-        name: Build and push
-        uses: docker/build-push-action@v3
-        with:
-          context: "./environments/docker"
-          push: false
-          platforms: amd64
-          tags: ${{ env.OWNER }}/osmnx:latest
-  aarch64-base:
-    runs-on: ubuntu-latest
-    steps:
-      -
-        name: Checkout
-        uses: actions/checkout@v3
-      -
-        name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v2
-      -
-        name: Build and push
-        uses: docker/build-push-action@v3
-        with:
-          context: "./environments/docker"
-          push: false
-          platforms: arm64
-          tags: ${{ env.OWNER }}/osmnx:aarch64-latest
-          build-args: ARCH=aarch64-latest
+    uses: ./.github/workflows/docker-build-push.yml
+    with:
+      platform: amd64
+      resulting_tag: 'latest'
+      base_tag: 'latest'
+  # aarch64-base:
+  #   uses: ./.github/workflows/docker-build-push.yml
+  #   with:
+  #     platform: arm64
+  #     resulting_tag: 'aarch64-latest'
+  #     base_tag: 'aarch64-latest'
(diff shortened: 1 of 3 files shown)


.github/workflows/docker.yml
```diff
@@ -0,0 +1,19 @@
...
+  OWNER: ${{ github.repository_owner }}
+
+on: [push]
+jobs:
+  amd64-base:
+    uses: ./.github/workflows/docker-build-push.yml
+    with:
+      platform: amd64
+      resulting_tag: 'latest'
+      base_tag: 'latest'
```
tomharvey, line 13: The `resulting_tag` is what is pushed to Dockerhub as `osmnx:resulting_tag` while the `base_tag` is what is pulled in the Dockerfile's `FROM jupyter/base-notebook:$BASE_TAG`

Could use the same thing to enforce strict compliance with the Jupyter notebook team naming convention, but wanted to float this option.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,19 @@
...
+  OWNER: ${{ github.repository_owner }}
+
+on: [push]
+jobs:
+  amd64-base:
+    uses: ./.github/workflows/docker-build-push.yml
+    with:
+      platform: amd64
+      resulting_tag: 'latest'
+      jupyter_tag: 'latest'
```
tomharvey, line 13: The `resulting_tag` is what is pushed to Dockerhub as `osmnx:resulting_tag` while the `jupyter_tag` is what is pulled in the Dockerfile's `FROM jupyter/base-notebook:$JUPYTER_TAG`

Could use the same argument for both to enforce strict compliance with the Jupyter notebook team naming convention, but wanted to float this option.

gboeing:
Closing this in accordance with the comments in #855

Issue #755, opened 2021-10-04
Use `nwr` shortcut to reduce the number of copies of a polygon in a request.


nickodell:
**Is your feature proposal related to a problem?**

The overpass query generated from geometries_from_place() or geometries_from_polygon() is not as efficient as it could be. When querying small, detailed geometry, the request can be orders of magnitude larger than the response. Much of that space is used to encode the polygon being searched for multiple times. For example, the following OSM script makes a request which repeats the polygon nine times:

```py
import geopandas as gpd
import osmnx as ox

ox.utils.config(
    log_console=True,
)

def pull_osm_hopewell():
    tags = {'amenity': ['hospital', 'fire_station', 'police']}
    gdf = ox.geometries_from_place("Hopewell, VA", tags)
    return gdf

emergencyservice = pull_osm_hopewell()
```

(Why nine times? Three for each kind of tag to search for, times three for each of node, way, and relation.)

**Describe the solution you'd like to propose**

In this loop, the polygon is repeated for each of node, way, and relation.

```py
    for d in tags_list:
        for key, value in d.items():

            if isinstance(value, bool):
                # if bool (ie, True) just pass the key, no value
                tag_str = f"['{key}'](poly:'{polygon_coord_str}');(._;>;);"
            else:
                # otherwise, pass "key"="value"
                tag_str = f"['{key}'='{value}'](poly:'{polygon_coord_str}');(._;>;);"

            for kind in ("node", "way", "relation"):
                components.append(f"({kind}{tag_str});")
```

Instead, I suggest using the shortcut nwr. This reduces the number of copies of the polygon in the example query from nine to three.

**Describe alternatives you've considered**

Another way to reduce the size of the query would be to combine the tag query. Instead of 

```
node[amenity=hospital](polygon)
node[amenity=fire_station](polygon)
```

you could use 

```
node[amenity=hospital or amenity=fire_station](polygon)
```

If you did it this way, this could reduce the number of copies of the polygon down to one. I didn't pursue that because it seemed more complicated.

**Additional context**

In order to measure the actual effect of this change, I added logging code to measure the size of the request query. Here's the size of the request made by the example code, without and with nwr:

Without nwr:

```
2021-10-04 15:57:05 Downloaded 4.9kB, uploaded 68.5kB from overpass-api.de
```

With nwr:
```
2021-10-04 15:58:02 Downloaded 4.9kB, uploaded 22.9kB from overpass-api.de
```

[...]

PR #868, opened 2022-09-22
Optimize _create_overpass_query function

### LINKED ISSUE #862: Streamline _create_overpass_query method
Hello, first time contributor here

Yes. The _create_overpass_query method creates a larger than necessary query string. In it are three main issues that cause a significant amount of redundancy: 1) each tag-value pair is explicitly called one by one, 2) each tag-value is called three times (node, way, relation), and 3) the polygon coordinate string is replicated for every tag-value pair.

The following changes address the above three deficiencies:
1) multiple values for the same tag should be combined together into a single component, as per Overpass documentation https://dev.overpass-api.de/overpass-doc/en/criteria/per_tag.html
Example: 
way[highway~"^(trunk|primary|secondary|tertiary)$"]

2) The node, way, and relation trifecta can be replaced with the single query statement nwr as per Overpass issue https://github.com/drolbr/Overpass-API/issues/453#issuecomment-378718323

3) The polygon can be globalized and referenced multiple times, rather than explicitly sending the same coordinate sequence for each query component, as per Overpass documentation https://dev.overpass-api.de/overpass-doc/en/criteria/union.html
Example: 
nwr({{bbox}})->.all;
(
  nwr.all[amenity=atm];
  nwr.all[atm=yes];
);

**A clear and concise description of what you want to change in OSMnx and how you plan to implement it.**

Replace the guts of the _create_overpass_query method to take advantage of the above. On an example query of mine the query size shrank from 20,625 bytes to 770 bytes (see attached files pre.txt and post.txt). osmnx lint and unit tests all pass in my forked copy of the repo.

[pre.txt](https://github.com/gboeing/osmnx/files/9437640/pre.txt)
[post.txt](https://github.com/gboeing/osmnx/files/9437641/post.txt)

**Provide a complete minimal working example of code to demonstrate your use case. Add any additional code snippets, context, or screenshots about the feature proposal to demonstrate its importance and feasibility.**

```python
def _create_overpass_query(polygon_coord_str, tags):
    """
    Create an overpass query string based on passed tags.

    Parameters
    ----------
    polygon_coord_str : list
        list of lat lng coordinates
    tags : dict
        dict of tags used for finding elements in the selected area

    Returns
    -------
    query : string
    """
    # create overpass settings string
    overpass_settings = _make_overpass_settings()

    # make sure every value in dict is bool, str, or list of str
    error_msg = "tags must be a dict with values of bool, str, or list of str"
    if not isinstance(tags, dict):  # pragma: no cover
        raise TypeError(error_msg)

    # make sure every value in dict is bool, str, or list of str
    error_msg = "tags must be a dict with values of bool, str, or list of str"
    if not isinstance(tags, dict):  # pragma: no cover
        raise TypeError(error_msg)

    # add node/way/relation query components one at a time
    components = []
    for key, value in tags.items():

        if isinstance(value, bool) and value:
            # if bool (ie, True) just pass the key, no value
            tag_str = f"[{key}]"

        elif isinstance(value, str):
            tag_str = f"[{key}='{value}']"

        elif isinstance(value, list):
            if not all(isinstance(s, str) for s in value):  # pragma: no cover
                raise TypeError(error_msg)

            if len(value) == 1:
                tag_str = f"[{key}='{value[0]}']"
            else:
                tag_str = "[{0}~'^({1})$']".format(key, "|".join(value))

        else:  # pragma: no cover
            raise TypeError(error_msg)

        components.append(f"nwr.all{tag_str};")

    # finalize query and return
    components = "".join(components)
    query = f"{overpass_settings};nwr(poly:'{polygon_coord_str}')->.all;({components});(._;>;);out;"

    return query
```

derinwalters:
Resolves #862 
Memory and speed optimization. No impact on usage or functionality.

### DIFF
--- a/osmnx/downloader.py
+++ b/osmnx/downloader.py
@@ -460,50 +460,39 @@ def _create_overpass_query(polygon_coord_str, tags):
     if not isinstance(tags, dict):  # pragma: no cover
         raise TypeError(error_msg)
 
-    tags_dict = dict()
+    # make sure every value in dict is bool, str, or list of str
+    error_msg = "tags must be a dict with values of bool, str, or list of str"
+    if not isinstance(tags, dict):  # pragma: no cover
+        raise TypeError(error_msg)
+
+    # add node/way/relation query components one at a time
+    components = []
     for key, value in tags.items():
 
-        if isinstance(value, bool):
-            tags_dict[key] = value
+        if isinstance(value, bool) and value:
+            # if bool (ie, True) just pass the key, no value
+            tag_str = f"[{key}]"
 
         elif isinstance(value, str):
-            tags_dict[key] = [value]
+            tag_str = f"[{key}='{value}']"
 
         elif isinstance(value, list):
             if not all(isinstance(s, str) for s in value):  # pragma: no cover
                 raise TypeError(error_msg)
-            tags_dict[key] = value
+
+            if len(value) == 1:
+                tag_str = f"[{key}='{value[0]}']"
+            else:
+                tag_str = "[{0}~'^({1})$']".format(key, "|".join(value))
 
         else:  # pragma: no cover
             raise TypeError(error_msg)
 
-    # convert the tags dict into a list of {tag:value} dicts
-    tags_list = []
-    for key, value in tags_dict.items():
-        if isinstance(value, bool):
-            tags_list.append({key: value})
-        else:
-            for value_item in value:
-                tags_list.append({key: value_item})
-
-    # add node/way/relation query components one at a time
-    components = []
-    for d in tags_list:
-        for key, value in d.items():
-
-            if isinstance(value, bool):
-                # if bool (ie, True) just pass the key, no value
-                tag_str = f"['{key}'](poly:'{polygon_coord_str}');(._;>;);"
-            else:
-                # otherwise, pass "key"="value"
-                tag_str = f"['{key}'='{value}'](poly:'{polygon_coord_str}');(._;>;);"
-
-            for kind in ("node", "way", "relation"):
-                components.append(f"({kind}{tag_str});")
+        components.append(f"nwr.all{tag_str};")
 
     # finalize query and return
     components = "".join(components)
-    query = f"{overpass_settings};({components});out;"
+    query = f"{overpass_settings};nwr(poly:'{polygon_coord_str}')->.all;({components});(._;>;);out;"
 
     return query
 


gboeing:
See also #755

gboeing:
Thanks @derinwalters. This looks good and the test results in https://github.com/gboeing/osmnx/issues/862#issuecomment-1241617409 look pretty thorough. Can you hit [L484](https://github.com/gboeing/osmnx/pull/868/files#diff-5dd5790dd713a0f16f1c39d734a358048e16f2ea20b36e9014e911d80ca12e54R484) in the CI [tests](https://github.com/gboeing/osmnx/blob/main/tests/test_osmnx.py)?

@nickodell you've investigated a similar change to the codebase in the past, and I'm wondering if you have any comments or suggestions on this PR?

nickodell:
I'm running into a strange issue with this PR: It seems like it causes a 4x performance regression in the benchmark I ran for #755.

Here's the benchmarking script:

```
import geopandas as gpd
import osmnx as ox
import time

# use local mirror
ox.settings.overpass_endpoint='*********'
ox.settings.overpass_rate_limit=False
ox.settings.use_cache = False

counties = gpd.read_file('zip://tl_2020_us_county.zip')

def pull_osm(polygon):
    tags = {'amenity': ['hospital', 'fire_station', 'police']}
    gdf = ox.geometries_from_polygon(polygon, tags)
    return gdf

counties = counties.head(10)


def main():
    total_rows = counties.shape[0]
    for i, row in counties.iterrows():
        polygon = row['geometry']
        try:
            emergencyservice = pull_osm(polygon)
        except:
            print(row)
            raise
        print(f"Progress:  {i+1:2d}/{total_rows}")
        #print(emergencyservice.shape[0])

if __name__ == "__main__":
    start = time.time()
    main()
    end = time.time()
    print(f"Wallclock: {round(end - start, 2)}")
```

Requires this data file: https://www2.census.gov/geo/tiger/TIGER2020/COUNTY/

This benchmark reads in the US Census county boundaries, and takes the first ten counties, and searches for several different types of ambulance, police, or fire services within those counties. I would expect the change by @derinwalters to make this faster, since these shapefiles are very large, and his change means that I am sending one copy instead of nine.

 But these are my actual benchmark results.

```
/home/nick/programming/osm_benchmark/baseline_venv/bin/python3
Progress:   1/10
Progress:   2/10
Progress:   3/10
Progress:   4/10
Progress:   5/10
Progress:   6/10
Progress:   7/10
Progress:   8/10
Progress:   9/10
Progress:  10/10
Wallclock: 10.83
/home/nick/programming/osm_benchmark/dw_venv/bin/python3
Progress:   1/10
Progress:   2/10
Progress:   3/10
Progress:   4/10
Progress:   5/10
Progress:   6/10
Progress:   7/10
Progress:   8/10
Progress:   9/10
Progress:  10/10
Wallclock: 44.09
```

(Baseline is osmnx v1.2.2, for reference.)

@derinwalters Are you able to reproduce this issue?

gboeing:
@nickodell I used your test script (four runs each) and got these timings in seconds:

Timings on branch "v1.2.2": `42.32 36.95 41.09 38.89`

Timings on branch "pull/868": `157.37 128.36 141.02 124.85`

I am not using a local instance, hence the difference in timings between your test and mine. Nevertheless, I am reproducing your performance regression. @derinwalters are you able to reproduce this as well?

derinwalters:
I ran each seven times. The mean is higher for the development branch, but an unpaired t-test of the results did not indicate a statistically significant difference between the two.

Timings on branch "v1.2.2": 140.91 80.47 62.29 214.51 302.57 148.4 152.47
Timings on branch "pull/868": 160.52 180.86 202.48 312.35 208.93 174.45 159.76

The two-tailed P value equals 0.2691
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">


<body link="#0563C1" vlink="#954F72">
<meta charset=utf-8>

<div style='box-sizing: border-box;-webkit-box-flex: 0;flex: 0 0 auto;
min-height: 0px;min-width: 0px;font-variant-ligatures: normal;font-variant-caps: normal;
orphans: 2;widows: 2;-webkit-text-stroke-width: 0px;text-decoration-thickness: initial;
text-decoration-style: initial;text-decoration-color: initial'>



Group | PROD | DEV
-- | -- | --
Mean | 157.3743 | 199.9071
SD | 81.3196 | 53.0808
SEM | 30.7359 | 20.0627
N | 7 | 7



</body>

</html>

@gboeing I hit the L484 case in my own usage without issue. I've never gotten into test coverage before. Is the expectation that I add some kind of test to the pull request that would exercise it?

nickodell:
> @gboeing I hit the L484 case in my own usage without issue. I've never gotten into test coverage before.

[Here's a patch which adds coverage of that line to the test suite.](https://github.com/nickodell/osmnx/commit/596c76d9daa5d880d842f00fc9c0189c6cffc28d)

How I did this: I looked at the lines around L484, and looked for the test which exercised them. Then I looked at that test, and tried to modify it so that it also exercises L484. After doing so, I ran the commands:

```
coverage run --source ./osmnx --module pytest --verbose
coverage xml -i
coverage report -m
```

and checked that line 484 was no longer missing.

>The mean is higher for the development branch, but an unpaired t-test of the results did not indicate a statistically significant difference between the two.

I think there is a difference here, and it shouldn't be regarded as just random noise. A lack of statistical significance can mean two things: it can mean they're very similar, or it can mean that there's too much noise in the measurement. Given that the standard deviation in one of them is half as much as the mean, I think it's the second one.

If you'd like, I could share the IP address of my private Overpass mirror via e-mail, and you could do your tests without getting ratelimited or having to share it with a hundred other users. Would likely result in faster, lower-noise tests. Or, if you want to set up your own, the server is open-source: https://github.com/wiktorn/Overpass-API

derinwalters:
@nickodell Thanks for checking out the coverage issue. Most excellent!
> A lack of statistical significance can mean two things: it can mean they're very similar, or it can mean that there's too much noise in the measurement. Given that the standard deviation in one of them is half as much as the mean, I think it's the second one.
>
I'm inclined to agree with you. There is a real possibility of runtime increase in this use case that could be teased out with more runs. However, I think I've taken this as far as I'm willing to go since it's working for me and I'm not really dependent on osmnx anymore. I took an initial stab at debug by reverting the regular expression tag-aggregation back to a 1:1 tag-value style as in v1.2.2 but that made things worse, increasing to around 400 seconds. Seems like that would indicate an issue with either NWR and/or the .all polygon. Do you have any bandwidth to fiddle around with this?

gboeing:
Let me know if anyone has the inclination to pursue this further. I like the idea of using this `nwr` querying, but as it stands the negative hit on runtime for certain queries makes it infeasible to merge this in.

derinwalters:
> Let me know if anyone has the inclination to pursue this further. I like the idea of using this `nwr` querying, but as it stands the negative hit on runtime for certain queries makes it infeasible to merge this in.

Unfortunately, I'm out for the time being. I believe the simpler modification to NWR should be a no-brainer should someone want to pursue that.

nickodell:
@derinwalters Here's an idea: I could take the code you wrote to validate that the two approaches returned equivalent results, and the code from #755, and validate that `nwr` queries produce equivalent results. That was the sticking point last time. It would be a smaller performance improvement than your PR, but still an improvement.

I see you posted a diff, but I don't see how you produced the diff? Do you mind posting that code?

gboeing:
Hi all, I was going through old issues to close out inactive ones and I wanted to check back on this. Is there appetite for ongoing work to continue development on this PR, or is it inactive at this point? As I mentioned back in October:

> I like the idea of using this nwr querying, but as it stands the negative hit on runtime for certain queries makes it infeasible to merge this in.

Still the same. This would be a nice to have if it simplifies the code and reduces query time consistently, but only if it is indeed consistent. If there is no appetite to continue development on this, that's ok too.

derinwalters:
I'd suggest we close. The issue appears inactive and I'm currently off osmnx. My specific use case has been chugging along with this PR without issue for the past several months.

nickodell:
> Hi all, I was going through old issues to close out inactive ones and I wanted to check back on this. Is there appetite for ongoing work to continue development on this PR, or is it inactive at this point?

It's inactive.

PR #870, opened 2022-09-28
union statement for custom_filter

jannefleischer:
I had the problem that I have to analyse the bicycle quality of a city. For that I needed all small roads and big roads with cyclelanes. For that I needed a union statement in the filter and therefore I needed a fix.

This for overpass-turbo:

```overpassQL
[out:json][timeout:240];(way["highway"]["area"!~"yes"]["access"!~"private"]["highway"!~"abandoned|construction|planned|platform|proposed|raceway|motorway|trunk|primary|secondary|tertiary"]["service"!~"private"]({{bbox}});>;
                         way["highway"]["area"!~"yes"]["access"!~"private"]["cycleway"]["cycleway"!~"no"]["service"!~"private"]({{bbox}});>;);out;
```

Don't know if this quick fix will pass tests... lets see.

### DIFF
--- a/osmnx/downloader.py
+++ b/osmnx/downloader.py
@@ -518,8 +518,9 @@ def _osm_network_download(polygon, network_type, custom_filter):
         boundary to fetch the network ways/nodes within
     network_type : string
         what type of street network to get if custom_filter is None
-    custom_filter : string
-        a custom ways filter to be used instead of the network_type presets
+    custom_filter : string | list of strings
+        a custom ways filter to be used instead of the network_type presets. 
+        If a list of strings is provided, those a combined with the union criteria.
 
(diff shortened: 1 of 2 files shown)


gboeing:
Thanks for considering this contribution. The tests are not passing and this also likely conflicts with ongoing work in #868.

Note that you can already do this with OSMnx (though less efficiently than with a single Overpass query). Some examples:
  - https://stackoverflow.com/a/62720802
  - https://stackoverflow.com/a/62883614
  - https://stackoverflow.com/a/62239377

jannefleischer:
Uuh, didn't realize there is nx.compose ... So this pull request is not neccassary for me anymore.

PR #892, opened 2022-12-08
Add 'merged_edges' as a result column of the 'simplify_graph' function

davidmurray:
This is a pretty straightforward change.
Basically, when simplifying a graph, it's sometimes useful to know exactly which edges were merged.
In my research, one of the first things that I do is to simplify a huge OSM graph to make it a little more manageable performance-wise. However, sometimes I need to know exactly which edges were merged in order to do route matching with OSRM responses. Therefore, this simply adds a column to the graph created by the `simplify_graph` function to let one go back to the unsimplified graph from the simplified graph.

Note: this is not possible to do based only on the list of merged OSM IDs as these IDs can span multiple edges.

I did not provide as code example as I thought this was simple enough.

Thanks,
David


gboeing:
The tests are currently failing.

davidmurray:
@gboeing Should be fixed now :)

gboeing:
@davidmurray you might want to `git rebase main` to get CI working. Lastly, could you also provide some runtime averages to see if this creates any noticeable performance hit? Graph simplification can be computationally intense, so we just want to ensure nothing substantially hits the runtime when making changes to the algorithm. Thanks.

PR #894, opened 2022-12-14
Compatibility with shapely 2 and matplotlib 3.4

cheginit:
This PR fixes an issue with `shapely` 2 in the `osm_xml` module and another compatibility issue with `matplotlib` 3.4.

I also have another suggestion regarding `shapely` 2. Since `geopandas` throws a warning when both `pygeos` and `shapely` 2.0 are installed, we can add these lines to `__init__.py` to take care of the warning:

```python
from importlib.metadata import version

if int(version("shapely").split(".")[0]) > 1:
    import os

    os.environ["USE_PYGEOS"] = "0"
```

Note that `importlib` is a built-in Python 3.8+ module.

### DIFF
--- a/osmnx/geometries.py
+++ b/osmnx/geometries.py
@@ -15,6 +15,10 @@ import geopandas as gpd
 import numpy as np
 import pandas as pd
 from shapely.errors import TopologicalError
+try:
+    from shapely.errors import GEOSException
+except ImportError:
+    GEOSException = ValueError
 from shapely.geometry import LineString
 from shapely.geometry import MultiPolygon
 from shapely.geometry import Point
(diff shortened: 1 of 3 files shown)


gboeing:
> This PR fixes an issue with shapely 2 in the osm_xml module and another compatibility issue with matplotlib 3.4.

Can you fully document the implied compatibility issues and provide steps to reproduce them?

gboeing:
Further thoughts regarding the pygeos/shapely compatibility warning, I do not think that hiding it does the user any favors. Better for them to hide it themselves if they do not need to see it, like:
```python
import os
os.environ["USE_PYGEOS"] = '0'
import osmnx as ox
```

cheginit:
At the moment, [this](https://github.com/gboeing/osmnx/blob/c1ca82daf3ac64a274ec50f7826ba42ccac00c8e/tests/test_osmnx.py#L499) test fails, this PR fixes that issue. I ran `osmnx`'s test suite with both `shapely` 1.8 and 2 and they both pass. Regarding `matplotlib`, it seems that the only incompatibility at the moment is related to `colormaps`. But since you decided to drop support for it, I can remove that part if you decide to merge this PR.

gboeing:
Thanks. OSMnx v1.3+ will officially support Shapely >=2.0 and drop support for Shapely versions <2.0. See the [shapely2](https://github.com/gboeing/osmnx/tree/shapely2) branch and #900 for related fixes.

Issue #893, opened 2022-12-12
Enhancement: Get elevation data from opentopodata.org


aburgr:
Hi!

It would be great to add the possiblity to get elevantion data from opentopodata.org.

The API is identical to the maps.googleapis.com/maps API.
`https://api.opentopodata.org/v1/eudem25m?locations={}&key={}`

The advantage is that no user or api-key is necessary. The maximum locations per batch is still 100.

Best regards,
Andreas

gboeing:
Resolved by #901

PR #895, opened 2022-12-14
add add_node_elevations_opentopodata function

aburgr:
hence there are APIs between google and opentopodata are quite similar, they share the same underlying function.

for opentopodata it is possible to customize the endpoint because it can be hosted locally. by default it uses the public API. the dataset is customizeable as well because the user might want use a different one. the default is aster30m, which works worldwide to 30m

Solving issue:
https://github.com/gboeing/osmnx/issues/893

Example: 
`G = ox.add_node_elevations_opentopodata(G)`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
   - hard code Google DNS IP address
   - resolve matplotlib deprecation warning
   - deprecate save_graph_shapefile function
+  - add add_node_elevations_opentopodata function
 
 ## 1.2.2 (2022-08-05)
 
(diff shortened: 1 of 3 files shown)


gboeing:
Tests are failing.

We don't need a whole new function: it would make the change simpler if we just expose an `api_url="https://maps.googleapis.com/maps/api/"` argument in the `add_node_elevations_google` function that the user can change to point to a different Google-equivalent API host.

aburgr:
Tests are fixed now

I would prefer an own function because

- it is more explicit that not google and the api_key is used.
- For opentopodata.org the api_key is unused
- the api_key cannot be removed because of compatiblity reasons

Nevertheless i prepared another PR, if you still prefer the other solution: https://github.com/gboeing/osmnx/pull/901

Please choose and close the PR with the unprefered solution

gboeing:
closed in favor of #901

PR #901, opened 2022-12-15
enable elevation data from opentopodata.org

aburgr:
expose url_template that function add_node_elevations_google can be used for opentopodata.org

Issue:
https://github.com/gboeing/osmnx/issues/893

Example:
```
    G = ox.add_node_elevations_google(G, 
        None,
        "https://api.opentopodata.org/v1/aster30m?locations={}&key={}",
        max_locations_per_batch=100)
```


gboeing:
Thank you @aburgr!

Issue #915, opened 2023-01-05
Simplify save_graph_xml output


davidmurray:
**Do not delete the following template. Fill it out completely.**

  - [x] I have read the OSMnx [contributing guidelines](../blob/main/CONTRIBUTING.md)
  - [x] I have read the OSMnx [documentation](https://osmnx.readthedocs.io/) and [usage examples](https://github.com/gboeing/osmnx-examples)
  - [x] I have searched the open/closed issues and pull requests for others similar to mine

**Is your feature proposal related to a problem?**

The `save_graph_xml` function adds a lot of (unnecessary?) data to the resulting XML file:
https://github.com/gboeing/osmnx/blob/586cf2651af80c2cdb1c6e30b94d8176beea7c4a/osmnx/osm_xml.py#L207-L212

The resulting XML file has a lot of extra fields compared to regular OSM XML files :
```xml
<?xml version="1.0"?>
<osm version="0.6" generator="OSMnx">
  <node id="0" timestamp="2017-01-01T00:00:00Z" uid="1" user="osmnx" version="1" changeset="1" lat="45.5571316" lon="-73.5963673">
    <tag k="highway" v="nan"/>
  </node>
  <node id="1" timestamp="2017-01-01T00:00:00Z" uid="1" user="osmnx" version="1" changeset="1" lat="nan" lon="nan">
    <tag k="highway" v="nan"/>
  </node>
  <node id="2" timestamp="2017-01-01T00:00:00Z" uid="1" user="osmnx" version="1" changeset="1" lat="nan" lon="nan">
    <tag k="highway" v="nan"/>
  </node>
  <node id="3" timestamp="2017-01-01T00:00:00Z" uid="1" user="osmnx" version="1" changeset="1" lat="45.5575365" lon="-73.5889391">
    <tag k="highway" v="nan"/>
  </node>
  ...
</osm>
```

In addition, what's the point of the timestamp if it's always the same date in 2017? 

**Describe the solution you'd like to propose**

Either:
1. Adding a parameter to `save_graph_xml` to decide whether or not these extra fields should be added to the xml file, or:
2. Removing this code altogether.

Thanks!

davidmurray:
In addition, it seems that the function adds tags even when the values do not exist:
```xml
<node id="2" timestamp="2017-01-01T00:00:00Z" uid="1" user="osmnx" version="1" changeset="1" lat="nan" lon="nan">
    <tag k="highway" v="nan"/>
  </node>
```

[...]

PR #917, opened 2023-01-05
Improvements to save_as_xml function

davidmurray:
Somewhat related to issue #915.

Commits are quite self-explanatory, but this fixes a few things:
1. A bug where non-existent tags would get written to the XML file when they shouldn't.
2. Adds a way to specify the OSM version to write in the header
3. Adds a way to specify how many decimals to keep when writing lat-long coordinates
4. Sets the encoding to UTF-8. This is useful for areas that have non-ascii characters in the name (for example, in Québec)

Thanks

### DIFF
--- a/osmnx/osm_xml.py
+++ b/osmnx/osm_xml.py
@@ -166,9 +166,11 @@ def save_graph_xml(
         component edges. Otherwise, the length attribute will simply
         reflect the length of the first edge associated with the way.
     osm_version : int
-        OpenStreetMap data version to write in the XML file header. Default 0.6.
+        OpenStreetMap data version to write in the XML file header.
+        Default 0.6.
     coordinate_precision : int
-        Number of decimal places to keep when writing latitude and longitude values. Default 7.
+        Number of decimal places to keep when writing latitude and longitude values.
+        Default 7.
 
     Returns
     -------
--- a/osmnx/osm_xml.py
+++ b/osmnx/osm_xml.py
@@ -164,7 +164,9 @@ def save_graph_xml(
         this method to aggregate the lengths of the individual
         component edges. Otherwise, the length attribute will simply
         reflect the length of the first edge associated with the way.
-    osm_version : OpenStreetMap data version to write in the XML file header. Default 0.6.
+    osm_version : int
+        OpenStreetMap data version to write in the XML file header. Default 0.6.
+    
     Returns
     -------
     None
--- a/osmnx/osm_xml.py
+++ b/osmnx/osm_xml.py
@@ -203,8 +203,8 @@ def save_graph_xml(
 
     # rename columns per osm specification
     gdf_nodes.rename(columns={"x": "lon", "y": "lat"}, inplace=True)
-    gdf_nodes.lon = gdf_nodes.lon.round(coordinate_precision)
-    gdf_nodes.lat = gdf_nodes.lat.round(coordinate_precision)
+    gdf_nodes['lon'] = gdf_nodes['lon'].round(coordinate_precision)
+    gdf_nodes['lat'] = gdf_nodes['lat'].round(coordinate_precision)
     gdf_nodes = gdf_nodes.reset_index().rename(columns={"osmid": "id"})
     if "id" in gdf_edges.columns:
         gdf_edges = gdf_edges[[col for col in gdf_edges if col != "id"]]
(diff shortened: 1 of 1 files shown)


davidmurray:
Why don't the CI tests just lint the code directly instead of failing?

osmnx/osm_xml.py
```diff
@@ -163,6 +165,10 @@ def save_graph_xml(
         this method to aggregate the lengths of the individual
         component edges. Otherwise, the length attribute will simply
         reflect the length of the first edge associated with the way.
+    osm_version : int
+        OpenStreetMap data version to write in the XML file header. Default 0.6.
+    coordinate_precision : int
+        Number of decimal places to keep when writing latitude and longitude values. Default 7.
```
gboeing, line 171: Can you ensure any comment lines are <= 79 characters long?

osmnx/osm_xml.py
```diff
@@ -195,6 +201,8 @@ def save_graph_xml(
 
     # rename columns per osm specification
     gdf_nodes.rename(columns={"x": "lon", "y": "lat"}, inplace=True)
+    gdf_nodes.lon = gdf_nodes.lon.round(coordinate_precision)
+    gdf_nodes.lat = gdf_nodes.lat.round(coordinate_precision)
```
gboeing, line 205: For Geo/DataFrame columns, please access them via `gdf["col_name"]` rather than `gdf.col_name`.

osmnx/osm_xml.py
```diff
@@ -211,10 +219,6 @@ def save_graph_xml(
         table["changeset"] = "1"
         table["timestamp"] = "2017-01-01T00:00:00Z"
 
-    # convert all datatypes to str
-    gdf_nodes = gdf_nodes.applymap(str)
-    gdf_edges = gdf_edges.applymap(str)
-
```
gboeing, line 217: Were these lines only necessary for converting `osm_version` to a string? Is there no other edge case where we must convert values to string? I see you added a few other string conversions below: is this more efficient than `applymap(str)`?

gboeing (changes requested):

gboeing:
The test coverage report states:
```
codecov/patch — 92.85% of diff hit (target 97.85%) 
```
When you make the requested changes to the PR, would you check the test coverage?

osmnx/osm_xml.py
davidmurray: I think they were necessary because `xml.etree` expects strings only. The API allows you to input int and float types, but when you actually try to write the XML file, it fails saying that types must all be strings. Thus, everything needs to be a string before passing to `xml.etree`, but we need the original types to check for null / NaN.

From what I can see, these changes do not break any other edge case based on the testing I've done locally.

davidmurray:
> The test coverage report states:
> 
> ```
> codecov/patch — 92.85% of diff hit (target 97.85%) 
> ```
> 
> When you make the requested changes to the PR, would you check the test coverage?

@gboeing I'm not sure that I understand how the tests work.
For example, these lines in the tests:
https://github.com/gboeing/osmnx/blob/5b87c2b14d78425e9d0fe2d485b8a4f5c75384b3/tests/test_osmnx.py#L171-L172

Unless I am mistaken, this does not test if the output is valid. It merely ensures that the code hasn't produced any exceptions. A real test suite would look at the written data and ensure that it is valid OSM xml. Is this correct?
In this case, how would you suggest that I modify the code to improve the test coverage?

Thank you very much

gboeing:
> Unless I am mistaken, this does not test if the output is valid. It merely ensures that the code hasn't produced any exceptions. A real test suite would look at the written data and ensure that it is valid OSM xml. Is this correct?
In this case, how would you suggest that I modify the code to improve the test coverage?

1) yes it looks like the XML tests are just checking for exceptions. At a minimum, could you ensure that the lines in your diff are hit by the tests and no exceptions result?

2) ideally yes it would be great to validate the XML output. I do not work with OSM XML files myself so my expertise there is limited. Are you aware of any validator tools?

gboeing:
Let me know if this PR is still under development and should remain open.

davidmurray:
> Let me know if this PR is still under development and should remain open.

Hi, yeah I haven't had time to finish it up but I'll get it done soon.

gboeing:
I've been going through and closing out stale PRs and issues. Let me know if this remains active.

davidmurray:
@gboeing Sorry, don't think I'll have time to fix the unit tests. However, the improvements in this PR are valuable and I think they should be merged (or considered) to be merged in the main tree. Cheers

gboeing:
> this does not test if the output is valid. It merely ensures that the code hasn't produced any exceptions. A real test suite would look at the written data and ensure that it is valid OSM xml

#1135 adds an OSM XML schema XSD file and now validates saved OSM XML files against it in the test suite.

PR #924, opened 2023-01-26
Fix downloader._get_host_by_name when the request fails.

a-gerhard:
I am running osmnx in a private network with a very restrictive proxy server which won't allow the HTTP/HTTPS request to Google DNS servers, so the proxy server would return a 403 error directly. This resulted in an uncaught `ProxyError` thrown when trying to resolve the hostname this way, making osmnx unusable in this environment.

I added a try-except clause around the requests methods to make sure that any requests-related exception thrown during the request is caught, and used the same error handling as used in the function's status code check.

I have tested this fix in said environment, and it's working flawlessly.

### DIFF
--- a/osmnx/downloader.py
+++ b/osmnx/downloader.py
@@ -267,8 +267,14 @@ def _get_host_by_name(host):
         resolved IP address
     """
     dns_url = f"https://8.8.8.8/resolve?name={host}"
-    response = requests.get(dns_url)
-    data = response.json()
+    try:
+        response = requests.get(dns_url)
+        data = response.json()
+    except requests.exceptions.RequestException as e:
+        # if anything goes wrong trying to reach the Google DNS servers, (e.g.,
+        # a proxy is restricting access), return the host itself
+        utils.log(f"Got an error when trying to resolve {host!r} from Google DNS: {str(e)}")
+        return host
 
     # status = 0 means NOERROR: standard DNS response code
     if response.ok and data["Status"] == 0:
--- a/osmnx/downloader.py
+++ b/osmnx/downloader.py
@@ -266,6 +266,9 @@ def _get_host_by_name(host):
     ip_address : string
         resolved IP address
     """
+    if settings.dns_http_url is None:
+        return host
+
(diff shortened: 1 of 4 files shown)


a-gerhard:
I am a bit unsure how to write a test for catching this exception in any environment. Maybe make the DNS server URL configurable or pass it as argument to the function?

gboeing:
@a-gerhard yes, it may be a good idea to move `dns_url` into the `settings` module to allow it be configurable.

pmlpm1986:
This PR appears to work and fixes a different problem for me: I have been trying to use osmnx using cache only and offline but the _config_dns method always throws an error unless I am online; with this PR, it no longer does; my alternative was just to catch any exception associated with _config_dns and ignore it; this solution seems more comprehensive and way less hacky. PS: I am aware that the cache only mode (cache_only_mode) is meant for something slightly different.

The PR appears to be blocked due to some linting issues (' instead of "). I added a PR to the for by @a-gerhard to get it going again.

gboeing:
@pmlpm1986 can you link to your PR? I don't see it.

pmlpm1986:
I didn't want to redo all the work @a-gerhard did so I just added a PR to original fork: https://github.com/a-gerhard/osmnx/pull/1

osmnx/downloader.py
```diff
@@ -266,9 +266,18 @@ def _get_host_by_name(host):
     ip_address : string
         resolved IP address
     """
-    dns_url = f"https://8.8.8.8/resolve?name={host}"
-    response = requests.get(dns_url)
-    data = response.json()
+    if settings.dns_http_url is None:
+        return host
+
+    dns_url = settings.dns_http_url % f'name={host}'
```
pmlpm1986, line 272: This appears to be the error. I fixed it in: https://github.com/a-gerhard/osmnx/pull/1

a-gerhard:
It looks like I missed some notifications (linter error + pull request), sorry about that!

Thank you for helping out @pmlpm1986, I just merged your pull request.

gboeing:
Thanks all.

PR #947, opened 2023-05-07
Update graph.py

Georgepop:
II used new   for read pbf from https://download.geofabrik.de/russia.html
after use of osmosis

Issue #946, opened 2023-05-07
Osmnx reading .osm file


Georgepop:
Hello !
I read xml file like .osm
I use osmosis to convert file pbf to .osm  and then open with  ox.graph_from_xml(

But, then, I get an error TypeError: loop of ufunc does not support argument 0 of type float which has no callable deg2rad method xml

gboeing:
Thanks for using OSMnx! Please re-read the [contributing guidelines](../blob/main/CONTRIBUTING.md) and the new issue template you saw when you opened this issue. As the template noted, your issue will be automatically closed if you didn't provide all of the information requested in the template, because we need *all* of this information to address your issue.

PR #437, opened 2020-04-29
Informative error message for data import

anisotropi4:
Patch to provide more informative error message with graph_from_file data import

fixes #323 and #436

PR #948, opened 2023-05-08
Update graph.py

Georgepop:
We tried to import ".osm" file from file from https://download.geofabrik.de/russia.html
But I got error: TypeError: loop of ufunc does not support argument 0 of type float which has no callable deg2rad method
So I downloaded the whole code from github and edited it. Found an error graph.py which I reproduced

Here is a file, to reproducable with:
[mo_0.zip](https://github.com/gboeing/osmnx/files/11424790/mo_0.zip)

### DIFF
--- a/osmnx/graph.py
+++ b/osmnx/graph.py
@@ -572,6 +572,8 @@ def _create_graph(response_jsons, retain_all=False, bidirectional=False):
     # add each osm node to the graph
     for node, data in nodes.items():
         G.add_node(node, **data)
+        
+    paths = {k: v for k, v in paths.items() if len(set(v['nodes'])-set(list(nodes.keys()))) == 0}
 
     # add each osm way (ie, a path of edges) to the graph
     _add_paths(G, paths.values(), bidirectional)


gboeing:
As mentioned in the previous issue (#946) and PR (#947), we need the complete details requested in the new issue template to investigate this issue. You mentioned that this XML file was downloaded from https://download.geofabrik.de/russia.html. However, the file you shared above does not match any of the files available for download from that page. When I test [this file](https://download.geofabrik.de/russia/kaliningrad-latest.osm.bz2) that is available for download, for example, it loads just fine:

```python
import osmnx as ox
G = ox.graph_from_xml("./kaliningrad-latest.osm.bz2")
print(len(G))
```

I cannot know what is going on with your data file without complete troubleshooting details. My best guess is that you clipped it at some point, resulting in ways in the XML referencing nodes that are not present in the XML. Indeed, looking into your XML file itself, way ID 399296135 references node ID 4020273366, but there is no such node present in the XML file.

gboeing:
See also #437.

gboeing:
PR superseded by #950.

Georgepop:
Thank you very much for your precisious look. You are very kind to give such a much attetion to it... !
Though the path is 'nodes': [4020273337, 4020273356, 4020273352, 4020273353, 4020273357, 4020273354, 4020273358, 4020273355, 4020273373, 4020273365, 4020273372, 4020273364, 4020273371, 4020273363, 4020273362, 4020273370, 4020273361, 4020273369, 4020273368, 4020273360, 4020273367, 4020273366, 4020273359] -starting from 10th node is not present . We can cut the paths, or either remove the whole edge. I looked at the picture, it looks like the same. So why we include such an option ?
'
> user clips their .osm XML file

' - And yes I cliped the file, so it could be loaded to osm.
I rewrote a little this function, like this:

```
for k, v in paths.items():
        x = [i for i in v['nodes'] if i in list(nodes.keys())]
        if len(x):
            v['nodes'] = x
        else:
            paths.drop(k)
```

  and suggest delete nodes that have no connection to ohters

Issue #436, opened 2020-04-29
More informative error message with graph_from_file data import


anisotropi4:
#### Importing using `graph_from_file` where edge data contains missing node raises a key error 

As reported by @bradleyswilson in #323 KeyError: 'y' when poorly clipped OSM data (where edges are present without corresponding nodes) is imported using `graph_from_file` this raises a `KeyError 'y'` exception in the `add_edge_lengths`. 

#### Fails with a `KeyError: 'y'` when `G.nodes[{u,v}]` does not exist but the edge `u, v, k` does:
```python
coords = np.array([[u, v, k, G.nodes[u]['y'], G.nodes[u]['x'], G.nodes[v]['y'], G.nodes[v]['x']] for u, v, k in G.edges(keys=True)])
```
The proposed fix places this in a`try/except KeyError`block, identifies the missing node(s) referenced by edges and executes a `raise TypeError` with a more informative error message:

#### Handles the `KeyError` and raises more informative clipping error `TypeError` detailing missing nodes
```python
    try:
        coords = np.array([[u, v, k, G.nodes[u]['y'], G.nodes[u]['x'], G.nodes[v]['y'], G.nodes[v]['x']] for u, v, k in G.edges(keys=True)])
    except KeyError:
        missing_nodes = {str(i) for u, v, _ in G.edges(keys=True) if not(G.nodes[u] or G.nodes[u]) for i in (u, v) if not G.nodes[i]}
        raise TypeError('Edge missing nodes {} likely due to a clipping issue'.format(','.join(missing_nodes)))
```

gboeing:
@anisotropi4 thanks. I'm going to move your comment to the PR (#437) where we can discuss, and close this issue as it duplicates #323.

Issue #323, opened 2019-08-07
KeyError: 'y'


bradleyswilson:
Greetings. Appreciate the great package. I'm struggling to graph_from_file to run, getting a key error (see Traceback below). Running osmnx in a clean conda environment with strict channel priority as recommended by the install docs. 

The .osm file is the raw New Zealand Geofrabrik download, clipped with osmconvert to a bounding box. Any thoughts for troubleshooting?

```python
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-6-cc08ebb90c6e> in <module>
----> 1 G = ox.graph_from_file("kaikoura_clip.osm")

~/anaconda3/envs/ox/lib/python3.7/site-packages/osmnx/core.py in graph_from_file(filename, bidirectional, simplify, retain_all, name)
   1909     # create graph using this response JSON
   1910     G = create_graph(response_jsons, bidirectional=bidirectional,
-> 1911                      retain_all=retain_all, name=name)
   1912 
   1913     # simplify the graph topology as the last step.

~/anaconda3/envs/ox/lib/python3.7/site-packages/osmnx/core.py in create_graph(response_jsons, name, retain_all, bidirectional)
   1366     # use as weight
   1367     if len(G.edges) > 0:
-> 1368         G = add_edge_lengths(G)
   1369 
   1370     return G

~/anaconda3/envs/ox/lib/python3.7/site-packages/osmnx/core.py in add_edge_lengths(G)
   1199     # first load all the edges' origin and destination coordinates as a
   1200     # dataframe indexed by u, v, key
-> 1201     coords = np.array([[u, v, k, G.nodes[u]['y'], G.nodes[u]['x'], G.nodes[v]['y'], G.nodes[v]['x']] for u, v, k in G.edges(keys=True)])
   1202     df_coords = pd.DataFrame(coords, columns=['u', 'v', 'k', 'u_y', 'u_x', 'v_y', 'v_x'])
   1203     df_coords[['u', 'v', 'k']] = df_coords[['u', 'v', 'k']].astype(np.int64)

~/anaconda3/envs/ox/lib/python3.7/site-packages/osmnx/core.py in <listcomp>(.0)
   1199     # first load all the edges' origin and destination coordinates as a
   1200     # dataframe indexed by u, v, key
[...]

PR #950, opened 2023-05-11
ensure node coordinates are non-null

### LINKED ISSUE #946: Osmnx reading .osm file

gboeing:
Resolves #946.
See also #947 #948 #437 #436 #323.

If a user clips their .osm XML file before loading with the `graph_from_xml` function, it can result in ways in the XML file referencing nodes that are not present. This causes exceptions in the `add_edge_lengths` function. See #948 for an example: way ID 399296135 references node ID 4020273366, but there is no such node present in the XML file.

#437 previously added a descriptive error message when missing nodes could not be found. This PR adds another check to see if all the nodes' coordinates are non-null and convertible to float, and throws an error message if not. @Georgepop see proposed resolution here.

### DIFF
--- a/osmnx/distance.py
+++ b/osmnx/distance.py
@@ -147,8 +147,10 @@ def add_edge_lengths(G, precision=3, edges=None):
     try:
         # two-dimensional array of coordinates: y0, x0, y1, x1
         c = np.array([(y[u], x[u], y[v], x[v]) for u, v, k in uvk])
-    except KeyError:  # pragma: no cover
-        raise KeyError("some edges missing nodes, possibly due to input data clipping issue")
+        # ensure all coordinates can be converted to float and are non-null
+        assert not np.isnan(c.astype(float)).any()
+    except (AssertionError, KeyError):  # pragma: no cover
+        raise ValueError("some edges missing nodes, possibly due to input data clipping issue")
 
     # calculate great circle distances, round, and fill nulls with zeros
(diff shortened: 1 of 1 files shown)


Georgepop:
Thank you very much for your precisious look. You are very kind to give such a much attetion to it... !
Though the path is 'nodes': [4020273337, 4020273356, 4020273352, 4020273353, 4020273357, 4020273354, 4020273358, 4020273355, 4020273373, 4020273365, 4020273372, 4020273364, 4020273371, 4020273363, 4020273362, 4020273370, 4020273361, 4020273369, 4020273368, 4020273360, 4020273367, 4020273366, 4020273359] -starting from 10th node  is not present . We can cut the paths, or either remove the whole edge. I looked at the picture, it looks like the same. So why we include such an option ?
"

> user clips their .osm XML file

" - And yes I cliped the file, so it could be  loaded to osm.
I rewrote a little this function, like this:
```
for k, v in paths.items():
        x = np.asarray(v['nodes'])
        x2 = np.where( np.isin(x,list(nodes.keys()) ) )[0]
        if len(x2):
            x = list(x[:x2.max()] if x2.mean()<0.5* len(x) else x[x2.min():])
            v['nodes'] = x
        else:
            paths.drop(k)
```

PR #957, opened 2023-05-22
Deprecate folium module in favor of GeoPandas explore

### LINKED ISSUE #802: Include Point geometries in folium module to plot nodes interactively
Hi @gboeing!
I'm working on a `streamlit` project where users can get bbox coordinates by clicking on a web map. The main goal is to let them  create their own street networks and get some relevant stats to study the mobility in the target area.  

The [folium](https://github.com/gboeing/osmnx/blob/main/osmnx/folium.py) module looks great to add some interactivity to the analysis but I faced that it only supports LineStrings geometries. And since I do not want to only describe the graph with edges but also with nodes, I was looking for some extra methods to allow plotting them based on their connectivity attributes.

Specifically, I needed to render Point geometries on a Leaflet web map and get them colored by some attribute (betweenness centrality, degree centrality, node degree, etc) with the `cmap` bar legend/reference. Something like this:

image

Below I share you the code and some examples:

1. [build a Point geoms gdf and add node colors based on some attribute](https://github.com/PyMap/ssyt/blob/main/ssyt/mobilidad.py#L56-L89)
2. [create a circle marker map based on color attributes](https://github.com/PyMap/ssyt/blob/main/ssyt/mobilidad.py#L91-L157)
3. [some examples can be found here](https://github.com/PyMap/ssyt/blob/main/ssyt/accesibility.ipynb)
 
If you find this could be an interesting feature to add in the folium.py module we can have some further discussions about the CircleMarker map implementation. 

Thanks!

### LINKED ISSUE #935: replace the folium module with geopandas.GeoDataFrame.explore
Shapely and GeoPandas have recently had some updates that raise a couple broad questions for OSMnx:

1. Is OSMnx's `folium` module useful given `geopandas.GeoDataFrame.explore`'s functionality?

I think we should deprecate the `folium` module in favor of using `geopandas.GeoDataFrame.explore` directly.

2. Is OSMnx's coordinate precision functionality useful given `shapely.set_precision`'s functionality?

No action for this one. I think the answer is yes. In particular, we use this to minimize file size when serializing graphs that otherwise would have attributes with arbitrarily fine precision. It may be worth considering moving precision from function-level arguments to the settings module at some point.

3. Should the `plot_orientation` function be moved from the bearing module to the plot module.

For consistency yes. See #956 for this work.

gboeing:
Resolves #802 and #935.

- [x] Deprecates the `folium` module in favor of using the more powerful and flexible `GeoPandas.GeoDataFrame.explore` directly. [GeoPandas docs](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.explore.html). [See also](https://github.com/geopandas/geopandas/pull/1953).
- [x] Adds a new `route_to_gdf` helper function to the `utils_graph` module to make it easy to explore a route
- [x] Deprecates the `get_route_edge_attributes` function as it becomes redundant with the superior `route_to_gdf`

See https://github.com/gboeing/osmnx-examples/pull/64 for demonstration code in the examples repo. You can:
- interactively explore nodes, edges, or both together, optionally colored by attribute values
- interactively explore a route
- interactively explore building footprints, transit, amenities, etc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 1.3.1 (TBD)
 
+  - deprecate folium module in favor of using GeoPandas.GeoDataFrame.explore directly
+  - add route_to_gdf function to utils_graph module
   - improve DNS resolution when using proxies or on networks blocking DNS-over-HTTPS
   - improve processing of per-lane values when adding edge speeds
   - ensure node coordinates are non-null and covertible to float in the add_edge_lengths function
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,16 @@
 # Change log
 
-## 1.3.1 (TBD)
+## 1.4.0 (TBD)
 
-  - deprecate folium module in favor of using GeoPandas.GeoDataFrame.explore directly
   - add route_to_gdf function to utils_graph module
+  - deprecate folium module in favor of using GeoPandas.GeoDataFrame.explore directly
+  - move plot_orientation function from bearing module to plot module
+
+## 1.3.1 (2023-05-24)
+
   - improve DNS resolution when using proxies or on networks blocking DNS-over-HTTPS
   - improve processing of per-lane values when adding edge speeds
+  - improve file writing in save_graph_xml function
   - ensure node coordinates are non-null and covertible to float in the add_edge_lengths function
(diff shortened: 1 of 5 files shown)


gboeing:
@PyMap I believe this covers all the uses you had considered in #802. See https://github.com/gboeing/osmnx-examples/pull/64 for demonstrations.

Issue #1016, opened 2023-06-27
Save memory by streaming more and duplicating less


McToel:
**Is your feature proposal related to a problem?**

For huge graphs, I found that OSMnx uses a lot of memory. That is not too surprising, but looking into the code, it seems like there at some points, there are data duplicates lying around.

**Describe the solution you'd like to propose**

A long time ago, I started to rewrite the download module to yield single chunks instead of returning a list with all chunks. This improves the memory usage, but many of the log messages are no longer possible.

gboeing:
Thanks for the suggestion. If you have identified specific opportunities for and techniques for memory efficiency improvements, would you provide the details? What specific changes to the code would it entail? What is the magnitude of the efficiency improvement?

> it seems like there at some points, there are data duplicates lying around

Can you identify these specific points?

McToel:
> Can you identify these specific points?
For example, in `graph.py`, line 568...
```py
for response_json in response_jsons:
        nodes_temp, paths_temp = _parse_nodes_paths(response_json)
        nodes.update(nodes_temp)
        paths.update(paths_temp)
```
The content of `response_jsons` is kept but not used again, which fills up memory.

If `response_jsons` was yielded chunk by chunk, it would be consumed by the for loop and no "duplicates" would happen.

For this to make a difference, more parts of the process of creating a graph would have to be rewritten in that streaming fashion.

gboeing:
Have you pinpointed this specific snippet as the source of the memory inefficiency? Can you provide any profiling and measurements? Like what is the magnitude of the memory inefficiency the current way vs using a generator? If it's trivial it may not be worth refactoring. But if this is a source of significant inefficiency, a refactor would be great. This should be a relatively straightforward refactor to make `_downloader._osm_network_download` use a new function that yields a generator of `response_jsons`.

McToel:
This single snippet bumped my memory usage from 5 to 8 GB for a specific problem. However, this will not solve other memory inefficiencies. For significant improvements, more refactoring would be needed. I could try to achieve as much as possible, but I will not put the effort into it if there is no chance for these changes being useful for OSMnx.

McToel:
OK, I looked into it again. I think I could reduce the memory usage of _create_graph by 50-70% (in my case 15 to 4 GB).

gboeing:
>  I think I could reduce the memory usage of _create_graph by 50-70%

Wonderful! That sounds like a great improvement. I think this could be a useful enhancement to make OSMnx more efficient.

PR #1021, opened 2023-06-28
Stream downloads to save memory

McToel:
This addresses #1016.

The changes were way simpler than I thought, but it only saves about 33% of memory. That's still quite good, I think.

Due to the streaming, I had to remove a log message, as it is no longer possible to count the downloaded data before consuming it.

Any suggestions for code improvements are welcome, and I have not run any benchmarks yet, on how much memory it is saving. The 33% come from looking at the memory usage of Python in my system monitor.


gboeing:
@McToel looks like tests are failing.

PR #1029, opened 2023-07-03
Better memory efficiency when downloading/constructing graph

### LINKED ISSUE #1016: Save memory by streaming more and duplicating less

gboeing:
Builds on #1021 and resolves #1016.

@McToel I merged your PR into a feature branch for more testing. There were a couple of bugs:
  - it broke `ox.settings.cache_only_mode=True` functionality
  - it added duplicate edges to a graph when making multiple requests to the server

I addressed both of those issues in this PR and restored a log message. It's slightly less memory efficient that your branch, but more memory efficient than main currently is.

You can use a code snippet like this to test:
```python
import osmnx as ox
ox.settings.log_console = True
ox.settings.use_cache = True
ox.settings.cache_only_mode = False
place = "Los Angeles County, CA, USA"
G = ox.graph.graph_from_place(place, network_type="drive")
len(G.nodes), len(G.edges)
```

The result should be `(174692, 461602)` but on your branch it was `(186022, 512722)` due to edge duplication. The solution necessitated accumulating nodes/paths as dicts in a for loop while consuming the `response_jsons` generator, which increases memory, but not by a lot because the dicts themselves are relatively compact and memory efficient.

Please take a look and test it out.

### DIFF
--- a/osmnx/graph.py
+++ b/osmnx/graph.py
@@ -580,12 +580,13 @@ def _create_graph(response_jsons, retain_all=False, bidirectional=False):
     utils.log(f"Retrieved all data from API in {response_count} request(s)")
     if settings.cache_only_mode:  # pragma: no cover
         # after consuming all response_jsons in loop, raise exception to catch
-        raise CacheOnlyModeInterrupt("settings.cache_only_mode=True")
+        raise CacheOnlyModeInterrupt("Interrupted because `settings.cache_only_mode=True`")
 
     # ensure we got some node/way data back from the server request(s)
     if (len(nodes) == 0) and (len(paths) == 0):  # pragma: no cover
-        msg = "There are no data elements in the server response. Check log and query location/filters."
-        raise EmptyOverpassResponse(msg)
+        raise EmptyOverpassResponse(
+            "No data elements in server response. Check query location/filters and log."
+        )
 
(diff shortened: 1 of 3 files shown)


osmnx/graph.py
```diff
@@ -560,41 +561,52 @@ def _create_graph(response_jsons, retain_all=False, bidirectional=False):
...
-    paths = {}
-    for response_json in response_jsons:
-        nodes_temp, paths_temp = _parse_nodes_paths(response_json)
-        nodes.update(nodes_temp)
-        paths.update(paths_temp)
-
-    # add each osm node to the graph
+    # add each OSM node and way (a path of edges) to the graph
+    utils.log(f"Creating graph from {len(nodes):,} OSM nodes and {len(paths):,} OSM ways...")
     for node, data in nodes.items():
```
McToel, line 601: Is there a reason why you don't use `G.add_nodes_from`?

McToel:
OK, I get that the nodes and paths need to be unduplicated.

I think the changes could be merged

PR #1044, opened 2023-07-25
deprecate buffer_dist and clean_periphery arguments

gboeing:
This PR:
  - deprecates the `buffer_dist` function parameter in the `graph`, `features`, and `geocoder` modules. It is seldom used, it clutters the codebase, and its functionality is easily reproduced by the user with ~1 line of code.
  - deprecates the `clean_periphery` function parameter in the `graph` module: the equivalent of `clean_periphery=True` will the the uniform behavior going forward. This is the only way to generate an accurate graph on the perimeter, and the only way to get accurate stats, so it really makes no sense to have a `False` option here. See also (#775).
  - updates the docs with some explanatory details regarding (now-standard) periphery cleaning

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - fix DNS resolution in Dask clusters (#1039)
 - improve memory efficiency during features GeoDataFrame creation (#1043)
 - handle the optional settings.cache_only_mode in the features module (#1043)
+- deprecate the buffer_dist and clean_periphery function parameters throughout (#1044)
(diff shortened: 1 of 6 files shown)


EwoutH:
> * its functionality is easily reproduced by the user with ~1 line of code.

What would be the proper way to replace `buffer_dist`? (in this case with `graph_from_place`)

gboeing:
Yes that example does the job. My initial comment was overly conservative... it takes ~3 lines of code. Though it could be a single line of code for simpler use cases.

gboeing:
See also the new `buffer_geometry` helper function added in #1214.

Issue #1067, opened 2023-10-14
Dictionaries attributes not handled in io.load_graphml()


ncotie:
### Contributing guidelines

- [X] I understand the contributing guidelines

### Documentation

- [X] My problem is not addressed by the documentation or examples

### Existing issues

- [X] My problem does not appear in an existing issue

### What operating system and Python version are you using?

Win11, Python 3.10.10

### What OSMnx version are you using?

1.6.0

### Environment packages and versions

```shell
#
# Name                    Version                   Build  Channel
affine                    2.4.0              pyhd8ed1ab_0    conda-forge
anyio                     3.6.2              pyhd8ed1ab_0    conda-forge
argon2-cffi               21.3.0             pyhd8ed1ab_0    conda-forge
argon2-cffi-bindings      21.2.0          py310h8d17308_3    conda-forge
asttokens                 2.2.1              pyhd8ed1ab_0    conda-forge
attrs                     23.1.0             pyh71513ae_1    conda-forge
babel                     2.12.1             pyhd8ed1ab_1    conda-forge
backcall                  0.2.0              pyh9f0ad1d_0    conda-forge
backports                 1.0                pyhd8ed1ab_3    conda-forge
backports.functools_lru_cache 1.6.4              pyhd8ed1ab_0    conda-forge
beautifulsoup4            4.12.2             pyha770c72_0    conda-forge
bleach                    6.0.0              pyhd8ed1ab_0    conda-forge
blosc                     1.21.3               hdccc3a2_0    conda-forge
boost-cpp                 1.78.0               h9f4b32c_3    conda-forge
branca                    0.6.0              pyhd8ed1ab_0    conda-forge
brotli                    1.0.9                hcfcfb64_8    conda-forge
brotli-bin                1.0.9                hcfcfb64_8    conda-forge
brotlipy                  0.7.0           py310h8d17308_1005    conda-forge
bzip2                     1.0.8                h8ffe710_4    conda-forge
ca-certificates           2023.7.22            h56e8100_0    conda-forge
cairo                     1.16.0            hdecc03f_1015    conda-forge
certifi                   2023.7.22          pyhd8ed1ab_0    conda-forge
cffi                      1.15.1          py310h628cb3f_3    conda-forge
cfitsio                   4.2.0                h9ebe7e4_0    conda-forge
charset-normalizer        3.1.0              pyhd8ed1ab_0    conda-forge
click                     8.1.3           win_pyhd8ed1ab_2    conda-forge
[...]

PR #1075, opened 2023-10-18
Add support for automatic dict and set attributes in module `io`

ncotie:
Proposing some extra code support in module `io` for decoding stringified lists, dicts and sets automatically when running load_graphml(), allowing users to avoid having to specify `ast.literal_eval` as `dtype` for customized (non-OSM) node and/or edge attributes.

The proposed code expands existing code from `_convert_edge_attr_types()` that handles stringified lists as edge attributes.

This is linked to Issue [#1067](https://github.com/gboeing/osmnx/issues/1067).

Changes are made in _convert_node_attr_types() and _convert_edge_attr_types(), adding the following to both:

        # check for stringified dicts or sets: both can be converted
        # to python objects using ast.literal_eval()
        if value.startswith("{") and value.endswith("}"):
            with contextlib.suppress(SyntaxError, ValueError):
                data[attr] = ast.literal_eval(value)

Tested with toy networks as in the issue description, for list, set and dict node and edge attributes.


osmnx/io.py
```diff
@@ -419,8 +419,25 @@ def _convert_node_attr_types(G, dtypes=None):
...
+            # check for stringified lists
+            if value.startswith("[") and value.endswith("]"):
+                with contextlib.suppress(SyntaxError, ValueError):
+                    data[attr] = ast.literal_eval(value)
+            # check for stringified dicts or sets: both can be converted
+            # to python objects using ast.literal_eval()
+            if value.startswith("{") and value.endswith("}"):
+                with contextlib.suppress(SyntaxError, ValueError):
+                    data[attr] = ast.literal_eval(value)
+
```
gboeing, line 440: We previously only ran `literal_eval` on edge attributes. Is there a use case for node attributes? OSMnx does not produce any node attributes of type list, dict, or set.

osmnx/io.py
```diff
@@ -440,18 +457,28 @@ def _convert_edge_attr_types(G, dtypes=None):
...
     for _, _, data in G.edges(data=True, keys=False):
         # remove extraneous "id" attribute added by graphml saving
         data.pop("id", None)
 
-        # first, eval stringified lists to convert them to list objects
+        # first, eval stringified lists, dicts or sets to convert them to objects
         # edge attributes might have a single value, or a list if simplified
         for attr, value in data.items():
+            # check for stringified lists
             if value.startswith("[") and value.endswith("]"):
```
gboeing, line 474: This could probably be simplified to a single `if` block like `if (value.startswith("[") and value.endswith("]")) or (value.startswith("{") and value.endswith("}")):`

gboeing (changes requested):

gboeing:
CI tests are failing. You can check the failure here and correct it manually, or see the scripts in the `tests` folder.

osmnx/io.py
ncotie: I can describe my usage, if it makes sense to you.  It's clearly all custom attributes, as I mentioned in the issue description, so not necessarily OSMnx scope.

I'm doing 15-minute city metrics, based on OSMnx with a lot of my own code on top.  I fetch a walking network via OSMnx, then integrate amenity POI nodes into it, adding linking edges with lengths.  The network nodes, apart from the amenities, are considered to be "residential locations", without getting into filtering details.  Processing then obtains, among other things, shortest path distances, per residential node, to each amenity node.  These are stored per node into a dict of dicts attribute keyed by amenity category and amenity node.  
Accessible amenities are counted per residential node and edge, also, which allows for producing isochrone buffered geometries, for example.

It's true that I could have managed these outside the graph object using dicts, but I had wanted to make the work graph-oriented, and I believe it simplifies things, avoiding the need to keep separate objects consistent when doing things like truncating, etc, etc.

osmnx/io.py
```diff
@@ -440,18 +457,28 @@ def _convert_edge_attr_types(G, dtypes=None):
     Returns
     -------
     G : networkx.MultiDiGraph
+    
+    Modified to include detection of list, dict, and set type
+    node attribute strings and handle their translation into
+    python objects of the correct type.
```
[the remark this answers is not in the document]
ncotie: Correct, there was no need for me to change the docstring.

osmnx/io.py
```diff
@@ -419,8 +419,25 @@ def _convert_node_attr_types(G, dtypes=None):
...
```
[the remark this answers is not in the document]
ncotie: Correct, there was no need for me to change the docstring.

osmnx/io.py
ncotie: Very true.  I'd done them separately, more for readability than anything, but that's subjective at best.

gboeing:
Tests are failing. Please lint and test your code locally before pushing. See the `tests` folder.

ncotie:
Geoff, apologies but the push earlier today surprised me, I didn't know that when I pushed to my own fork that it would update here immediately...  Impressive integration in git...
As noted in the last merge, I created a local env from /tests/env-ci.yml, and looked to see how to test it.  Called format.sh and lint-test.sh, which made two formatting changes automatically, but I saw that by coincidence, and I can't find any other results from the local test, so I can't tell if it passes properly now or not.

osmnx/io.py
```diff
@@ -421,6 +421,16 @@ def _convert_node_attr_types(G, dtypes=None):
     G : networkx.MultiDiGraph
     """
     for _, data in G.nodes(data=True):
+        # first, eval stringified lists, dicts or sets to convert them to objects
+        # edge attributes might have a single value, or a list if simplified
```
gboeing, line 425: This comment is specific to edge attributes, but this code block concerns node attributres.

gboeing (changes requested):

osmnx/io.py
ncotie: Whoops!

ncotie: Fixed.  Should I try to address the code coverage issue with additional tests?

gboeing: Yes please add a test for this new functionality.

ncotie: I've modified test_osmnx.py to add tests for list, set, dict attributes for both nodes and edges, if that scope change is ok. 

A local test, running `pytest --cov=./osmnx --cov-report=term-missing --verbose` in a terminal on the local git code shows all tests are passing, so the testing code changes appear to be working:

      tests/test_osmnx.py::test_logging PASSED                                 [  7%]
      tests/test_osmnx.py::test_coords_rounding PASSED                         [ 15%]
      tests/test_osmnx.py::test_geocoder PASSED                                [ 23%]
      tests/test_osmnx.py::test_stats PASSED                                   [ 30%]
      tests/test_osmnx.py::test_osm_xml PASSED                                 [ 38%]
      tests/test_osmnx.py::test_elevation PASSED                               [ 46%]
      tests/test_osmnx.py::test_routing PASSED                                 [ 53%]
      tests/test_osmnx.py::test_plots PASSED                                   [ 61%]
      tests/test_osmnx.py::test_find_nearest PASSED                            [ 69%]
      tests/test_osmnx.py::test_endpoints PASSED                               [ 76%]
      tests/test_osmnx.py::test_graph_save_load PASSED                         [ 84%]
      tests/test_osmnx.py::test_graph_from_functions PASSED                    [ 92%]
      tests/test_osmnx.py::test_features PASSED                                [100%]

However, there was a warning immediately afterward the above: `CoverageWarning: No data was collected.`, which I'm not sure how to interpret.

Separately, format.sh and lint_test.sh were run but no apparent automatic changes to the file, so I will push it to my fork, and I suppose the test_osmnx.py file will get added to this PR automatically in addition to the io.py file ...

ncotie:
Geoff, tell me please if I should be doing anything specific with the PR now.  I'm assuming the "review" is for you to do and that there's nothing needed from me for now.  The "1 change requested" tagged just above is I think already commented in the discussion earlier, and I don't believe I need to take any action on it.  If that's not correct, please let me know.

gboeing (approved):

ncotie:
Thanks Geoff!

PR #1080, opened 2023-10-29
update docs to refer to server usage limitations

### LINKED ISSUE #1079: Document usage limitations
### Contributing guidelines

### Documentation

### Existing issues

Several users use this library without understanding usage limits of called services.

Document usage limitations and policy of underlying services and mention it clearly in quick start

Errors related to usage limits may also fail with error messages mentioning over use of services and suggesting alternatives for workloads requiring heavy use.

### Additional context

I can track down support requests appearing at OpenStreetMap forums.

Users of this library are among most common cases assisting there unaware of what is going wrong and that their scale of usage is exceeding underlying services, and that they violated their usage policies

gboeing:
resolves #1079

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,4 +31,4 @@ If you have any trouble, consult the [User Reference](https://osmnx.readthedocs.
 
 ## License
 
-OSMnx is open source and licensed under the MIT license. OpenStreetMap's open data [license](https://www.openstreetmap.org/copyright/) requires that derivative works provide proper attribution.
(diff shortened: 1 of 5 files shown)


matkoniecz:
> If you need to exceed these limitations, consider installing your own hosted instance and setting OSMnx to use it.

Maybe mention that you can also download entire OSM dataset or other major parts? See https://wiki.openstreetmap.org/wiki/Downloading_data

Issue #1098, opened 2023-12-19
Removal of inner_polygons from outer_polygons (creation of holes), creates maximum one hole.


orrsim:
### Contributing guidelines

- [X] I understand the contributing guidelines

### Documentation

- [X] My problem is not addressed by the documentation or examples

### Existing issues

- [X] My problem does not appear in an existing issue

### What operating system and Python version are you using?

Windows 11 Python 3.12

### What OSMnx version are you using?

1.8.0

### Environment packages and versions

```shell
Package            Version
------------------ ---------
affine             2.4.0
asttokens          2.4.0
attrs              22.2.0
backcall           0.2.0
basemap            1.3.8
basemap-data       1.3.2
beautifulsoup4     4.11.2
branca             0.7.0
Cartopy            0.22.0
certifi            2022.12.7
charset-normalizer 3.0.1
click              8.1.3
click-plugins      1.1.1
cligj              0.7.2
cloudpickle        2.2.1
colorama           0.4.6
comm               0.1.4
contourpy          1.1.1
cycler             0.12.0
debugpy            1.8.0
decorator          5.1.1
executing          2.0.0
Fiona              1.9.1
fonttools          4.43.0
GDAL               3.4.3
geopandas          0.14.0
html5lib           1.1
idna               3.4
ipykernel          6.25.2
ipyleaflet         0.18.0
ipython            8.16.1
ipywidgets         8.1.1
jedi               0.19.1
Jinja2             3.1.2
jupyter_client     8.3.1
jupyter_core       5.3.2
jupyterlab-widgets 3.0.9
kiwisolver         1.4.5
lxml               4.9.2
MarkupSafe         2.1.3
matplotlib         3.7.2
matplotlib-inline  0.1.6
munch              2.5.0
nest-asyncio       1.5.8
networkx           3.1
numpy              1.25.2
osmnx              1.8.0
packaging          23.0
pandas             2.0.3
parso              0.8.3
pickleshare        0.7.5
Pillow             10.0.1
pip                23.3.2
platformdirs       3.11.0
prettymapp         0.2.0
prettymaps         1.0.0
prompt-toolkit     3.0.39
psutil             5.9.5
pure-eval          0.2.2
Pygments           2.16.1
pyparsing          3.0.9
pyproj             3.4.1
PyQt5              5.15.9
PyQt5-Qt5          5.15.2
PyQt5-sip          12.12.2
pyshp              2.3.1
python-dateutil    2.8.2
[...]

PR #1104, opened 2024-01-05
fix removal of inner_polygons from outer_polygons

AtelierLibre:
This is a fix for issue #1098. As described by @orrsim, the difference in the inner loop was referring back to the original `outer_polygon` rather than the result of any previous difference `outer_polygon_diff`.

With this fix,
`
ox.features_from_point(
    (60.4436,10.2364),
    tags = {'natural' : 'water'},
    dist = 100
)
`
returns Fjorda with the islands removed.

@orrsim, thanks for pointing this out and also for identifying the bug.


gboeing:
Thank you @AtelierLibre!

Closes #1098

PR #1107, opened 2024-01-11
add type annotations

### LINKED ISSUE #1062: Adding type hints to the function signatures
### Contributing guidelines

### Documentation

### Existing issues

The types of the parameters and the returned data of the functions is already very well documented. Currently the data is in docstrings, which unfortunately IDEs like VS Code cannot use. Putting them additionally directly in the function signatures would have several advantages:
* Make it easier for users to see what they need to input without having to scroll through the docstrings
* Enable type checkers to find quickly find errors if the user uses something wrong
* Enable autocomplete on returned values

One could simply copy the info out of the docstrings and put it in the function signatures. I think using `from __future__ import annotations` would be the best choice, as we could use modern constructs from Python 3.10+ and still have a runnable program on Python 3.8.

One could also create type stubs, but they would be harder to maintain and get out of date.

### Additional context

-

gboeing:
*Comments, feedback, and improvements welcome.*

This PR adds complete type annotations throughout the package (validated via `mypy . --strict` and `typeguard` in CI) for user type hinting and type checking. It also fixes minor bugs and inconsistencies revealed by the subsequent type check validation. Resolves #1062.

This PR includes the following enhancements:
  - [x] adds type annotations to all public and private functions
  - [x] adds mypy strict validation to CI
  - [x] adds typeguard runtime validation to CI
  - [x] maintains test coverage
  - [x] fix bug where `_downloader._save_to_cache` function expected `requests.response.ok` bool values as an argument, but was being passed `requests.response.status_code` int values instead
  - [x] minor fixes throughout to address inconsistencies revealed by mypy static checking and typeguard runtime checking

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,7 @@ jobs:
         run: make -C ./docs html
 
       - name: Test code
-        run: pytest --cov=./osmnx --cov-report=xml --cov-report=term-missing --verbose
+        run: pytest --maxfail=1 --typeguard-packages=osmnx --cov=./osmnx --cov-report=term-missing --verbose
 
       - name: Upload coverage report
         uses: codecov/codecov-action@v3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: CI
 
 on:
   push:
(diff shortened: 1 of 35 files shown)


gboeing:

PR #1113, opened 2024-01-16
Remove previously deprecated functionality

gboeing:
This PR removes previously deprecated functionality for v2.0.0 to streamline the API and make it more user friendly.


gboeing:

PR #1114, opened 2024-01-20
rename private _downloader module, remove unneeded _api module, and fix unreachable code

gboeing:
See also #1106


gboeing:

PR #1117, opened 2024-01-26
add endpoint_attrs param to simplify_graph to flexibly relax strictness

### LINKED ISSUE #625: Customizable Graph Simplification

I've been working on a side project with pedestrian/sidewalk networks recently (and some bike, but mostly ped) and have found myself in situations where it would be useful to simply the network down from the unsimplified OSM network but the existing simplification function is far too strict. In particular, I would like to be able to leave nodes in places where the infrastructure in an area switches from dedicated to non-dedicated. By doing this, nodes capture all of the places where someone walking along the graph has to make a decision—not only to go left or right at an intersection, but also the decision to turn around because (oh no) the sidewalk ends or the trail doesn't have a marked crosswalk across a busy road.

I would like to be able to be able to pass a custom `_is_endpoint` function to `simplification.simplify_graph` (and `simplification._get_paths_to_simplify` as a result) as a parameter. The existing `strict` argument could be left as is or appended to a kwargs dictionary if the chosen function is the default. 

These changes might also require a new `G.graph["simplified"]` flag and `simplification._is_simplified` function, though I have not fully thought through how to reimplement these. My initial impression is that [this line](https://github.com/gboeing/osmnx/blob/95b7deb4bc325012b97908304c444cca84c39f61/osmnx/simplification.py#L319) could be modified by replacing `True` with the function name and the second condition in [this line](https://github.com/gboeing/osmnx/blob/95b7deb4bc325012b97908304c444cca84c39f61/osmnx/simplification.py#L206) could simply be removed. 

**Describe alternatives you've considered**

I do have the functionality I want as it is, but it's required me to fully duplicate the `simplify_graph` and `_get_paths_to_simplify` functions in their near entirety. My current setup looks like this: 

```python
def custom_is_endpoint(G, node):
    """
    Is node a true endpoint of an edge (considering dedicated ped infrastructure).
    Return True for same cases as osmnx.simplification._is_endpoint OR if
    node represents transition between dedicated and non-dedicated infrastructure.
    An end point is a node that either:
    1) is its own neighbor, ie, it self-loops.
    2) or, has no incoming edges or no outgoing edges, ie, all its incident edges point
    inward or all its incident edges point outward.
    3) or, it does not have exactly two neighbors and degree of 2 or 4.
    4) or, represents the transition between any of the following: sidewalks, marked or
    unmarked crosswalks, and streets without dedicated infrastructure.
    ----------
    G : networkx.MultiDiGraph
        input graph
    node : int
        the node to examine
    strict : bool
        if False, allow nodes to be end points even if they fail all other rules
        but have edges with different OSM IDs
    Returns
    -------
    bool
    """
    is_strict_endpoint = ox.simplification._is_endpoint(G, node, strict=True)

    if is_strict_endpoint:
        return True

    pedestrian_highway_types = set(["footway", "path", "cycleway", "steps"])
    marked_crossing_types = [
        "marked",
        "zebra",
        "pedestrian_signals",
        "traffic_signals",
        "traffic_lights",
    ]

    node_dedicated_adj = False
    node_marked_adj = False
    node_unmarked_adj = False
    node_undedicated_adj = False
    for neighboring_node_multi_edge in G[node].values():
        for edge_attributes in neighboring_node_multi_edge.values():
            undedicated_edge = (
                edge_attributes["highway"] not in pedestrian_highway_types
            )

            marked_edge = False
            if "crossing" in edge_attributes:
                marked_edge = edge_attributes["crossing"] in marked_crossing_types
                unmarked_edge = not marked_edge
                dedicated_edge = False
            else:
                dedicated_edge = not undedicated_edge
                marked_edge = False
                unmarked_edge = False

            node_dedicated_adj = node_dedicated_adj or dedicated_edge
            node_undedicated_adj = node_undedicated_adj or undedicated_edge
            node_marked_adj = node_marked_adj or marked_edge
            node_unmarked_adj = node_unmarked_adj or unmarked_edge

    is_transition_node = (
        node_dedicated_adj + node_marked_adj + node_unmarked_adj + node_undedicated_adj
    ) > 1

    return is_transition_node
```
```python
def custom_get_paths_to_simplify(G):
    """
    Generate all the paths to be simplified between endpoint nodes.
    The path is ordered from the first endpoint, through the interstitial nodes,
    to the second endpoint.
    Parameters
    ----------
    G : networkx.MultiDiGraph
        input graph
    Yields
    ------
    path_to_simplify : list
    """
    # first identify all the nodes that are endpoints
    endpoints = set([n for n in G.nodes() if custom_is_endpoint(G, n)])
    ox.utils.log(f"Identified {len(endpoints)} edge endpoints")

    # for each endpoint node, look at each of its successor nodes
    for endpoint in endpoints:
        for successor in G.successors(endpoint):
            if successor not in endpoints:
                # if endpoint node's successor is not an endpoint, build a path
                # from the endpoint node, through the successor, and on to the
                # next endpoint node
                yield ox.simplification._build_path(G, endpoint, successor, endpoints)
```
```python
def custom_simplify_graph(G):
    """
    Simplify a graph's topology by removing interstitial nodes.
    Simplify graph topology by removing all nodes that are not intersections,
    dead-ends, or infrastructure transitions. Create an edge directly between
    the end points that encapsulate them, but retain the geometry of the
    original edges, saved as a new `geometry` attribute on the new edge. Note
    that only simplified edges receive a `geometry` attribute. Some of the
    resulting consolidated edges may comprise multiple OSM ways, and if so,
    their multiple attribute values are stored as a list.
    Parameters
    ----------
    G : networkx.MultiDiGraph
        input graph
    Returns
    -------
    G : networkx.MultiDiGraph
        topologically simplified graph, with a new `geometry` attribute on
        each simplified edge
    """
    if ox.simplification._is_simplified(G):
        raise Exception(
            "This graph has already been simplified, cannot simplify it again."
        )

    ox.utils.log("Begin topologically simplifying the graph...")

    # make a copy to not mutate original graph object caller passed in
    G = G.copy()
    initial_node_count = len(G)
    initial_edge_count = len(G.edges)
    all_nodes_to_remove = []
    all_edges_to_add = []

    # generate each path that needs to be simplified
    for path in custom_get_paths_to_simplify(G):

        # add the interstitial edges we're removing to a list so we can retain
        # their spatial geometry
        edge_attributes = dict()
        for u, v in zip(path[:-1], path[1:]):

            # there should rarely be multiple edges between interstitial nodes
            # usually happens if OSM has duplicate ways digitized for just one
            # street... we will keep only one of the edges (see below)
            if not G.number_of_edges(u, v) == 1:
                ox.utils.log(
                    f"Found multiple edges between {u} and {v} when simplifying"
                )

            # get edge between these nodes: if multiple edges exist between
            # them (see above), we retain only one in the simplified graph
            edge = G.edges[u, v, 0]
            for key in edge:
                if key in edge_attributes:
                    # if this key already exists in the dict, append it to the
                    # value list
                    edge_attributes[key].append(edge[key])
                else:
                    # if this key doesn't already exist, set the value to a list
                    # containing the one value
                    edge_attributes[key] = [edge[key]]

        for key in edge_attributes:
            # don't touch the length attribute, we'll sum it at the end
            if len(set(edge_attributes[key])) == 1 and not key == "length":
                # if there's only 1 unique value in this attribute list,
                # consolidate it to the single value (the zero-th)
                edge_attributes[key] = edge_attributes[key][0]
            elif not key == "length":
                # otherwise, if there are multiple values, keep one of each value
                edge_attributes[key] = list(set(edge_attributes[key]))

        # construct the geometry and sum the lengths of the segments
        edge_attributes["geometry"] = LineString(
            [Point((G.nodes[node]["x"], G.nodes[node]["y"])) for node in path]
        )
        edge_attributes["length"] = sum(edge_attributes["length"])

        # add the nodes and edges to their lists for processing at the end
        all_nodes_to_remove.extend(path[1:-1])
        all_edges_to_add.append(
            {"origin": path[0], "destination": path[-1], "attr_dict": edge_attributes}
        )

    # for each edge to add in the list we assembled, create a new edge between
    # the origin and destination
    for edge in all_edges_to_add:
        G.add_edge(edge["origin"], edge["destination"], **edge["attr_dict"])

    # finally remove all the interstitial nodes between the new edges
    G.remove_nodes_from(set(all_nodes_to_remove))

    # mark graph as having been simplified
    G.graph["simplified"] = True

    msg = (
        f"Simplified graph: {initial_node_count} to {len(G)} nodes, "
        f"{initial_edge_count} to {len(G.edges)} edges"
    )
    ox.utils.log(msg)
    return G
```

This level of duplication seems obviously quite silly to me though, and I'd much rather not have to. And to be clear, I'm only proposing changes to the latter two functions—not the inclusion of my `custom_is_endpoint` function.

**Additional context**
As a quick and dirty demonstration of what this network structure allows me to do, I used the following code to plot the dedicated/undedicated pedestrian network for a Seattle area suburb:

```python
def check_tag_match(G, tag_values={}):
    for e in G.edges:
        for tag, values in tag_values.items():
            edge_data = G.edges[e]
            if tag in edge_data:
                e_tag_values = edge_data[tag]

                if type(e_tag_values) is list:
                    value_match = set(e_tag_values) & set(values)
                else:
                    value_match = e_tag_values in values

                if not value_match:
                    break

        yield value_match


dedicated_types = set(["footway", "steps", "path", "cycleway"])
marked_crossing_types = [
    "marked",
    "zebra",
    "pedestrian_signals",
    "traffic_signals",
    "traffic_lights",
]

dedicated_infrastructure = check_tag_match(
    se_bellevue_cs_G, {"highway": dedicated_types, "crossing": marked_crossing_types}
)

colors = ["tab:green" if match else "#999" for match in dedicated_infrastructure]
ns = [(se_bellevue_cs_G.in_degree[node] == 2) * 10 for node in se_bellevue_cs_G.nodes]
ox.plot_graph(se_bellevue_cs_G, node_size=ns, edge_color=colors, figsize=(20, 20))
plt.show()
```

image
Green indicates a sidewalk or marked crosswalk, grey indicates unmarked crosswalk or pedestrian legal street without sidewalks. Note numerous two way intersections where pedestrian infrastructure transitions.

gboeing:
Resolves #625

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,11 @@
 
 ## 1.9.0 (Unreleased)
 
+- add endpoint_attrs argument to simplification.simplify_graph function to flexibly relax strictness (#1117)
 - fix a bug in the features module's polygon handling (#1104)
 - update obsolete numpy random number generation (#1108)
 - update warning messages to note that deprecated code will be removed in v2.0.0 (#1111)
+- deprecate strict argument in simplification.simplify_graph function in favor of new endpoint_attrs argument (#1117)
 - deprecate north, south, east, west arguments throughout package in favor of bbox tuple argument (#1112)
 - deprecate return_coords argument in graph.graph_from_address function (#1105)
 - deprecate return_hex argument in plot.get_colors function (#1109)
--- a/osmnx/simplification.py
+++ b/osmnx/simplification.py
@@ -15,18 +15,25 @@ from . import utils_graph
 from ._errors import GraphSimplificationError
 
 
-def _is_endpoint(G, node, strict=True):
+def _is_endpoint(G, node, endpoint_attrs):
     """
     Determine if a node is a true endpoint of an edge.
 
-    Return True if the node is a "real" endpoint of an edge in the network,
-    otherwise False. OSM data includes lots of nodes that exist only as points
-    to help streets bend around curves. An end point is a node that either:
-    1) is its own neighbor, ie, it self-loops.
-    2) or, has no incoming edges or no outgoing edges, ie, all its incident
-    edges point inward or all its incident edges point outward.
-    3) or, it does not have exactly two neighbors and degree of 2 or 4.
-    4) or, if strict mode is false, if its edges have different OSM IDs.
+    Return True if the node is a "true" endpoint of an edge in the network,
+    otherwise False. OpenStreetMap data includes many nodes that exist only as
+    geometric vertices to allow ways to curve. A true edge endpoint is a node
+    that satisfies at least 1 of the following 4 rules:
+
+    1) It is its own neighbor (ie, it self-loops).
+
+    2) Or, it has no incoming edges or no outgoing edges (ie, all its incident
+    edges are inbound or all its incident edges are outbound).
+
+    3) Or, it does not have exactly two neighbors and degree of 2 or 4.
+
+    4) Or, if `endpoint_attrs` is not None, and its incident edges have
+    different values than each other for any of the edge attributes in
+    `endpoint_attrs`.
 
     Parameters
     ----------
@@ -34,9 +41,11 @@ def _is_endpoint(G, node, strict=True):
         input graph
     node : int
         the node to examine
-    strict : bool
-        if False, allow nodes to be end points even if they fail all other rules
-        but have edges with different OSM IDs
+    endpoint_attrs : iterable
+        An iterable of edge attribute names for relaxing the strictness of
+        endpoint determination. If not None, a node is an endpoint if its
+        incident edges have different values then each other for any of the
+        edge attributes in `endpoint_attrs`.
 
     Returns
     -------
@@ -46,36 +55,37 @@ def _is_endpoint(G, node, strict=True):
     n = len(neighbors)
     d = G.degree(node)
 
-    # rule 1
+    # RULE 1
+    # if the node appears in its list of neighbors, it self-loops: this is
+    # always an endpoint
     if node in neighbors:
-        # if the node appears in its list of neighbors, it self-loops
-        # this is always an endpoint.
         return True
 
-    # rule 2
+    # RULE 2
+    # if node has no incoming edges or no outgoing edges, it is an endpoint
     if G.out_degree(node) == 0 or G.in_degree(node) == 0:
-        # if node has no incoming edges or no outgoing edges, it is an endpoint
         return True
 
-    # rule 3
+    # RULE 3
+    # else, if it does NOT have 2 neighbors AND either 2 or 4 directed edges,
+    # it is an endpoint. either it has 1 or 3+ neighbors, in which case it is
+    # a dead-end or an intersection of multiple streets or it has 2 neighbors
+    # but 3 degree (indicating a change from oneway to twoway) or more than 4
+    # degree (indicating a parallel edge) and thus is an endpoint
     if not ((n == 2) and (d in {2, 4})):  # noqa: PLR2004
-        # else, if it does NOT have 2 neighbors AND either 2 or 4 directed
-        # edges, it is an endpoint. either it has 1 or 3+ neighbors, in which
-        # case it is a dead-end or an intersection of multiple streets or it has
-        # 2 neighbors but 3 degree (indicating a change from oneway to twoway)
-        # or more than 4 degree (indicating a parallel edge) and thus is an
-        # endpoint
         return True
 
-    # rule 4
-    if not strict:
-        # non-strict mode: do its incident edges have different OSM IDs?
-        # first collect all the OSM way IDs for incoming edges
-        # then collect all the OSM way IDs for outgoing edges
-        # if there is more than 1 OSM ID then it is an endpoint, otherwise not
-        incoming = [G.edges[u, node, k]["osmid"] for u in G.predecessors(node) for k in G[u][node]]
-        outgoing = [G.edges[node, v, k]["osmid"] for v in G.successors(node) for k in G[node][v]]
-        return len(set(incoming + outgoing)) > 1
+    # RULE 4
+    # non-strict mode: do its incident edges have different attr values? for
+    # each attribute to check, collect the attribute's values in all inbound
+    # and outbound edges. if there is more than 1 unique value then then this
+    # node is an endpoint
+    if endpoint_attrs is not None:
+        for attr in endpoint_attrs:
+            in_values = {v for _, _, v in G.in_edges(node, data=attr, keys=False)}
+            out_values = {v for _, _, v in G.out_edges(node, data=attr, keys=False)}
+            if len(in_values | out_values) > 1:
+                return True
 
     # if none of the preceding rules passed, then it is not an endpoint
     return False
@@ -151,7 +161,7 @@ def _build_path(G, endpoint, endpoint_successor, endpoints):
     return path
 
 
-def _get_paths_to_simplify(G, strict=True):
+def _get_paths_to_simplify(G, endpoint_attrs):
     """
     Generate all the paths to be simplified between endpoint nodes.
 
@@ -162,9 +172,11 @@ def _get_paths_to_simplify(G, strict=True):
     ----------
     G : networkx.MultiDiGraph
         input graph
-    strict : bool
-        if False, allow nodes to be end points even if they fail all other rules
-        but have edges with different OSM IDs
+    endpoint_attrs : iterable
+        An iterable of edge attribute names for relaxing the strictness of
+        endpoint determination. If not None, a node is an endpoint if its
+        incident edges have different values then each other for any of the
+        edge attributes in `endpoint_attrs`.
 
     Yields
     ------
@@ -172,7 +184,7 @@ def _get_paths_to_simplify(G, strict=True):
         a generator of paths to simplify
     """
     # first identify all the nodes that are endpoints
-    endpoints = {n for n in G.nodes if _is_endpoint(G, n, strict=strict)}
+    endpoints = {n for n in G.nodes if _is_endpoint(G, n, endpoint_attrs)}
     utils.log(f"Identified {len(endpoints):,} edge endpoints")
 
     # for each endpoint node, look at each of its successor nodes
@@ -185,7 +197,7 @@ def _get_paths_to_simplify(G, strict=True):
                 yield _build_path(G, endpoint, successor, endpoints)
 
 
-def _remove_rings(G):
+def _remove_rings(G, endpoint_attrs):
     """
     Remove all self-contained rings from a graph.
 
@@ -196,6 +208,11 @@ def _remove_rings(G):
     ----------
     G : networkx.MultiDiGraph
         input graph
+    endpoint_attrs : iterable
+        An iterable of edge attribute names for relaxing the strictness of
+        endpoint determination. If not None, a node is an endpoint if its
+        incident edges have different values then each other for any of the
+        edge attributes in `endpoint_attrs`.
 
     Returns
     -------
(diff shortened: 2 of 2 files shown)


gboeing:
@csebastiao would you like to test this out with a specific case?

csebastiao:
@gboeing I have tested it on my side on the two examples of the [Toy graph](https://github.com/gboeing/osmnx/issues/625#issuecomment-1893483141) and [Copenhagen custom bicycle attribute](https://github.com/gboeing/osmnx/issues/625#issuecomment-1895836512) and both are returning the expected behaviors.

It's all good for me.

gboeing:
Thanks. Can you provide a code snippet for testing your use case?

csebastiao:
Oh sure sorry about that, here is the one for the Toy Graph made by hand:

```python
import networkx as nx
import shapely

import osmnx as ox

G = nx.Graph()
G.add_node(1, x=1, y=1)
G.add_node(2, x=2, y=1)
G.add_node(3, x=2.5, y=1.5)
G.add_node(4, x=3, y=2.5)
G.add_node(5, x=3.5, y=3.5)
G.add_node(6, x=3, y=4)
G.add_node(7, x=3, y=5)
G.add_node(8, x=3.5, y=5.5)
G.add_node(9, x=3.5, y=6.5)
G.add_node(10, x=4, y=7)
G.add_node(11, x=5, y=8)
G.add_node(12, x=6, y=8)
G.add_node(13, x=6.5, y=8.5)
G.add_node(14, x=7, y=9)
G.add_node(15, x=7.5, y=8.5)
G.add_node(16, x=8, y=8)
G.add_node(17, x=7.5, y=7.5)
G.add_node(18, x=7, y=7)
G.add_node(19, x=6.5, y=7.5)
# add length and osmid just for the osmnx function to work
for i in range(1, 19):
    G.add_edge(i, i + 1, length=1, osmid=i)
G.add_node(20, x=4, y=4)
G.add_node(21, x=4, y=5)
G.add_edge(5, 20, length=1, osmid=20)
G.add_edge(20, 21, length=1, osmid=21)
G.add_edge(21, 8, length=1, osmid=22)
G.add_edge(19, 12, length=1, osmid=23)
# give three value of color to see the discrimination for an attribute
for i in range(2, 8):
    G.edges[i, i + 1]["color"] = 1
G.edges[1, 2]["color"] = 2
G.edges[5, 20]["color"] = 1
G.edges[20, 21]["color"] = 1
G.edges[21, 8]["color"] = 1
for i in range(8, 11):
    G.edges[i, i + 1]["color"] = 3
for i in range(11, 19):
    G.edges[i, i + 1]["color"] = 2
G.edges[19, 12]["color"] = 2
G = nx.MultiDiGraph(G)
# add crs for the ox_plot_graph to work
G.graph["crs"] = "epsg:4326"
ec = ox.plot.get_edge_colors_by_attr(G, "color", cmap="Set1")
ox.plot_graph(
    G,
    figsize=(12, 8),
    bgcolor="w",
    node_color="black",
    node_size=30,
    edge_color=ec,
    edge_linewidth=3,
)
G_simple = ox.simplify_graph(G, endpoint_attrs=["color"])
ec_s = ox.plot.get_edge_colors_by_attr(G_simple, "color", cmap="Set1")
ox.plot_graph(
    G_simple,
    figsize=(12, 8),
    bgcolor="w",
    node_color="black",
    node_size=30,
    edge_color=ec_s,
    edge_linewidth=3,
)
```

Which can be asserted with 
```python
assert len(G_simple) == len(ox.simplify_graph(G)) + 2
```

One way to see it on an actual OSMnx graph is with this basic example:
```python
import osmnx as ox

protected_dict = {}
protected_dict["sidewalk:left:bicycle"] = "yes"
protected_dict["sidewalk:left:right"] = "yes"
protected_dict["cycleway:left"] = ["shared_lane", "shared_busway", "track"]
protected_dict["cycleway:right"] = ["shared_lane", "shared_busway", "track"]
protected_dict["cycleway:both"] = "lane"
protected_dict["cycleway"] = ["shared_lane", "shared_busway", "opposite_lane", "opposite"]
protected_dict["bicycle"] = ["designated", "yes", "official", "use_sidepath"]
protected_dict["highway"] = ["cycleway", "bridleway"]
protected_dict["cyclestreet"] = "yes"
protected_dict["bicycle_road"] = "yes"
for val in protected_dict:
    if val not in ox.settings.useful_tags_way:
        ox.settings.useful_tags_way += [val]
G = ox.graph_from_place("Frederiksberg Municipality, Denmark", simplify=False)
for edge in G.edges:
    for key in list(protected_dict.keys()):
        if key in list(G.edges[edge].keys()):
            if G.edges[edge][key] in protected_dict[key]:
                G.edges[edge]["cycling"] = 1
                break
            else:
                G.edges[edge]["cycling"] = 0
        else:
            G.edges[edge]["cycling"] = 0
G_s = ox.simplify_graph(G, endpoint_attrs=None)
G_ns = ox.simplify_graph(G, endpoint_attrs=["osmid"])
G_attr = ox.simplify_graph(G, endpoint_attrs=["cycling"])
ec = ox.plot.get_edge_colors_by_attr(G_attr, "cycling", cmap="RdYlGn")
ox.plot_graph(
    G_attr,
    figsize=(12, 8),
    bgcolor="w",
    node_color="black",
    node_size=10,
    edge_color=ec,
    edge_linewidth=1,
)
```

To take a specific node (even thought with the ever-updating OSM I don't know if the exact values will stay) one can look at the node `1262273553`:
```python
G_attr.edges(1262273553, data=True)
```

It only exists when discriminating with the cycling (or osmid) attribute:
```python
assert 1262273553 not in G_s
assert 1262273553 in G_attr
```

This should be the case as there are only two edges `(1262273553, 9740065276, 0)` and `(1262273553, 9917054947, 0)` with a different value for the cycling attribute, because for the latter we have `'highway': 'cycleway'` and `'bicycle': 'designated'` and for the former `'highway': 'footway'`.

EwoutH:
`endpoint_attrs` looks very useful, thanks for working on this!

gboeing:
I believe this is ready to merge.

CHANGELOG.md
```diff
@@ -2,9 +2,11 @@
 
 ## 1.9.0 (Unreleased)
 
+- add endpoint_attrs argument to simplification.simplify_graph function to flexibly relax strictness (#1117)
 - fix a bug in the features module's polygon handling (#1104)
 - update obsolete numpy random number generation (#1108)
 - update warning messages to note that deprecated code will be removed in v2.0.0 (#1111)
+- deprecate strict argument in simplification.simplify_graph function in favor of new endpoint_attrs argument (#1117)
```
EwoutH, line 9: It's not deprecated, but removed altogether, right?

osmnx/simplification.py
```diff
@@ -245,6 +272,17 @@ def simplify_graph(G, strict=True, remove_rings=True, track_merged=False):
         topologically simplified graph, with a new `geometry` attribute on
         each simplified edge
     """
+    if strict is not None:
+        msg = (
+            "The `strict` parameter has been deprecated and will be removed in "
+            "the v2.0.0 release. Use the `endpoint_attrs` parameter instead to "
+            "relax simplification strictness. For example, `endpoint_attrs=None` "
+            "reproduces the old `strict=True` behvavior and `endpoint_attrs=['osmid']` "
+            "reproduces the old `strict=False` behavior."
```
EwoutH, line 281: Maybe add this in the changelog, a migration guide and/or the PR message.

EwoutH (commented):
Few minor comments. Really like the explicit numbered rules in the comments, helps a lot with understanding the code!

CHANGELOG.md
gboeing: Deprecated. Will be removed in the v2.0.0 release.

osmnx/simplification.py
gboeing: Good idea. Added to PR comment.

gboeing:
Note that #1145 additionally adds a `node_attrs_include` param to the `simplify_graph` function to further flexibly relax graph simplification strictness. It also renames the `endpoint_attrs` param to `edge_attrs_differ` for consistent and clear naming, given the new param.

Issue #1123, opened 2024-01-31
OSMnx 2.0 Migration Guide


gboeing:
# Overview

OSMnx 2.0.0 was released in November 2024. This major release includes some breaking changes (including removing previously deprecated functionality) that are not backwards compatible with v1. This issue provides guidance on migrating from v1 to v2.

# Migrating to v2

If you want to ensure your code is fully compatible with OSMnx v2 before upgrading to it, you can install the latest v1 release (currently `1.9.4`), then run your code. The latest v1 release will remain compatible with the v1 API, but any deprecated functionality you're using will issue a FutureWarning with simple guidance to make it v2 compatible. Additionally, you can:

- refer to the new official OSMnx [reference paper](https://geoffboeing.com/publications/osmnx-paper/) up-to-date for v2
- check out the [Examples Gallery](https://github.com/gboeing/osmnx-examples), which has been updated for v2 compatibility if you're looking for sample code
- see the [changelog](https://github.com/gboeing/osmnx/blob/main/CHANGELOG.md) for details

# Key changes

The three biggest overarching changes in 2.0 are:

1. The package is now fully type-annotated throughout.
2. Several modules have been refactored or wholly rewritten for performance optimization.
3. Flexible and customizable graph simplification and node consolidation.
4. The package's API has been streamlined and made more consistent, including some breaking changes and some moved/renamed functions.

In addition to these, there have been many smaller enhancements, fixes, and changes. See the [changelog](https://github.com/gboeing/osmnx/blob/main/CHANGELOG.md) for details and see the latest [documentation](https://osmnx.readthedocs.io/en/latest/) for an up-to-date user reference.

gboeing:
The first OSMnx v2.0.0 pre-release (beta0) has been released. Testers needed!

EwoutH:
Briefly tested beta0 on a few of my existing notebooks. Not major issues so far!

jGaboardi:
I'm also seeing no major/unexpected issues thus far.

csebastiao:
No issues observed on a couple of my typical use cases !

Noting there in case the FutureWarning coming from Geopandas in the [1.0.0 latest version](https://github.com/geopandas/geopandas/issues/3201) when calling `geopandas.GeoSeries.unary_union()`, in `osmnx.graph.graph_from_place()`, or `osmnx.plot.plot_figure_ground()` for instance.

gboeing:
[...]

PR #1125, opened 2024-02-01
Improve logging and warnings

gboeing:
This PR:
  - [x] improves log messaging and makes level explicit
  - [x] makes user warnings explicitly UserWarnings
  - [x] ensure imports used only for type hinting are guarded in `if TYPE_CHECKING:` block and ignore these blocks in test coverage report ([relevant](https://github.com/nedbat/coveragepy/issues/831))

See also #1106


gboeing:

Issue #970, opened 2023-05-30
save_graph_xml fails if roundabout present


stl-maxgardner:
**Read these instructions carefully. Use the template below and fill out every section. Your issue will be automatically closed if you don't provide all of the information requested in the template, because we need this information to assist you. Before you proceed, review the contributing guidelines in the CONTRIBUTING.md file.**

Please don't use the issue tracker to ask "how-to" or usage questions. If asked here, they will be closed automatically. Instead, ask "how-to" and usage questions on [StackOverflow](https://stackoverflow.com/search?q=osmnx). If you installed OSMnx via conda and are experiencing installation problems, please open an issue at its [feedstock](https://github.com/conda-forge/osmnx-feedstock/issues) instead.

Bug reports are for reporting a bug you have found in OSMnx's codebase. First search the open/closed issues and StackOverflow to see if the problem has already been noted. If it hasn't, fill in the bug reporting template below.

**Completely fill out the following template.**

  - [x] I have read the OSMnx [contributing guidelines](../blob/main/CONTRIBUTING.md)
  - [x] I have read the OSMnx [documentation](https://osmnx.readthedocs.io/) and [usage examples](https://github.com/gboeing/osmnx-examples)
  - [x] I have searched the open/closed issues and [StackOverflow](https://stackoverflow.com/search?q=osmnx) for problems similar to mine

**Problem description**
  - What did you do? Tried to create an osm xml file from a region containing a roundabout
  - What did you expect to happen? An .osm xml file created
  - And what actually happened instead? `networkx.NetworkXUnfeasible` error was thrown as soon as osmnx tried to perform a topological sort on a table of edges corresponding to a roundabout (https://overpass-turbo.eu/s/1vwy).

**Environment information**
  - What operating system are you using? Ubuntu
  - What Python version are you using? 3.11
  - What OSMnx version are you using? 1.3.1
  - How did you install OSMnx? conda-forge
  - Provide a complete list of your environment's packages and their versions (for example, run `conda list` or `pip list` then paste the output between the two "details" tags below)

<details>
osmnx=1.3.1
numpy=1.24.3
</details>

**Provide a complete minimal reproducible example**

[...]

PR #1135, opened 2024-02-21
Improve OSM XML file saving

gboeing:
Targeting the forthcoming OSMnx v2.0.0 release (see also #1106), this PR wholly refactors the `_osm_xml` module and fixes several bugs, significantly speeds up the `save_graph_xml` function, and streamlines the  function's signature.

The PR does the following:

- [x] Speed up the `save_graph_xml` function runtime by a factor of 5x or so, depending on graph size.
- [x] Fix a bug where edge merging [grouped](https://github.com/gboeing/osmnx/blob/v1.9.1/osmnx/osm_xml.py#L471) edges into ways on the "id" column, but this column [was not](https://github.com/gboeing/osmnx/blob/v1.9.1/osmnx/osm_xml.py#L261-L266) the way OSM ID, so each group just contained a single edge.
- [x] Fix a bug where the old topological sort only worked when the entire way was a single cycle (such as is the case with a roundabout way), but could not sort other ways merely containing a cycle (or multiple cycles, such as [here](https://www.openstreetmap.org/way/13366963) and [here](https://www.openstreetmap.org/way/13339816)) within them. See also #970.
- [x] Fix a bug by adding the missing "bounds" sub-element to the XML file and adding a "visible" attribute to each node and way sub-element (see [this example](https://wiki.openstreetmap.org/wiki/OSM_XML#Example_OSM_XML_file)).
- [x] Remove the `node_tags`, `node_attrs`, `edge_tags`, and `edge_attrs` parameters from the `save_graph_xml` function in favor of using the `settings` module directly.
- [x] Remove the `oneway`, `api_version`, and `precision` parameters from the `save_graph_xml` function in favor of hard-coded sensible defaults.  See also #917 and #961
- [x] Remove the `merge_edges` parameter from the `save_graph_xml` function and instead always operate as though this were True.
- [x] Add an `encoding` parameter to the `save_graph_xml` function to configure file encoding and make it consistent with the other file saving functions
- [x] Change the `save_graph_xml` function's `data` parameter to `G` and its argument type to `nx.MultiDiGraph` instead of `nx.MultiDiGraph | tuple[gpd.GeoDataFrame, gpd.GeoDataFrame` (that is, the function no longer accepts a tuple of GeoDataFrames: you can only pass a MultiDiGraph).
- [x] Change the `save_graph_xml` function's `edge_tag_aggs` parameter to `way_tag_aggs` and its argument type to `dict` instead of `tuple`.
- [x] Require that the graph be unsimplified when saving as OSM XML, because ways have a one-to-many cardinality with edges but graph simplification makes the cardinality many-to-many instead, which breaks the grouping of edges into ways.
- [x] Remove the settings module's `osm_xml_node_attrs`, `osm_xml_node_tags`, `osm_xml_way_attrs`, and `osm_xml_way_tags` settings because they are either unnecessary or redundant (see https://github.com/gboeing/osmnx/pull/1135#issuecomment-1960105020)
- [x] Create an XSD file to validate the XML output against (see [schema](../blob/xml/tests/input_data/osm_schema.xsd)).

Deprecate these now-removed function params and settings (with FutureWarning) for v1 in #1138.


gboeing:
As a tangential aside, it would be nice someday to replace all the OSMnx XML read/write functionality with PBF read/write functionality instead and outsource it to a dedicated third-party library. See also https://github.com/HTenkanen/pyrosm/issues/228.

stl-maxgardner:
@gboeing I think its safe (and advisable) to hardcode `oneway=False`, and to operate as if `merge_edges=True`. Both settings were parameterized to maximize flexibility but they're probably not necessary. You might consider just defining a `ONEWAY = False` constant at the top of the module and using that within `_save_graph_xml()`  rather than hardcoding the value into the method, just to make the assumption more explicit.

gboeing:
Thanks @stl-maxgardner . Did this change above also make sense:

> Require that the graph be unsimplified when saving with merge_edges=True, because ways have a one-to-many cardinality with edges but graph simplification makes the cardinality many-to-many instead, which breaks the grouping of edges into ways.

If we make the function always operate as though `merge_edges=True`, users moving forward will be unable to save a simplified graph as XML. But I believe this would be appropriate given the nature of the data.

stl-maxgardner:
Yes I think so!

gboeing:
@stl-maxgardner one more question for you.

In 54170cd I proposed that we remove the `settings` module's `osm_xml_node_attrs`, `osm_xml_node_tags`, `osm_xml_way_attrs`, and `osm_xml_way_tags` settings. The reasoning is as follows...

The `osm_xml_node_attrs` and `osm_xml_way_attrs` settings are unnecessary because the  node and way XML elements should have a consistent set of attributes as defined in the OSM XML spec. That is, the same attributes are needed every time so this shouldn't be configurable. These attributes are "changeset", "timestamp", "uid", "user", "version", "visible" and "id" for both nodes/ways, plus "lat" and "lon" for nodes.

The `osm_xml_node_tags` and `osm_xml_way_tags` settings are unnecessary because they are redundant with the `settings` modules `useful_tags_node` and `useful_tags_way` settings. That is, the latter are used to retain desired tags from the raw Overpass data and lets the user specify what data the graph nodes and edges should have. OSMnx should just save to XML the graph that the user has created, rather than offering another set of settings for changing what gets saved. If the user wants to only save a subset of these tags, they can just change the `useful_tags_node` and `useful_tags_way` settings, or just edit their graph with 1-2 lines of code prior to calling the `save_graph_xml` function.

Does that all make sense and seem reasonable to you?

MikeSoft (approved):

gboeing:
Some timings comparing this `xml` branch to the `main` branch with a small graph and a large graph... Given a code snippet like:

```python
import osmnx as ox
ox.settings.all_oneway = True
G = ox.graph.graph_from_place(place, network_type="drive", simplify=False)

%%timeit
ox.io.save_graph_xml(G, filepath="./test.xml")
```

The `%%timeit` magic gives the following timings:

```
# place = "Piedmont, CA, USA"
# main branch: 1100 ms ± 4.94 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# xml branch:   167 ms ± 411 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
# ~6.6x speed-up

# place = "Los Angeles, CA, USA"
# main branch: 113.0 s ± 790 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# xml branch:   26.2 s ± 82.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# ~4.3x speed-up
```

gboeing:
An XSD schema file was added in 06edb3d and successfully validates a variety of test OSM XML files saved by OSMnx, exported from Overpass turbo, and exported from OpenStreetMap directly.

gboeing:
Added quality-of-life improvements, warning, and additional tests for saving projected/consolidated graphs as OSM XML in ca9744b.

gboeing:

Issue #1137, opened 2024-02-28
Support directed bearing/orientation distributions and plots


dhimmel:
### What problem does your feature proposal solve?

Creating orientation/bearing distributions and plots for directed graphs.

### What is your proposed solution?

`_extract_edge_bearings` can support directed graphs and in the case of a directed graph skip adding reverse bearings (`bearings_r`):

https://github.com/gboeing/osmnx/blob/78af8ed55f1dfd2ebc079eded928be4b6af537c1/osmnx/bearing.py#L148-L191

### What alternatives have you considered?

Implementing bearing calculation, distributions, and plotting outside of osmnx and networkx, but would like to avoid such duplication.

### Additional context

Possible use cases are one-way transportation networks or [downhill ski trails](https://github.com/russellporter/openskimap.org/issues/82).

gboeing:
This would be easy enough to implement but I want to make sure that it's theoretically sound first. I think it may be, though users may need a warning to ensure things proceed according to expectations. That is, *should* we allow the calculation of bearings on directed graphs?

If a user is interested in the bearings/orientation of *flows*, then maybe yes. If the user is interested in the bearings/orientation of *physical infrastructure*, then no. Perhaps a UserWarning as such is sufficient if a user passes a MultiDiGraph instead of a MultiGraph.

dhimmel:
Thanks @gboeing for your insights here.

> _should_ we allow the calculation of bearings on directed graphs? If a user is interested in the bearings/orientation of _flows_, then maybe yes. If the user is interested in the bearings/orientation of _physical infrastructure_, then no.

That is a helpful conceptual distinction, although there is certainly a relation between physical infrastructure and flow. Some examples are waterways, separated highways, one ways trails (common in mountain biking, skiing, etc). In all these cases, the physical infrastructure is designed to support directional flow.

> Perhaps a UserWarning as such is sufficient if a user passes a MultiDiGraph instead of a MultiGraph.

Another option would be to keep the "graph must be undirected" error unless the user sets an additional argument like `directed_bearings=True` or `omit_reverse=True`.

The warning proposal also sounds good paired with documentation like: "Supplying a `MultiDiGraph` results in directional bearings. For bidirectional bearings, use `MultiDiGraph.to_undirected()`".

[...]

PR #1139, opened 2024-03-03
plot_orientation: support directed graph bearings

dhimmel:
closes https://github.com/gboeing/osmnx/issues/1137
for v2 / https://github.com/gboeing/osmnx/pull/1106

TODOs:

- [x] update changelog
- [x] add tests / code example
- [x] consider directed bearing support in `orientation_entropy`
- [x] ensure tests passing on CI (workflow approval required)


osmnx/plot.py
```diff
@@ -664,7 +664,7 @@ def plot_footprints(  # noqa: PLR0913
 
 
 def plot_orientation(  # noqa: PLR0913
-    Gu: nx.MultiGraph,
+    Gu: nx.MultiGraph | nx.MultiDiGraph,
```
dhimmel, line 667: @gboeing do we want to keep the argument name as `Gu` (assuming that u refers to undirected)? Applies here and other bearing functions?

gboeing: > assuming that u refers to undirected

Yes the convention we try to use consistently is that `Gu` means something like $G_u$: the undirected model of the network. Since this can now be either directed or undirected, you might as well rename the variable `G` for clarity.

osmnx/bearing.py
```diff
@@ -122,27 +123,30 @@ def add_edge_bearings(G: nx.MultiDiGraph) -> nx.MultiDiGraph:
 
 
 def orientation_entropy(
-    Gu: nx.MultiGraph,
+    Gu: nx.MultiGraph | nx.MultiDiGraph,
```
gboeing, line 126: This change is unnecessary for type annotating: the `MultiDiGraph` class inherits from `MultiGraph`.

```python
import networkx as nx
G = nx.MultiDiGraph()
isinstance(G, nx.MultiGraph) # True
```

osmnx/bearing.py
```diff
@@ -211,15 +217,25 @@ def _extract_edge_bearings(
...
-    # drop any nulls, calculate reverse bearings, concatenate and return
+    # drop any nulls
     bearings_array = np.array(bearings)
     bearings_array = bearings_array[~np.isnan(bearings_array)]
+    if nx.is_directed(Gu):
+        # https://github.com/gboeing/osmnx/issues/1137
+        msg = (
+            "Extracting directional bearings (one bearing per edge) due to MultiDiGraph input. "
+            "To extract bidirectional bearings (two bearings per edge, including the reverse bearing), "
+            "supply an undirected graph instead via `Gu.to_undirected()`."
```
gboeing, line 228: Best to recommend the `osmnx.utils_graph.get_undirected` function instead for its better handling of parallel edges.

osmnx/bearing.py
```diff
@@ -211,15 +217,25 @@ def _extract_edge_bearings(
...
+        warn(msg, category=UserWarning, stacklevel=2)
+        return bearings_array
+    # for undirected graphs, add reverse bearings and return
     bearings_array_r = (bearings_array - 180) % 360
     return np.concatenate([bearings_array, bearings_array_r])
 
 
 def _bearings_distribution(
-    Gu: nx.MultiGraph,
+    Gu: nx.MultiGraph | nx.MultiDiGraph,
```
gboeing, line 238: Same type annotation comment as before.

osmnx/plot.py
gboeing: Also see prior comment on type annotation.

tests/test_osmnx.py
```diff
@@ -29,6 +29,7 @@
 from typeguard import TypeCheckError
 
 import osmnx as ox
+from osmnx.bearing import _extract_edge_bearings
```
gboeing, line 32: Remove this import and access this function via `ox.bearing._extract_edge_bearings` below when you need it.

tests/test_osmnx.py
```diff
@@ -141,6 +142,23 @@ def test_stats() -> None:
     G_clean = ox.consolidate_intersections(G, rebuild_graph=False)
 
 
+def test_extract_edge_bearings_directionality() -> None:
```
gboeing, line 145: These unit tests can be folded into the `test_bearings` function.

gboeing (changes requested):
Thanks, looking good. Just a few comments.

osmnx/bearing.py
dhimmel: Nice catch. FYI the `nx.MultiGraph | nx.MultiDiGraph` annotation is used in other places in case we want to update those separately.

osmnx/bearing.py
dhimmel: I updated to mention `osmnx.get_undirected` since I see the function is exported at root level.

tests/test_osmnx.py
dhimmel: I'm not seeing a `test_bearings` test. I do see `test_stats` but that test is already doing a lot.

dhimmel:
@gboeing looks like I need another workflow approval. If the test passes then this PR is good on my end absent any additional suggestions.

dhimmel:
hoping 9cb9c99c7bea8e8913e24881c90f43d53c9a723a will fix test.

unrelated to PR, when `min_length=0`, it might make sense to skip the `data["length"]` such that setting length is not required:

https://github.com/gboeing/osmnx/blob/9cb9c99c7bea8e8913e24881c90f43d53c9a723a/osmnx/bearing.py#L212

osmnx/bearing.py
gboeing: Thanks for flagging that. I'll review the other instances in the codebase separately so we can keep this PR clean on point.

tests/test_osmnx.py
gboeing: I thought I had already named one that but apparently it was just in my mental to-do list and I forgot :) Nevermind!

gboeing (approved):

osmnx/bearing.py
gboeing: FYI I ended up restoring your `nx.MultiGraph | nx.MultiDiGraph` type annotations because I think they are clearer for the user even if they're redundant. That seems more important.

gboeing:

Issue #1143, opened 2024-03-12
_bearings_distribution: apply weight during histogram


dhimmel:
Here is how `_bearings_distribution` currently applies the `weight` parameter:

https://github.com/gboeing/osmnx/blob/ea397d3aec77c7fb394712d5af6d490e0b159d4c/osmnx/bearing.py#L276-L277

`_extract_edge_bearings` implements the weight by repeating bearings:

https://github.com/gboeing/osmnx/blob/ea397d3aec77c7fb394712d5af6d490e0b159d4c/osmnx/bearing.py#L214-L220

Instead, the weight could be applied by [`numpy.histogram`](https://numpy.org/doc/stable/reference/generated/numpy.histogram.html). The benefits would be:

1. avoiding a loss of precision when converting a float weight value to an int
2. greater computational efficiency and reduced memory usage

The resulting counts could have decimals (non-integers), but `_bearings_distribution` is already typed to return `npt.NDArray[np.float64]` for `bin_counts`.

gboeing:
Thanks. Sounds like a promising efficiency improvement. Would you like to open a PR for me to review?

gboeing:
closed by #1147

PR #1147, opened 2024-03-13
_bearings_distribution: defer weighting to np.histogram

dhimmel:
closes https://github.com/gboeing/osmnx/issues/1143

TODO:

- [x] pass CI


osmnx/bearing.py
```diff
@@ -195,44 +195,46 @@ def _extract_edge_bearings(
...
-                # weight edges' bearings by some edge attribute value
-                bearings.extend([data["bearing"]] * int(data[weight]))
-            else:
-                # don't weight bearings, just take one value per edge
-                bearings.append(data["bearing"])
+        if u == v or data["length"] < min_length:
+            # ignore self-loops and any edges below min_length
+            continue
+        bearings.append(data["bearing"])
+        weights.append(data[weight] if weight is not None else 1.0)
```
gboeing, line 218: It would be simpler to make this `if` statement the inverse, so you can avoid the `continue` and move lines 217 and 218 under the `if` block.

gboeing (changes requested):
Thanks @dhimmel looks good. Just one comment to simplify a bit of the code.

osmnx/bearing.py
dhimmel: done in https://github.com/gboeing/osmnx/pull/1147/commits/6ca535266967a53abdbf64669f455281f4ef3afb

gboeing (approved):

PR #1149, opened 2024-03-16
_bearings_distribution: bin_centers terminology

dhimmel:
This PR addresses terminology in internal-use functions for greater clarity:

- bin centers were referred to as "edges"
- `bin_counts` and `bin_centers` now have the same length, which is best for most use cases

### DIFF
--- a/osmnx/bearing.py
+++ b/osmnx/bearing.py
@@ -268,8 +268,9 @@ def _bearings_distribution(
 
     Returns
     -------
-    bin_counts, bin_edges
-        Counts of bearings per bin and the bins edges.
+    bin_counts, bin_centers
+        Counts of bearings per bin and the bins' centers in degrees.
+        Both arrays are of length `num_bins`.
     """
     n = num_bins * 2
     bins = np.arange(n + 1) * 360 / n
@@ -282,6 +283,6 @@ def _bearings_distribution(
     count = np.roll(count, 1)
     bin_counts = count[::2] + count[1::2]
 
-    # because we merged the bins, their edges are now only every other one
-    bin_edges = bin_edges[range(0, len(bin_edges), 2)]
-    return bin_counts, bin_edges
+    # because we merged the bins, their centers are now only every other one
+    bin_centers = bin_edges[range(0, len(bin_edges) - 1, 2)]
+    return bin_counts, bin_centers
--- a/osmnx/bearing.py
+++ b/osmnx/bearing.py
@@ -272,17 +272,24 @@ def _bearings_distribution(
         Counts of bearings per bin and the bins' centers in degrees.
         Both arrays are of length `num_bins`.
     """
-    n = num_bins * 2
-    bins = np.arange(n + 1) * 360 / n
+    # Split bins in half to prevent bin-edge effects around common values.
+    # Bins will be merged in pairs after the histogram is computed.
+    # The last bin edge is the same as the first (i.e., 0 degrees = 360 degrees).
+    num_split_bins = num_bins * 2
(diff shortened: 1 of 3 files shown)


gboeing:
@dhimmel help me understand what's going on here. Aside from the renaming changes, I see a couple numerical and indexing changes. What's the goal and what are they doing?

dhimmel:
> I see a couple numerical and indexing changes. What's the goal and what are they doing?

Currently, `_bearings_distribution` returns `(bin_counts, bin_edges)` where bin_edges are actually the centers of each bin with the first bin occurring at index position 0 and -1 (because 0 degrees is equivalent to 360 degrees).

The non-renaming changes are entirely related to removing the final repetitive value from the bin centers array. When giving the center of each bin, there is not much rationale for repeating the first bin at the end of the list. The repeated final value was a relic of `numpy.histogram`'s more general design, which in the case of a polar / degree distribution is not needed.

osmnx/bearing.py
```diff
@@ -281,6 +282,6 @@ def _bearings_distribution(
     count = np.roll(count, 1)
     bin_counts = count[::2] + count[1::2]
 
-    # because we merged the bins, their edges are now only every other one
-    bin_edges = bin_edges[range(0, len(bin_edges), 2)]
-    return bin_counts, bin_edges
+    # because we merged the bins, their centers are now only every other one
+    bin_centers = bin_edges[range(0, len(bin_edges) - 1, 2)]
```
dhimmel, line 286: `len(bin_edges)` is always even due to the double-bin implementation. It is 2 * num_bins.

If user inputs num_bins=1, then `range(0, len(bin_edges) - 1, 2)` return `0` (there is only a single bin centered at 0 degrees.

If user inputs num_bins=2, then `range(0, len(bin_edges) - 1, 2)` returns `0, 180`

gboeing: Ok I think that makes sense now.

osmnx/plot.py
```diff
@@ -758,7 +758,7 @@ def plot_orientation(  # noqa: PLR0913
 
     # positions: where to center each bar. ignore the last bin edge, because
     # it's the same as the first (i.e., 0 degrees = 360 degrees)
```
gboeing, line 760: It looks like this comment should be updated then.

gboeing (changes requested):

osmnx/plot.py
```diff
@@ -749,7 +749,7 @@ def plot_orientation(  # noqa: PLR0913
         }
 
     # get the bearings' distribution's bin counts and edges
```
gboeing, line 751: Another reference to edges here.

dhimmel:
Given the elegant but potentially confusing implementation of `_bearings_distribution` with respect to the bin splitting and merging, I revised variable names in this function as well as comments to provide greater clarity. Also added tests to `test_bearings`

osmnx/bearing.py
```diff
@@ -268,20 +268,28 @@ def _bearings_distribution(
...
-    bin_counts = count[::2] + count[1::2]
-
-    # because we merged the bins, their edges are now only every other one
-    bin_edges = bin_edges[range(0, len(bin_edges), 2)]
-    return bin_counts, bin_edges
+    split_bin_counts, split_bin_edges = np.histogram(
+        bearings,
+        bins=split_bin_edges,
+        weights=weights,
+    )
```
dhimmel, line 286: `np.histogram` passes through the `bins=split_bin_edges` argument to the output, so changed to the same variable name for clarity.

gboeing (approved):

Issue #1152, opened 2024-03-27
Add support for implicit maxspeed values


nikita03565:
### Contributing guidelines

- [X] I understand the contributing guidelines

### Documentation

- [X] My proposal is not addressed by the documentation or examples

### Existing issues

- [X] Nothing similar appears in an existing issue

### What problem does your feature proposal solve?

There are several hundreds of thousands edges with implicit maxspeed value like `"RO:urban"`. It'd be helpful to convert those string values to numbers in order to get the most accurate maxspeeds as well as travel times.

### What is your proposed solution?

Scrape maxspeed mapping from [wiki](https://wiki.openstreetmap.org/wiki/Key:maxspeed), put the mapping somewhere in config so users could override values as they see fit, like with other config values: `ox.settings.implicit_maxspeed_values["AR:rural"]=200` for example.
In `_clean_maxspeed` function try to safely get value from the mapping if parsing as number failed.

### What alternatives have you considered?

Don't put whole mapping in config, accept new parameter with this mapping in `add_edge_speeds` so users provide it themselves, but I think that keeping it with all default values in `settings` would result in a better out of the box experience for users.

### Additional context

Taginfo [link](https://taginfo.openstreetmap.org/keys/maxspeed#values) for maxspeed values.
Related PR #1153

PR #1153, opened 2024-03-27
Add support for implicit maxspeed

nikita03565:
refers to #1152

Added implicit speed mapping scraped from osm wiki to settings module
Updated `_clean_maxspeed` method to get values from that mapping if parsing as number fails


osmnx/settings.py
```diff
@@ -180,3 +184,152 @@
     "tunnel",
     "width",
 ]
+implicit_maxspeed_values = {
```
gboeing, line 187: My one main comment is that I don't think this should be a user configurable setting, but rather a constant in the `routing` module. This would make it consistent with how [we handle](https://github.com/gboeing/osmnx/blob/d614c9608165a614f796c7ab8e57257e7f9f1a63/osmnx/features.py#L49) the _POLYGON_FEATURES constant in the `features` module, for example. That is, if we scrape some defined values from OSM/Overpass and hardcode them, we should make them a constant and update them if the data source changes in the future. Because this is Python, there's the added benefit that a user can access and update these "private" variables if they wish to. For instance, `ox.features._POLYGON_FEATURES = {"new_keys": "new_values"}`. Let's handle these implicit maxspeed values the same way in the `routing` module.

gboeing (changes requested):
Thanks for this suggestion @nikita03565. I think you're right that handling these implicit values would be useful. I've left one comment for a requested change.

gboeing (approved):

gboeing:

PR #1157, opened 2024-04-03
Refactor the features module

gboeing:
In advance of v2.0.0 (see #1106), this PR refactors the internal functionality of the `features` module, leaving its public API unchanged. It produces only minor differences to the final GeoDataFrame (namely, no more "nodes" or "ways" columns containing lists of the members of the ways and relations, respectively).

It streamlines the code for maintainability and improves runtime and memory efficiency. We have cut the lines of code by about ~1/3 and runtime for a medium-size query by ~20%. See below for example (cached) timings:

```python
import osmnx as ox
import osmnx.features_v1
place = "Manhattan, New York, USA"
tags = {"building": True}

%%timeit
gdf1 = ox.features.features_from_place(place, tags)
# 3.89 s ± 49.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
gdf2 = ox.features_v1.features_from_place(place, tags)
# 4.83 s ± 71.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

PR #1160, opened 2024-04-18
Add node-specific tolerances to intersection consolidation

### LINKED ISSUE #1150: Conditional tolerance for intersection consolidation
### Contributing guidelines

### Documentation

### Existing issues

Current `consolidate_intersections` in OSMnx applies a uniform tolerance across a network, which may not suit complex urban models where different areas require different levels of detail. For example, a dense city center often needs a lower tolerance due to more intricate street layouts, while suburban areas might warrant higher tolerances.

I propose enhancing the `consolidate_intersections` function to accept a `tolerance_dict` parameter. This dictionary would map node identifiers to tolerance values, allowing conditional consolidation based on node-specific criteria. If a node does not appear in this dictionary, a default tolerance would be used.

One workaround is to consolidate subsets of the network separately with different tolerances and then attempt to merge them. However, this is cumbersome and can lead to complex merging issues, particularly at the boundaries of the subsets.

### Additional context

A minimal example of the proposed usage:

```python
import networkx as nx
import osmnx as ox
from shapely.geometry import Polygon

def consolidate_intersections(G, tolerance=10, rebuild_graph=True, dead_ends=False, reconnect_edges=True, tolerance_dict=None):
    """
    The extended description of parameters...

    tolerance_dict : dict
        A dictionary mapping node identifiers to tolerance values. This enables
        applying different consolidation tolerances to different nodes.
    """
    # Rest of the existing function's code...

    if tolerance_dict:
        # New logic to apply tolerances conditionally based on tolerance_dict
        pass  # Placeholder for the new functionality

    # Rest of the existing function's code...
```

Example usage with the proposed feature:

```python
# Assume G is a projected graph of an urban area
tolerance_dict = {node: 5 if node in inner_city_nodes else 15 for node in G.nodes()}
G_consolidated = ox.consolidate_intersections(G, tolerance_dict=tolerance_dict)
```

This example outlines how users might apply a tighter tolerance for a set of nodes (e.g., `inner_city_nodes`) while using a looser tolerance for the rest, making the function more flexible for many applications.

EwoutH:
This PR enhances the `consolidate_intersections` function in the OSMnx simplification module to support variable node consolidation tolerances. Instead of a single tolerance parameter, this update introduces the `tolerance` parameter that can accept either a float or a dictionary. This allows users to specify precise tolerance levels per node directly within their graph, enabling finer control over intersection consolidation.

#### Motivation
Currently `consolidate_intersections` applies a uniform tolerance across a network, which may not suit complex urban models where different areas require different levels of detail. For example, a dense city center often needs a lower tolerance due to more intricate street layouts, while suburban areas might warrant higher tolerances.

See #1150.

#### Key Changes
- The `tolerance` parameter now accepts a dictionary mapping node IDs to specific tolerance values, in addition to the existing option of a single float.
- The internal function `_merge_nodes_geometric` has been updated to dynamically adjust buffer distances based on node-specific tolerances provided in the dictionary, with a default fallback tolerance if specific values are not provided.

#### Usage Example
```python
import osmnx as ox
G = ox.graph_from_place('Delft, The Netherlands', network_type='drive')

# Apply a lower tolerance for high-connectivity nodes
tolerances = {node: 5 if count >= 4 else 10 for node, count in ox.stats.streets_per_node(G).items()}

G_consolidated = ox.consolidate_intersections(G, tolerance=tolerances)
```
In this example, nodes with higher street connectivity are assigned a smaller tolerance value (5), suitable for areas with denser road networks. Other nodes default to a higher tolerance of 10.

<hr>

Closes #1150.

This [branch](https://github.com/EwoutH/osmnx/tree/consolidate_conditional_tolerance) can be installed with the following command:
```
pip install -U -e git+https://github.com/EwoutH/osmnx@consolidate_conditional_tolerance#egg=osmnx
```
@anastassiavybornova, @martinfleis, @jdmcbr and @jGaboardi please test and review if you have the chance :)


osmnx/simplification.py
```diff
@@ -509,15 +519,25 @@ def _merge_nodes_geometric(G, tolerance):
...
-    merged = convert.graph_to_gdfs(G, edges=False)["geometry"].buffer(tolerance).unary_union
+    gdf_nodes = convert.graph_to_gdfs(G, edges=False)
+
+    if tolerance_column and tolerance_column in gdf_nodes.columns:
+        # Use the values from the tolerance_column as an array for buffering
+        buffer_distances = gdf_nodes[tolerance_column].values
+        # If a node does not have a value in the tolerance_column, use the default tolerance
+        buffer_distances[np.isnan(buffer_distances)] = tolerance
+        # Buffer nodes to the specified distances and merge them
+        merged = gdf_nodes['geometry'].buffer(distance=buffer_distances).unary_union
```
martinfleis, line 530:
```suggestion
        # If a node does not have a value in the tolerance_column, use the default tolerance
        buffer_distances = gdf_nodes[tolerance_column].fillna(tolerance)
        # Buffer nodes to the specified distances and merge them
        merged = gdf_nodes['geometry'].buffer(distance=buffer_distances).unary_union
```

A bit of simplification.

martinfleis (commented):
Given we are in the realm of networkx.MultiDiGraph here, a "column" does not really have a meaning. The conversion to gdf is internal only. Since we are fetching the value from a node attribute, I would suggest using something like `tolerance_attribute` instead.

Do you have any idea how would you define local tolerance for each node? One thing is to enable it in the function, the other how to actually use it.

martinfleis (commented):
Since you are touching the function anyway, I would advise using `dissolve` and `explode` over unary_union and iteration over parts. 

Like this (including my comments from above)

```py
def _merge_nodes_geometric(G, tolerance, tolerance_column):
    gdf_nodes = convert.graph_to_gdfs(G, edges=False)

    if tolerance_column and tolerance_column in gdf_nodes.columns:
        buffer_distances = gdf_nodes[tolerance_column].fillna(tolerance)
        merged = gdf_nodes['geometry'].buffer(distance=buffer_distances).dissolve()
    else:
        merged = gdf_nodes['geometry'].buffer(distance=tolerance).dissolve()
    
    return merged.explode(ignore_index=True)
```

osmnx/simplification.py
EwoutH: Good suggestion!

EwoutH:
Thanks for the swift review! I implemented the suggested changes.

> Do you have any idea how would you define local tolerance for each node?

I think we can leave that up to users. An example is given in the PR description, but you can basically do any kind of custom mask, loop, function etc.

> Since you are touching the function anyway, I would advise using `dissolve` and `explode` over unary_union and iteration over parts.

Sounds like a good idea, but I would like to keep this PR as small as possible (and atomic), so let's do that in another PR.

gboeing:
I'll try to take a look later this week. In the meantime, please base this PR off the `v2` branch. The `main` branch (soon to be `v1`) is for maintenance only at this point, rather than feature development.

EwoutH:
Right. If that’s the case, could you set `v2` as the [default branch](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/changing-the-default-branch)?

gboeing:
Yes, soon the `v2` branch will finally be merged into `main` after #1157 (and potentially this PR) are merged in, ideally later this week (and a permanent `v1` maintenance branch will split off of main beforehand). The branch chaos will at long last be at its end.

EwoutH:
I think I resolved al the conflicts when targeting the v2 branch correctly.

osmnx/simplification.py
```diff
@@ -451,7 +451,9 @@ def consolidate_intersections(
     dead_ends: bool = False,
     reconnect_edges: bool = True,
     node_attr_aggs: dict[str, Any] | None = None,
+    tolerance_attribute=None,
```
gboeing, line 454: Let's simplify this. Rather than introducing a new `tolerance_attribute` function parameter, please change the type hint of the existing `tolerance` parameter to `tolerance: float | dict[int, float] = 10`. This allows the user to either pass a single tolerance value, or a mapping of node IDs -> tolerance values. The argument can be handled accordingly with a simple conditional when it comes time to execute the buffer operation.

gboeing (changes requested):
Thanks @EwoutH. If a simple user-defined per-node tolerance like this is useful to the community, then it certainly makes sense to add it. I've requested a change. Additionally, the tests are currently failing. Consider running the pre-commit hooks locally for ease.

osmnx/simplification.py
EwoutH: I thought about this and it was one of the options discussed in https://github.com/gboeing/osmnx/issues/1150 (here: https://github.com/gboeing/osmnx/issues/1150#issuecomment-2062006902). I went for the second option, because I think it's more robust to have the tolerance values attached to the nodes themselves, than keep track of them in a dict. That means you have a separate data structure you have to keep up to date etc.

It also introduces cases like "what if a node id is missing in the dict"? Now it falls back on the default value, otherwise you need to handle that.

I'm open to discussing this further, let me know what you think.

gboeing: Let's keep it as a single `tolerance` parameter that accepts float or dict.

A few responses below.

> it's more robust to have the tolerance values attached to the nodes themselves, than keep track of them in a dict. That means you have a separate data structure you have to keep up to date etc.

The benefit of accepting a tolerance dict is that users *can* still store the tolerance values as node attributes and use that, if they wish, rather than keeping track of a separate data structure. Like this:

```python
ox.consolidate_intersections(G, tolerance=G.nodes(data="tolerance"))
```

So no drawback there. And they can run whatever tolerance estimation algorithm elsewhere then skip the overhead of having to create a special node attribute just to perform the node consolidation, if they wish. It's flexible.

> "what if a node id is missing in the dict"? Now it falls back on the default value, otherwise you need to handle that.

If a node is missing in the dict, it just shouldn't be buffered. The docs can state that explicitly. Users may not want to consolidate intersection in all parts of the network.

The function would work something like:

```python
# ...if the passed tolerance argument is type dict instead of float
gdf_nodes = ox.convert.graph_to_gdfs(G, edges=False)
tolerances = pd.Series(tolerance).reindex(gdf_nodes.index)
gdf_nodes.buffer(tolerances).fillna(gdf_nodes["geometry"])
```

EwoutH: Thanks for the detailed response. I agree, this is a cleaner API!

I will try to implement it later today.

EwoutH:
Reimplemented `tolerance` to accept a dict, removed the `tolerance_attribute` argument, updated the docstring and updated the PR message.

I tested it on one of my own notebooks, and it works beautifully!
```Python
# Create a dict mapping node ids to tolerances: 10 if part of the city network, 50 if part of the surrounding area network
tolerance_dict = {node: 10 if data["network"] == city_name else 50 for node, data in merged_network.nodes(data=True)}

# Consolidate intersections
merged_network = ox.consolidate_intersections(merged_network, tolerance=tolerance_dict, rebuild_graph=True)
```

CC @anastassiavybornova, @martinfleis, @jdmcbr and @jGaboardi, please test and review if you have the chance :)

osmnx/simplification.py
```diff
@@ -558,14 +567,25 @@ def _merge_nodes_geometric(G: nx.MultiDiGraph, tolerance: float) -> gpd.GeoSerie
...
     merged
         The merged overlapping polygons of the buffered nodes.
     """
-    # buffer nodes GeoSeries then get unary union to merge overlaps
-    merged = convert.graph_to_gdfs(G, edges=False)["geometry"].buffer(tolerance).unary_union
+    gdf_nodes = convert.graph_to_gdfs(G, edges=False)
+
+    if isinstance(tolerance, dict):
+        # Map tolerances to node IDs, using NaN for nodes not in the dictionary, and fill NaN with zero
+        buffer_distances = gdf_nodes.index.to_series().map(tolerance).fillna(0)
```
martinfleis, line 582: Just to ensure you are aware - buffering a point with a distance 0 returns POLYGON EMPTY. After union, some points will not be within any geometry. I am not sure if that is an issue or not, just wanted to flag this. The left sjoin that happens later will result in many points with cluster None. I would just double check that these are not actually considered one cluster and merged together.

This certainly needs a test case with missing keys in the dict.

EwoutH: Agreed. I won’t have time in the near future to write those, is anyone wants to continue from this please do so!

gboeing: See my code suggestion at https://github.com/gboeing/osmnx/pull/1160#discussion_r1578804479

It essentially buffers missing nodes from dict with None then fills resulting null geometries with the original point geometries. This was meant to avoid empty geometries. @martinfleis does this address the issue you're raising?

martinfleis: In principle yes, that code not really. `fillna` does not see POLYGON EMPTY as missing values, so you need an explicit mask for that.

gboeing: @martinfleis to clarify, it was this snippet:

```python
# ...if the passed tolerance argument is type dict instead of float
gdf_nodes = ox.convert.graph_to_gdfs(G, edges=False)
tolerances = pd.Series(tolerance).reindex(gdf_nodes.index)
gdf_nodes.buffer(tolerances).fillna(gdf_nodes["geometry"])
```

It does not buffer a missing node's point with distance 0 (which would indeed result in POLYGON EMPTY). It instead buffers them with None (which results in None, right?). Then `fillna` should work.

gboeing: Just ran a quick local test and it appears to be so. The result is a collection of buffer Polygons (when node tolerance is missing in dict) and original Points (when it is missing).

martinfleis: Ah, yes. I got confused by `fillna(0)` in this PR. Yes, this should resolve it and given the point is within self, the sjoin later should be fine.

EwoutH: Implemented!

Thanks for catching this.

gboeing: Thanks @martinfleis. @EwoutH I'll take a final review this morning and provide any final comments.

osmnx/simplification.py
```diff
@@ -9,10 +9,11 @@
 import geopandas as gpd
 import networkx as nx
 import numpy as np
-from shapely import LineString
-from shapely import MultiPolygon
-from shapely import Point
-from shapely import Polygon
+import pandas as pd
+from shapely.geometry import LineString
```
gboeing, line 13: These geometry imports can be left as `from shapely import ...`

osmnx/simplification.py
```diff
@@ -463,6 +464,11 @@ def consolidate_intersections(
     Note `tolerance` represents a per-node buffering radius: for example, to
     consolidate nodes within 10 meters of each other, use `tolerance=5`.
 
+    It's also possible to specify difference tolerances for each node. This can
+    be done by adding an attribute to each node with contains the tolerance, and
+    passing the name of that argument as tolerance_attribute argument. If a node
```
gboeing, line 469: Obsolete reference to `tolerance_attribute`.

osmnx/simplification.py
```diff
@@ -558,14 +568,29 @@ def _merge_nodes_geometric(G: nx.MultiDiGraph, tolerance: float) -> gpd.GeoSerie
...
-    # buffer nodes GeoSeries then get unary union to merge overlaps
-    merged = convert.graph_to_gdfs(G, edges=False)["geometry"].buffer(tolerance).unary_union
+    gdf_nodes = convert.graph_to_gdfs(G, edges=False)
+
+    if isinstance(tolerance, dict):
+        # Create a Series of tolerances, reindexed to match the nodes
+        tolerances = pd.Series(tolerance).reindex(gdf_nodes.index)
+        # Buffer nodes to the specified distance
+        buffered_geoms = gdf_nodes.geometry.buffer(tolerances)
+        # Replace POLYGON EMPTY with original geometries if buffer distance was effectively zero or missing
```
gboeing, line 586: This only replaces resulting null geometries with original geometries. Any POLYGON EMPTY resulting from a 0 buffer is ignored (and eventually discarded in a subsequent step), which is fine, but this comment should be edited for accuracy.

osmnx/simplification.py
```diff
@@ -558,14 +568,29 @@ def _merge_nodes_geometric(G: nx.MultiDiGraph, tolerance: float) -> gpd.GeoSerie
...
+        buffered_geoms = buffered_geoms.fillna(gdf_nodes["geometry"])
+    else:
+        # Buffer nodes to the specified distance
+        buffered_geoms = gdf_nodes.geometry.buffer(tolerance)
+
+    # Merge overlapping geometries into a single geometry
+    merged = buffered_geoms.unary_union
 
     # if only a single node results, make it iterable to convert to GeoSeries
     merged = MultiPolygon([merged]) if isinstance(merged, Polygon) else merged
```
gboeing, line 596: This needs to be:

```python
merged = merged.geoms if hasattr(merged, "geoms") else merged
```

because we can no longer guarantee the narrow result of Polygon or MultiPolygon (because we're doing `fillna` with the original Point geometries, which would result in a GeometryCollection, for instance).

gboeing (changes requested):
Thanks @EwoutH. I left a few final comments.

In addition, please 1) update the changelog, and 2) add simple tests to hit the changed lines to maintain coverage. I will then merge.

osmnx/simplification.py
EwoutH: I haven't touched those lines in this PR, I would like to keep this changeset atomic. Also maybe good to leave it explicit over implicit.

osmnx/simplification.py
EwoutH: Fixed

osmnx/simplification.py
EwoutH: Implemented

EwoutH:
Thanks for the review. All comments should be addressed.

Note that the tests don't cover the actual behaviour, just check if passing a dictionary (with and without all nodes covered) leads to a runtime error or not. While not ideal, this is in line with the other tests currently.

EwoutH:
CI is stuck on a type hinting thing, it doesn't like the int values in the tolerance dict, while is doesn't find it a problem when inputted directly just a few lines above.

Maybe we should create a custom "number" type that `int | float` and use that throughout.

Preferably in a separate PR though, I'm leaving for a week tonight.

martinfleis:
> only GeoDataFrame has dissolve, and not GeoSeries right?

Correct, I didn't realise that.

gboeing:
> CI is stuck on a type hinting thing, it doesn't like the int values in the tolerance dict, while is doesn't find it a problem when inputted directly just a few lines above.
> 
> Maybe we should create a custom "number" type that `int | float` and use that throughout.

You had to type hint the dict for typeguard to know what data type comes out of `G.nodes`. Fixed in e4e972e9602e84394d81f5f8085a67b6f68cf4e2.

gboeing (approved):

gboeing:
> Do you have any idea how would you define local tolerance for each node? One thing is to enable it in the function, the other how to actually use it.

@martinfleis yes, that is along the lines of what I was saying earlier in the issue: https://github.com/gboeing/osmnx/issues/1150#issuecomment-2061936624

It's relatively straightforward to implement a per-node tolerance, but the real challenge is in actually defining it per node. They are only useful in tandem, and the latter is infinitely variable depending on the theoretical and empirical needs of the individual analysis, so I have been reluctant to include this functionality in the past. I believe defining the local tolerance will most likely remain outside of OSMnx's scope accordingly. @EwoutH identified a group of folks who want to implement per-node tolerance, and since his solution hands off the definition of tolerance values to the user, I think this PR satisfies my scope and implementation concerns. The rest is up to the users now.

EwoutH:
Thanks for merging!

> I think this PR satisfies my scope and implementation concerns. The rest is up to the users now.

Agreed! I think we landed on a quite elegant solution that still offers a lot of flexibility to users.

gboeing:

PR #1171, opened 2024-05-02
release version 2.0.0 beta

gboeing:
First pre-release of v2.0.0 for beta testing. See changelog for details.


gboeing:
Just a note that [PEP 440](https://peps.python.org/pep-0440/) is now a historical document and points to PyPA's [Version Specifiers](https://packaging.python.org/en/latest/specifications/version-specifiers/) as the canonical standard now. So the OSMnx project will conform to those specs for version specs including pre-releases.

Issue #952, opened 2023-05-15
Buffer distance should be flexible in the function graph_from_polygon() by adding parameter in file 'osmnx/graph.py': parameter 'buffer_dist' should be added in the list of parameters of function graph_from_polygon() so that flexible buffer distance can be specified.


Prajwalita:
**Original function:**

 def graph_from_polygon(
    polygon,
    network_type="all_private",
    simplify=True,
    retain_all=False,
    truncate_by_edge=False,
    clean_periphery=True,
    custom_filter=None,
):
 """
    Create a graph from OSM within the boundaries of some shapely polygon.

    Parameters
    ----------
    polygon : shapely.geometry.Polygon or shapely.geometry.MultiPolygon
        the shape to get network data within. coordinates should be in
        unprojected latitude-longitude degrees (EPSG:4326).
    network_type : string {"all_private", "all", "bike", "drive", "drive_service", "walk"}
        what type of street network to get if custom_filter is None
    simplify : bool
        if True, simplify graph topology with the `simplify_graph` function
    retain_all : bool
        if True, return the entire graph even if it is not connected.
        otherwise, retain only the largest weakly connected component.
    truncate_by_edge : bool
        if True, retain nodes outside boundary polygon if at least one of
        node's neighbors is within the polygon
    clean_periphery : bool
        if True, buffer 500m to get a graph larger than requested, then
        simplify, then truncate it to requested spatial boundaries
    custom_filter : string
        a custom ways filter to be used instead of the network_type presets
        e.g., '["power"~"line"]' or '["highway"~"motorway|trunk"]'. Also pass
        in a network_type that is in settings.bidirectional_network_types if
        you want graph to be fully bi-directional.

    Returns
    -------
    G : networkx.MultiDiGraph

   Notes
    -----
    You can configure the Overpass server timeout, memory allocation, and
    other custom settings via the `settings` module. Very large query areas
    will use the utils_geo._consolidate_subdivide_geometry function to perform
    multiple queries: see that function's documentation for caveats.
    """

**Suggested Feature proposal:**

def graph_from_polygon(
    polygon,
    network_type="all_private",
    simplify=True,
    retain_all=False,
    truncate_by_edge=False,
    clean_periphery=True,
    **buffer_dist=None,**
    custom_filter=None,
):

 """
    Create a graph from OSM within the boundaries of some shapely polygon.

    Parameters
    ----------
    polygon : shapely.geometry.Polygon or shapely.geometry.MultiPolygon
[...]

PR #1174, opened 2024-05-06
Similar functionality is available for the function graph_from_point(). But it is not available for def graph_from_polygon(). Here user can get buffer with fixed 500 meters. However the value of buffer is not flexible. My proposal is that we can also add one new parameter buffer_dist(float value) so that user can give any buffer value to create a graph with the buffer. Because I faced this problem while working on a government project. I am sending push request for this update. Please review my proposal.

Prajwalita:
Similar functionality is available for the function graph_from_point(). But it is not available for def graph_from_polygon(). Here user can get buffer with fixed 500 meters. However the value of buffer is not flexible. My proposal is that we can also add one new parameter buffer_dist(float value) so that user can give any buffer value to create a graph with the buffer.

# Read these instructions carefully

Before you proceed, review the contributing guidelines in the CONTRIBUTING.md file, especially the sections on project coding standards and tests. Please edit the changelog to reflect your changes.

In this pull request, please include:

- a reference to related issue(s)
- a description of the changes proposed in the pull request
- an example code snippet illustrating usage of the new functionality

### DIFF
--- a/osmnx/graph.py
+++ b/osmnx/graph.py
@@ -102,6 +102,7 @@ def graph_from_bbox(
         retain_all=retain_all,
         truncate_by_edge=truncate_by_edge,
         custom_filter=custom_filter,
+        buffer_dist=None,
     )
 
     msg = f"graph_from_bbox returned graph with {len(G):,} nodes and {len(G.edges):,} edges"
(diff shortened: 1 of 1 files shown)


gboeing:
Closing for the reasons explained in https://github.com/gboeing/osmnx/issues/952#issuecomment-2096382173

PR #1179, opened 2024-05-30
Replace deprecated GeoPandas unary_union with union_all(), require geopandas>=1.0.1

### LINKED ISSUE #1178: FutureWarning: The 'unary_union' attribute is deprecated
### Contributing guidelines

### Documentation

- [X] My problem is not addressed by the documentation or examples

### Existing issues

- [X] My problem does not appear in an existing issue

### What operating system and Python version are you using?

Python 3.12

### What OSMnx version are you using?

2.0.0beta0

### Environment packages and versions

```shell
Python 3.12
OSMnx 2.0.0beta0
Shapely 2.0.4
Geopandas 1.0.0a1
```

### How did you install OSMnx?

Other (describe below)

### Problem description

When using `consolidate_intersections()`, the following warning pops up:

osmnx\simplification.py:668: FutureWarning: The 'unary_union' attribute is deprecated, use the 'union_all()' method instead.
  subcluster_centroid = node_points.loc[idx].unary_union.centroid

Looks like this is the source:

https://github.com/gboeing/osmnx/blob/a99fc4d7dd9ee13b256347af7cdcb58d214e8915/osmnx/simplification.py#L668

I'm running Geopandas 1.0.0a1, that might be the source of it. So just a head up.

### Complete minimal reproducible example

```python
Run consolidate_intersections() in the environment listed above.
```

EwoutH:
Replace the deprecated GeoPandas `.unary_union` attribute with the `.union_all()` method, as recommended in https://github.com/geopandas/geopandas/pull/3007.

Closes #1178 and should help make OSMnx be compatible with GeoPandas 1.0.0.

Furthermore, note that `unary_union` is also imported from `shapely.ops` somewhere:

https://github.com/gboeing/osmnx/blob/a99fc4d7dd9ee13b256347af7cdcb58d214e8915/osmnx/features.py#L32

This PR doesn't adres that, but maybe we should?

### DIFF
--- a/osmnx/_osm_xml.py
+++ b/osmnx/_osm_xml.py
@@ -213,7 +213,7 @@ def _save_graph_xml(
 
     # convert graph to node/edge gdfs and create dict of spatial bounds
     gdf_nodes, gdf_edges = convert.graph_to_gdfs(G, fill_edge_geometry=False)
-    coords = [str(round(c, PRECISION)) for c in gdf_nodes.unary_union.bounds]
+    coords = [str(round(c, PRECISION)) for c in gdf_nodes.union_all().bounds]
     bounds = dict(zip(["minlon", "minlat", "maxlon", "maxlat"], coords))
 
     # add default values (if missing) for standard attrs
--- a/osmnx/features.py
+++ b/osmnx/features.py
@@ -267,7 +267,7 @@ def features_from_place(
     """
     # extract the geometry from the GeoDataFrame to use in query
     gdf_place = geocoder.geocode_to_gdf(query, which_result=which_result)
-    polygon = gdf_place["geometry"].unary_union
+    polygon = gdf_place["geometry"].union_all()
     msg = "Constructed place geometry polygon(s) to query Overpass"
     utils.log(msg, level=lg.INFO)
 
--- a/osmnx/graph.py
+++ b/osmnx/graph.py
@@ -357,7 +357,7 @@ def graph_from_place(
     """
     # extract the geometry from the GeoDataFrame to use in query
     gdf_place = geocoder.geocode_to_gdf(query, which_result=which_result)
-    polygon = gdf_place["geometry"].unary_union
+    polygon = gdf_place["geometry"].union_all()
 
     msg = "Constructed place geometry polygon(s) to query Overpass"
     utils.log(msg, level=lg.INFO)
--- a/osmnx/plot.py
+++ b/osmnx/plot.py
@@ -553,7 +553,7 @@ def plot_figure_ground(
     node_sizes: list[float] | float = [node_widths[node] for node in Gu.nodes]
 
     # define the view extents of the plotting figure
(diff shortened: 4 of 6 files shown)


EwoutH:
Right, I now understand: `unary_union` will be deprecated _starting from_ GeoPandas 1.0. It's replacement, `union_all()`, isn't yet available in current stable GeoPandas.

So this PR has to wait until the GeoPandas 1.0.0 release (unless we get a backport).

martinfleis:
It needs to wait for 1.0.0 but you may want to do this conditionally based on a version. See https://github.com/pysal/momepy/blob/a64d274d3278eaca112deda7446173c49c26dba6/momepy/elements.py#L218

gboeing:
@EwoutH thanks. I have a branch (apparently only locally because I don't seem to have pushed it) that uses the GeoPandas 1.0.0 pre-release. Our CI uses conda-forge for installation, so until 1.0.0 is available there we're not going to be testing against it formally here.

The current plan is to make OSMnx 2.0 require GeoPandas >= 1.0. And OSMnx < 2.0 will pin max dependency versions (including GeoPandas < 1.0) so it remains working into the future without having to update it as dependencies change.

gboeing:
Max dependency versions pinned for v1 branch in 385a90a.

EwoutH:
GeoPandas 1.0 is released, and I rebased this PR. The CI still fails because the GeoPandas is capped at `<0.15`.

martinfleis:
Not sure what are the backwards compatibility promises of osmnx but I've been doing this elsewhere conditionally, based on the loaded version of geopandas. See https://github.com/pysal/momepy/blob/92bd9186430bafd21b5027107f8356cdf994bf3d/momepy/elements.py#L237 for an example.

EwoutH:
Personally, I think for the new osmnx 2.0 version we can (and should) require `geopandas>=1.0`.

gboeing:
> GeoPandas 1.0 is released, and I rebased this PR. The CI still fails because the GeoPandas is capped at <0.15.

@EwoutH the main branch requires geopandas>=0.12 in pyproject.toml, with no cap. If geopandas 1.0 is available now, you could edit that line in your PR to require >=1.0.

I'm on travel this week so can't take a more careful look right now. But if you have trouble I can check closer in a couple days.

EwoutH:
Took some figuring out, but Mamba cache only rebuilds when the `environments/tests/env-ci.yml` is modified. Not a big problem, but it might be less convenient for less experienced contributors.

Anyway, it's now [stuck](https://github.com/gboeing/osmnx/actions/runs/9876113429/job/27274393291?pr=1179#step:8:56) on a `AttributeError: 'GeoSeries' object has no attribute 'union_all'`. @martinfleis is `.union_all()` not an exposed method in a GeoSeries? How would you handle this properly?

EwoutH:
Okay, it took some more forcing the CI and bumping the requirement, to actually install GeoPandas 1.0.1. But everything is green now.

gboeing (approved):

PR #1188, opened 2024-06-23
added open-elevation public api in addition to google's

mitanshu7:
- a reference to related issue(s): https://github.com/gboeing/osmnx/issues/341#issuecomment-548892510
- a description of the changes proposed in the pull request: Adds another function called `add_node_elevations_open` which only requires the locations and nothing else.
- an example code snippet illustrating usage of the new functionality: 
```
import osmnx as ox
# get the street network for san francisco
place = "San Francisco"
place_query = {"city": "San Francisco", "state": "California", "country": "USA"}
G = ox.graph_from_place(place_query, network_type="drive")
# add elevation to each of the nodes, using the open elevation API
G = ox.elevation.add_node_elevations_open(G)
```

In case we want to fully replace google's elevation api, please let me know, the swap is fairly simple and i will create a new PR. 

Although I should mention the free [public api](https://github.com/Jorl17/open-elevation/blob/master/docs/api.md) may not be very performant. Hence the reduction of batch size from 512 to 128, and a default wait of 0.5 seconds to not overload the server.

### DIFF
--- a/osmnx/elevation.py
+++ b/osmnx/elevation.py
@@ -215,7 +215,7 @@ def add_node_elevations_google(
     # round coordinates to 6 decimal places (approx 5 to 10 cm resolution)
     node_points = pd.Series({n: f"{d['y']:.6f},{d['x']:.6f}" for n, d in G.nodes(data=True)})
     n_calls = int(np.ceil(len(node_points) / batch_size))
-    domain = _http._hostname_from_url(settings.elevation_url_template)
+    domain = _http._hostname_from_url(settings.elevation_url_template_google)
 
     msg = f"Requesting node elevations from {domain!r} in {n_calls} request(s)"
     utils.log(msg, level=lg.INFO)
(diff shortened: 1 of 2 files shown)


gboeing:
You don't need a new function for this. OSMnx can already do it:

```python
import osmnx as ox
ox.settings.elevation_url_template = "https://api.open-elevation.com/api/v1/lookup?locations={locations}"
G = ox.graph_from_place("Piedmont, CA, USA", network_type="drive")
Ge = ox.elevation.add_node_elevations_google(G, batch_size=128, pause=1)
```

Per [the docs](https://osmnx.readthedocs.io/en/latest/user-reference.html#osmnx.elevation.add_node_elevations_google):

> By default, this uses the Google Maps Elevation API but you can optionally use an equivalent API with the same interface and response format, such as Open Topo Data, via the settings module’s elevation_url_template. The Google Maps Elevation API requires an API key but other providers may not.

OSMnx supports any elevation API with an public interface compatible with Google's, including https://www.opentopodata.org/ and https://open-elevation.com/

mitanshu7:
Whoops. Thank you so much for clarifying! I should have thought about it earlier than just messing up the code 😅

gboeing:
I added further clarification to the docs in #1189.

PR #1192, opened 2024-07-03
Error from XML String results from reverse geocode _nominatim_request

willforker:
While developing a script that uses osmnx capabilities, I encountered an issue when I wanted to use the _nominatim_request feature to reverse geocode. For example, when running the code...
`import osmnx as ox`
`params = {'lat':43.70776104984752,'lon':-72.29470591308431}`
`reverse_dict = ox._nominatim._nominatim_request(params, request_type='reverse')`

...an InsufficientResponseError is returned, with the results of the reverse geocode populating the error message:

`InsufficientResponseError: 'nominatim.openstreetmap.org' responded: 200 OK <?xml version="1.0" encoding="UTF-8" ?>
<reversegeocode timestamp="Wed, 03 Jul 2024 16:42:24 +00:00" attribution="Data © OpenStreetMap contributors, ODbL 1.0. http://osm.org/copyright" querystring="lat=43.70776104984752&amp;lon=-72.29470591308431&amp;format=xml"><result place_id="10173084" osm_type="way" osm_id="245204742" ref="The Goat Path" lat="43.7079428" lon="-72.2948434" boundingbox="43.7065509,43.7153819,-72.2964453,-72.2859228" place_rank="27" address_rank="27">The Goat Path, Hanover, Grafton County, New Hampshire, 03755, United States</result><addressparts><road>The Goat Path</road><town>Hanover</town><county>Grafton County</county><state>New Hampshire</state><ISO3166-2-lvl4>US-NH</ISO3166-2-lvl4><postcode>03755</postcode><country>United States</country><country_code>us</country_code></addressparts></reversegeocode>`

Looking into the _nominatim.py and _http.py scripts, I discovered this was due to the expectation in the _parse_response function that the response from the API request would be either a dictionary or list of dictionaries (_http.py line 320): `response_json: dict[str, Any] | list[dict[str, Any]] = response.json()`

Since it appears the reverse geocode results are a xml formatted string, this trips the InsufficientResponseError. My proposed code suggests that the error only be raised should the string not be able to be parsed as an XML (I have included functions developed out of the python standard library 'xml' module to parse a string XML into a JSON / Dict). This solution then returns the result as a list of dicts when prompted with the same request shown earlier:
`[{'reversegeocode': {'result': 'The Goat Path, Hanover, Grafton County, New Hampshire, 03755, United States',
   'addressparts': {'road': 'The Goat Path',
    'town': 'Hanover',
    'county': 'Grafton County',
    'state': 'New Hampshire',
    'ISO3166-2-lvl4': 'US-NH',
    'postcode': '03755',
    'country': 'United States',
    'country_code': 'us'}}}]`

Only the _http.py and _nominatim.py scripts are altered, and all in order to account for these String-XML type responses seen in the reverse geocoding request.

### DIFF
--- a/osmnx/_http.py
+++ b/osmnx/_http.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import json
 import logging as lg
 import socket
+import xml.etree.ElementTree as ET
 from hashlib import sha1
 from pathlib import Path
 from typing import Any
@@ -21,6 +22,31 @@ from ._errors import ResponseStatusCodeError
 # capture getaddrinfo function to use original later after mutating it
 _original_getaddrinfo = socket.getaddrinfo
 
+def xml_to_dict(element):
+    # If the element has no children, return its text content
+    if not list(element):
+        return element.text
+    
+    # Create a dictionary to hold the element's data
+    result = {}
+    
+    for child in element:
+        child_result = xml_to_dict(child)
+        
+        # Handle multiple children with the same tag
+        if child.tag in result:
+            if isinstance(result[child.tag], list):
+                result[child.tag].append(child_result)
+            else:
+                result[child.tag] = [result[child.tag], child_result]
+        else:
+            result[child.tag] = child_result
+            
+    return result
+
+def parse_xml_string(xml_string):
+    root = ET.fromstring(xml_string)
+    return {root.tag: xml_to_dict(root)}
 
 def _save_to_cache(
     url: str,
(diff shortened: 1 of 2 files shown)


gboeing:
As you can see from its `_` naming, this module and this function are private/internal to the package and not part of the package's public API. But it's Python, so end users can use them at their own risk.

> it appears the reverse geocode results are a xml formatted string

That's only because you didn't specify a "json" format in your params.

```python
import osmnx as ox
ox.settings.log_console = True
params = {'lat': 43.70776, 'lon': -72.29470, 'format': 'json'}
d = ox._nominatim._nominatim_request(params, request_type='reverse')
```

PR #1196, opened 2024-07-20
Standardize bbox coord order to left, bottom, right, top

### LINKED ISSUE #1186: Change bounding box coordinates' order for v2.0.0?

Geographic bounding boxes usually follow the standard format of `left, bottom, right, top` (e.g., at [OpenStreetMap](https://wiki.openstreetmap.org/wiki/Bounding_box) and [Shapely](https://shapely.readthedocs.io/en/stable/manual.html#shapely.geometry.box)). For whatever reason, OSMnx instead expects them in the format of `top, bottom, right, left` (equivalent to `north, south, east, west`). I'm probably responsible for this oddity, but the reason behind it is lost to time. Currently, users have to read the docstring to know to use this non-standard format.

We should consider changing the expected bounding box format package-wide to the usual standard `left, bottom, right, top` in advance of the v2.0.0 release (which will have some breaking changes anyway).

Drawbacks: old code using the old format will break in v2.0.0.

Advantages: OSMnx will be aligned with global standards, creating less surprise for new users and eliminating a common source of unexpected bugs.

### Additional context

I'd be grateful for any comments/feedback on whether the advantages are worth the drawbacks. It'd be an easy fix prior to the next beta release, if it's worth doing. FWIW, my current opinion is "yes."

gboeing:
Resolves #1186 and makes a couple fixes to the CI tests now that numpy 2.0 is out.

Also added a FutureWarning to the v1 branch in 1e52f500ce2b8a71e720eb19aa919dbb7835f8a6.


EwoutH:
> added a FutureWarning to the v1 branch in [1e52f50](https://github.com/gboeing/osmnx/commit/1e52f500ce2b8a71e720eb19aa919dbb7835f8a6).

This is a good start. However, there is no action I can take directly as a user. Can we add a `future_bounds_order` parameter to the v1 branch that’s `False` by default, but can be set to `True` to use the future bounds order? This way users facing the warning can already apply this migration.

EwoutH:
I'm on vacation for the next two weeks, if it's still needed after that, sure.

PR #1203, opened 2024-08-12
replace gdal optional dependency with rio-vrt

### LINKED ISSUE #916: drop GDAL dependency
OSMnx currently depends on GDAL directly in the `elevation` module to build VRTs, because rasterio doesn't expose equivalent functionality directly. However, it's not a good idea to import both rasterio and GDAL into the same working environment (see https://github.com/rasterio/rasterio/issues/2573#issuecomment-1368135052).

Rasterio devs are currently proposing new functionality that would allow us to build VRTs directly with rasterio. If so, we could drop the GDAL dependency. See [this issue](https://github.com/rasterio/rasterio/issues/2573)  and this [PR](https://github.com/rasterio/rasterio/pull/2699).

gboeing:
Resolves #916


gboeing:
@12rambau do you have any insights? `rio-vrt` is giving me a "KeyError: 'blockxsize'" in the CI tests.

12rambau:
what I do in the lib is basically rebuilding the xml from the spec definition using the outputs from rasterio, if you share the image pieces that need to be stiched together I can make the appropriate adjustments. 

I'll be in holidays for 3 week so expect lots of delay in my next answers.

12rambau:
I released a v0.3.1 version of the lib with many modifications related to your issue. Would you mind relaunching the tests and see how it goes ?

gboeing:
@12rambau thanks. I will re-run the CI, but can't until this new version appears on conda-forge.

gboeing:
Ok CI tests are passing now.

Issue #1199, opened 2024-08-06
Unions filter


arkty:
### Contributing guidelines

- [X] I understand the contributing guidelines

### Documentation

- [X] My proposal is not addressed by the documentation or examples

### Existing issues

- [X] Nothing similar appears in an existing issue

### What problem does your feature proposal solve?

When developing city graph models, the size of the resulting graph might be important. Thanks to OSMnx, it's possible to use a pre-defined filters to request  a network, based on desired types of OSM ways. But sometimes, it's needed to add an additional ways from OSM into network or union them based on multiple filters.

### What is your proposed solution?

Make `custom_filter` accepts not only string, but a list of filters and then combine the results into single graph. 
Because OSMnx already supports sub-divided requests, it's easy to implement this without changing logic on actual graph creation (simplification, connectivity, etc). 

### What alternatives have you considered?

Requesting `osmnx.graph_from_*` multiple time and then merging result graphs manually without simplification.

### Additional context

The idea is described in [this PR](https://github.com/arkty/osmnx/pull/1).

gboeing:
Interesting idea for a relatively minimal change to make this a bit smoother for the user. I'm trying to think through the implications and trade-offs of this technique. Please share any insights as such. 

See also the discussion at https://github.com/gboeing/osmnx-examples/issues/90 for another (possibly infeasible) idea of addressing this.

arkty:
Another way to implement this is actually use overpass unions. 
This will require to change query creation. 

Example code (not tested).
```
    for polygon_coord_str in polygon_coord_strs:
        way_subquery_str = ""
        for way_filter_str in way_filter:
            way_subquery_str += f"(way{way_filter_str}(poly:{polygon_coord_str!r});"
        query_str = f"{overpass_settings};{way_subquery_str}>;);out;"
        yield _overpass_request(OrderedDict(data=query_str))
```

In some cases, `polygon_coord_str` can be really long (because poly statement will appear in each way query), therefore request payload size will be increased. I have not found any limits about request size on Overpass side, so this still can be an implementation option.

---

Adding support for `custom_filter` to be a list of filters looks like a simple change, but I am not sure will this confuse users or not, because it might be not clear that `custom_filter: list[str]` means union filter.

On the other hand, creating own query structure for `custom_filter` will lead to reproducing Overpass query language. And this not looks like a thing that can be easily developed and maintained.

gboeing:
[...]

PR #1204, opened 2024-08-13
Union filters

arkty:
# Union filters

Discussion: https://github.com/gboeing/osmnx/issues/1199

## Problem
When developing city graph models, the size of the resulting graph might be important. Thanks to OSMnx, it's possible to use a pre-defined filters to request  a network, based on desired types of OSM ways. But sometimes, it's needed to add an additional ways from OSM into network or union them based on multiple filters.

## Technical details
When using `custom_filter`, it's not possible to request additional roads by Overpass union condition.

## Example
We need to request additional road by name (The Road), ignoring all other conditions.

In Overpass this request will look like this:
```
(way[...](poly:'***'); way[name="The Road"](poly:'***'););
```

OSMnx constructs Overpass request like this, where `$filter` is pre-defined filter or provided by user.
```
[out:json][timeout:180];
(way[$filter](poly:'***');>;);
out;
```

## Proposal
Make `custom_filter` accepts not only string, but a list of filters and then combine the result into single graph. 
Because OSMnx already supports sub-divided requests, it's easy to implement this without changing logic on actual graph creation (simplification, connectivity, etc). 

## Usage example
On example of city Petrozavodsk we will use `graph_from_place ` with `network_type = "drive"`. This will give us a graph on a perfect level of details. But for some reason, two additional roads are required to be in model.

The request below correctly adds these roads.

```
ptz_graph = ox.graph_from_place(
    'Petrozavodsk, Russia',
    custom_filter=[
        op._get_network_filter("drive"),
        f'[name="улица Достоевского"]',
        f'[name="улица Калевалы"]'
    ]
)
```

### DIFF
--- a/osmnx/_overpass.py
+++ b/osmnx/_overpass.py
@@ -391,7 +391,9 @@ def _download_overpass_network(
     # the '>' makes it recurse so we get ways and the ways' nodes.
     for polygon_coord_str in polygon_coord_strs:
         for way_filter_str in way_filter:
-            query_str = f"{overpass_settings};(way{way_filter_str}(poly:{polygon_coord_str!r});>;);out;"
+            query_str = (
+                f"{overpass_settings};(way{way_filter_str}(poly:{polygon_coord_str!r});>;);out;"
+            )
             yield _overpass_request(OrderedDict(data=query_str))
 
 
--- a/osmnx/_overpass.py
+++ b/osmnx/_overpass.py
@@ -350,7 +350,7 @@ def _create_overpass_features_query(  # noqa: PLR0912
 def _download_overpass_network(
     polygon: Polygon | MultiPolygon,
     network_type: str,
-    custom_filter: str | None,
+    custom_filter: str | list[str] | None,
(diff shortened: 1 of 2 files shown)


EwoutH:
Would something like this be possible? How would that work API wise in with this PR?

```Python
custom_filter = '["highway"~"motorway|motorway_link"]'
construction_filter = '["construction"~"motorway|motorway_link"]'
full_filter = ???
```

arkty:
@EwoutH for OR filter it will be
```
highway_filter = '["highway"~"motorway|motorway_link"]'
construction_filter = '["construction"~"motorway|motorway_link"]'

# Full OR filter
custom_filter = [highway_filter, construction_filter]
```

EwoutH:
That’s a really clean syntax, love it!

gboeing:
@arkty tests are failing. Run the pre-commit hooks locally then push again.

arkty:
@gboeing i've added a commit with style fix, can you please run wf again.

EwoutH:
Awesome! Looking forward to using this in the next pre-release!

PR #1208, opened 2024-08-17
Resolve external warnings

EwoutH:
This PR addresses resolves several 3 issues caused by external dependencies, that were present one or multiple times in osmnx CI test runs.

- Upgrade scikit-learn to >=1.0:
   - Updated the dependency to `scikit-learn>=1.0`, resolving deprecation warnings related to `np.float` when using NumPy > 1.20.

- Resolve `SettingWithCopyWarning` in Pandas:
   - Fixed a `SettingWithCopyWarning` in `features.py` by ensuring a copy of the DataFrame slice is made before modifications.

- Fix `VisibleDeprecationWarning` in NumPy:
   - Addressed a `VisibleDeprecationWarning` in the `_verify_edge_attribute` function by explicitly setting `dtype=object` for a NumPy array to handle ragged nested sequences, maintaining future compatibility.

Some more details are in the commit messages of each commit.

CI is [passing](https://github.com/EwoutH/osmnx/actions?query=branch%3Aresolve_warnings) on my test branch, where I triggered the CI explicitly.

There's one external issue left, but I think that has to be fixed upstream (tracking: https://github.com/geopandas/geopandas/issues/3403). The other issues look to be triggered by OSMnx (or its tests) itself.


osmnx/features.py
```diff
@@ -680,7 +680,7 @@ def _filter_features(
         Filtered GeoDataFrame of features.
     """
     # remove any null or empty geometries then fix any invalid geometries
-    gdf = gdf[~(gdf["geometry"].isna() | gdf["geometry"].is_empty)]
```
gboeing, line 683: Isn't the `SettingWithCopyWarning` here a false positive? That is, we don't care about writing back to the original dataframe, because we're overwriting the variable itself. Given the expensiveness of `.copy()`, it may be better to filter/disable the warning on this line instead (assuming this is indeed a false positive).

gboeing: Also, regarding:

> Addressed a VisibleDeprecationWarning in the _verify_edge_attribute function by explicitly setting dtype=object for a NumPy array to handle ragged nested sequences, maintaining future compatibility.

How did you produce this warning? I'm using numpy 2 and haven't seen it. I'm trying to figure it out, but `dtype=object` doesn't make sense to me. This shouldn't be a ragged array, but rather always a $m \times 3$ array created from the tuples of `(u, v, attr_value)`.

EwoutH: All warnings were from a CI run at that time on the main branch.

Unfortunately, at this time I don't have the capacity to follow up on this, so feel free to close or modify this PR.

PR #1221, opened 2024-10-04
use sphinx 8

gboeing:
Bump sphinx version to build docs.


gboeing:
Looks like sphinx 8 is incompatible with Python 3.9. We'll punt this until we drop 3.9 support.

Issue #1253, opened 2024-12-12
Better "walk" filter for networks with separately mapped sidewalks


winnewoerp:
### Contributing guidelines

- [X] I understand the contributing guidelines

### Documentation

- [X] My proposal is not addressed by the documentation or examples

### Existing issues

- [X] Nothing similar appears in an existing issue

### What problem does your feature proposal solve?

Sidewalks have already been mapped as separate geometries in numerous neighborhoods worldwide. Where this is done correctly, the roadway (street centerline) is being tagged as not to be used by pedestrians using sidewalk=separate, sidewalk:both=separate, sidewalk:left=separate or sidewalk:right=separate. With the current version, the network type "walk" in OSMnx does still include such edges.

Test both versions here:
Current https://overpass-turbo.eu/s/1VBe
Proposed https://overpass-turbo.eu/s/1VBg

### What is your proposed solution?

I propose adding the necessary rules to the "walk" filter.

### What alternatives have you considered?

A workaround is of course to filter using a manual solution, but this can be avoided with the proposed solution.

### Additional context

```
    # New version of the "walk" filter in osmnx/_overpass.py

    filters = {}
    
    filters["walk"] = (
        f'["highway"]["area"!~"yes"]{settings.default_access}'
        f'["highway"!~"abandoned|bus_guideway|construction|cycleway|motor|no|planned|platform|'
        f'proposed|raceway|razed"]'
        f'["foot"!~"no"]["service"!~"private"]'
        f'["sidewalk"!~"separate"]["sidewalk:both"!~"separate"]'
        f'["sidewalk:left"!~"separate"]["sidewalk:right"!~"separate"]'
    )
```

gboeing:
Thanks @winnewoerp this looks great. Is this tagging convention documented as official on the OSM wiki? Could you link it here in a comment for reference?

gboeing:
Here's the answer to the question: https://wiki.openstreetmap.org/wiki/Tag:sidewalk%3Dseparate

winnewoerp:
So, the answer is "yes", right? Of course, this is currently not relevant for the largest part of the world, as this LoD of sidewalk mapping is still rare. But the other way round, there are no problems when it's included.

BTW: The filter(s) could also be written like this (saving a few bytes):

```
filters["walk"] = (
        f'[highway][area!=yes]{settings.default_access}'
        f'[highway!~"abandoned|bus_guideway|construction|cycleway|motor|no|planned|platform|'
[...]

PR #1254, opened 2024-12-12
extend 'walk' filter for separately mapped sidewalks

winnewoerp:
Solves #1253 (extended walk filter for separately mapped sidewalks). Description and code example can be found in the issue description.


gboeing:
Thanks!

Issue #1187, opened 2024-06-22
Graph route animations


nkleinbaer:
### Contributing guidelines

- [X] I understand the contributing guidelines

### Documentation

- [X] My proposal is not addressed by the documentation or examples

### Existing issues

- [X] Nothing similar appears in an existing issue

### What problem does your feature proposal solve?

Allow users to create animated plots of routes on graphs. For routes that cover all or most edges of a graph, a static plot doesn't show the route very clearly. 

### What is your proposed solution?

Use `matplotlib.animations.FuncAnimation` to animate route(s) step-by-step. 
Either:
1. Add a new `osmnx.plot.animate_graph_route` function for this feature
2. Add `animate: bool` argument to the existing `osmnx.plot.plot_graph_route` function

### What alternatives have you considered?

Of course users can implement this themselves outside of `osmnx`, not sure if this passes the "not trivially easy" criteria from contributing guidelines.

### Additional context

I've been using `osmnx` for a toy project looking at the [route inspection problem](https://en.wikipedia.org/wiki/Chinese_postman_problem). *Huge* thanks for this awesome project, made it easy for me to get started playing with the actual problem quickly instead of writing code to fetch road data and convert it into graphs. For my use case I found that plotting the entire route at once isn't particularly insightful -- my routes always cover the entire graph -- so I implemented animation for myself. Figured it may be useful to others and worth contributing back. 

I've technically made a liar of myself by checking the "Nothing similar appears in an existing issue" box. I see two old, closed issues mentioning animation:

- https://github.com/gboeing/osmnx/issues/226
- https://github.com/gboeing/osmnx/issues/109

In #226 the animation piece is only mentioned tangentially, in #109 a user asks for help doing this but doesn't explicitly propose/request a feature.

Can post a brief code snippet with minimal example a bit later today.

gboeing:
Thanks for using OSMnx and suggesting this.

> Can post a brief code snippet with minimal example a bit later today.

Sure, would you post the proposed code and I'll take a look?

> Of course users can implement this themselves outside of osmnx, not sure if this passes the "not trivially easy" criteria from contributing guidelines.

It's really just a matter of the complexity of the proposed code. Sometimes it's just easier to share a dozen lines of code here in a comment or on StackOverflow for others to reuse than it is to incorporate it into the package, depending on the general applicability of the use case.

nkleinbaer:
[...]

PR #1285, opened 2025-03-13
Add function to animate graph routes

nkleinbaer:
Opening a PR to implement feature proposed in #1187. Adds a new function `animate_graph_route` that is analogous to `plot_graph_route`, but shows a edge-by-edge animation instead of a static image. 

In 59fd139ac3f5c2046009065ef7e7650c9097042e I simply copied the existing `plot_graph_route` function and modified it to create an animation. Since this created duplication of code shared by the two funcs (i.e. plotting the graph itself and origin/destination nodes) I then added a 52ffb46372d220e21290d173f1a088b5398a4ff6 to refactor this out. Wanted to keep that as a separate commit commit so it's easy to drop per your advice (thinking perhaps you may want to avoid touching existing functionality?). 

For testing just added a single call to the new function in `test_routing`, following from what I saw done for `plot_graph_route`. Can add more tests if desired. 

Haven't updated the changelog yet, not sure if I should put this under the `2.0.2` section or a new section.

Example usage (adapted from [this example](https://github.com/gboeing/osmnx-examples/blob/main/notebooks/15-advanced-plotting.ipynb)):
```python
import osmnx as ox

G = ox.graph.graph_from_place(place, network_type="drive")

# impute missing edge speeds and calculate free-flow travel times
G = ox.routing.add_edge_speeds(G)
G = ox.routing.add_edge_travel_times(G)

# calculate 3 shortest paths, minimizing travel time
w = "travel_time"
orig, dest = list(G)[10], list(G)[-10]
route1 = ox.routing.shortest_path(G, orig, dest, weight=w)
orig, dest = list(G)[0], list(G)[-1]
route2 = ox.routing.shortest_path(G, orig, dest, weight=w)
orig, dest = list(G)[-100], list(G)[100]
route3 = ox.routing.shortest_path(G, orig, dest, weight=w)

# animate routes
fig, ax = ox.plot.animate_graph_route(G, route1, orig_dest_size=0, node_size=0, save=True, show=True, filepath='images/animation.mp4')
fig, ax = ox.plot.animate_graph_route(G, route2, orig_dest_size=0, node_size=0, save=True, show=True, filepath='images/animation.gif')
fig, ax = ox.plot.animate_graph_route(G, route3, orig_dest_size=0, node_size=0, save=True, show=True, filepath='images/animation.mov')
```

### DIFF
--- a/osmnx/__init__.py
+++ b/osmnx/__init__.py
@@ -34,6 +34,7 @@ from .io import load_graphml as load_graphml
 from .io import save_graph_geopackage as save_graph_geopackage
 from .io import save_graph_xml as save_graph_xml
 from .io import save_graphml as save_graphml
+from .plot import animate_graph_route as animate_graph_route
 from .plot import plot_figure_ground as plot_figure_ground
 from .plot import plot_footprints as plot_footprints
 from .plot import plot_graph as plot_graph
--- a/osmnx/__init__.py
+++ b/osmnx/__init__.py
@@ -34,7 +34,6 @@ from .io import load_graphml as load_graphml
 from .io import save_graph_geopackage as save_graph_geopackage
 from .io import save_graph_xml as save_graph_xml
 from .io import save_graphml as save_graphml
-from .plot import animate_graph_route as animate_graph_route
 from .plot import plot_figure_ground as plot_figure_ground
 from .plot import plot_footprints as plot_footprints
 from .plot import plot_graph as plot_graph
--- a/osmnx/plot.py
+++ b/osmnx/plot.py
@@ -337,7 +337,7 @@ def plot_graph_route(
     fig, ax
         The resulting matplotlib figure and axes objects.
     """
-    fig, ax = _prepare_graph_route_plot(
+    fig, ax, x, y = _prepare_graph_route_plot(
         G,
         route,
         orig_dest_color=route_color,
@@ -347,22 +347,14 @@ def plot_graph_route(
         **pg_kwargs,
     )
 
-    # assemble the route edge geometries' x and y coords then plot the line
-    x = []
-    y = []
-    for u, v in zip(route[:-1], route[1:]):
-        # if there are parallel edges, select the shortest in length
-        data = min(G.get_edge_data(u, v).values(), key=lambda d: d["length"])
-        if "geometry" in data:
-            # if geometry attribute exists, add all its coords to list
-            xs, ys = data["geometry"].xy
-            x.extend(xs)
-            y.extend(ys)
-        else:
-            # otherwise, the edge is a straight line from node to node
-            x.extend((G.nodes[u]["x"], G.nodes[v]["x"]))
-            y.extend((G.nodes[u]["y"], G.nodes[v]["y"]))
-    ax.plot(x, y, c=route_color, lw=route_linewidth, alpha=route_alpha)
+    # Flatten lists of coords and plot
+    ax.plot(
+        [coord for edge in x for coord in edge],
+        [coord for edge in y for coord in edge],
+        c=route_color,
+        lw=route_linewidth,
+        alpha=route_alpha,
+    )
 
     # save and show the figure as specified, passing relevant kwargs
     sas_kwargs = {"show", "close", "save", "filepath", "dpi"}
@@ -409,7 +401,7 @@ def animate_graph_route(
     fig, ax
         The resulting matplotlib figure and axes objects.
     """
-    fig, ax = _prepare_graph_route_plot(
+    fig, ax, x, y = _prepare_graph_route_plot(
         G,
         route,
         orig_dest_color=route_color,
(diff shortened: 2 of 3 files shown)


osmnx/plot.py
```diff
@@ -374,6 +376,119 @@ def plot_graph_route(
...
+        orig_dest_alpha=route_alpha,
+        orig_dest_size=orig_dest_size,
+        ax=ax,
+        **pg_kwargs,
+    )
+
+    # assemble list of lists of x and y coords for each edge
+    x = []
+    y = []
+    for u, v in zip(route[:-1], route[1:]):
```
gboeing, line 430: Probably better to use `pairwise` (everywhere we're doing this): https://docs.python.org/3/library/itertools.html#itertools.pairwise

osmnx/plot.py
```diff
@@ -374,6 +376,119 @@ def plot_graph_route(
...
+        data = min(G.get_edge_data(u, v).values(), key=lambda d: d["length"])
+        if "geometry" in data:
+            # if geometry attribute exists, append a new sublist with all its coords
+            xs, ys = data["geometry"].xy
+            x.append(list(xs))
+            y.append(list(ys))
+        else:
+            # otherwise, the edge is a straight line from node to node
+            x.append([G.nodes[u]["x"], G.nodes[v]["x"]])
+            y.append([G.nodes[u]["y"], G.nodes[v]["y"]])
```
gboeing, line 441: Isn't all of this code duplicated in the `plot_graph_route` function? Seems like it could be extracted from both and into the common helper function to keep it DRY.

osmnx/plot.py
```diff
@@ -374,6 +376,119 @@ def plot_graph_route(
...
+def animate_graph_route(
+    G: nx.MultiDiGraph,
+    route: list[int],
+    *,
+    route_color: str = "r",
+    route_linewidth: float = 4,
+    route_alpha: float = 0.5,
+    orig_dest_size: float = 100,
+    ax: Axes | None = None,
+    **pg_kwargs: Any,  # noqa: ANN401
```
gboeing, line 388: Does the user need more control than this? Animation speed? Repetition? Would fine-grain control be more easily exposed via keyword arguments to pass to matplotlib? Or?

gboeing (changes requested):

osmnx/plot.py
nkleinbaer: Technically it's a bit different, because in `plot_graph_route` it's just a flat list of coordinates, where as here it's creating sub-lists for each edge (so that the each step in the animation will be a full edge). But I can take this code and move it to the helper function, then just flatten the list in `plot_graph_route`.

osmnx/plot.py
```diff
@@ -335,15 +337,15 @@ def plot_graph_route(
...
+        route,
+        orig_dest_color=route_color,
+        orig_dest_alpha=route_alpha,
+        orig_dest_size=orig_dest_size,
+        ax=ax,
+        **pg_kwargs,
+    )
 
     # scatterplot origin and destination points (first/last nodes in route)
     od_x = (G.nodes[route[0]]["x"], G.nodes[route[-1]]["x"])
```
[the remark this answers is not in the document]
nkleinbaer: Yes it does, I think I messed this up when rebasing, thanks for catching that.

osmnx/plot.py
nkleinbaer: Yeah that makes sense, IMO it's worth adding the two you mentioned (speed and repetition) as explicitly defined/documented arguments here, then additionally allow passing arbitrary kwargs to matplotlib for more advanced/obscure things.

osmnx/__init__.py
```diff
@@ -34,6 +34,7 @@
 from .io import save_graph_geopackage as save_graph_geopackage
 from .io import save_graph_xml as save_graph_xml
 from .io import save_graphml as save_graphml
+from .plot import animate_graph_route as animate_graph_route
```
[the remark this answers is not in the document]
nkleinbaer: Addressed by [a3a0aa5](https://github.com/gboeing/osmnx/pull/1285/commits/a3a0aa54c0ec75df2bea4db2bbc41f43d971b7a2)

osmnx/plot.py
nkleinbaer: Refactored this out into the helper function in [d5801e8](https://github.com/gboeing/osmnx/pull/1285/commits/d5801e81519e58c05b6ad6248d3eb66fe7f24da8)

osmnx/plot.py
nkleinbaer: I tried changing this, but mypy wasn't happy with it. It looks like the minimum python version specified in pyproject.toml is 3.9, but this feature wasn't introduced until 3.10, so unless you want to drop support for 3.9 we can't use it.

osmnx/plot.py
nkleinbaer: The issue I'm running into with this is that we can only gobble up one dict of arbitrary kwargs, but then there are several places in this function where they may need to be passed: some may need to go to `plot_graph` (by way of `_prepare_graph_route_plot`), others may need to go to the constructor of `FuncAnimation`, others may need to go to `Animation.save`'s `savefig_kwargs` argument, and finally some may need to be passed directly to `Animation.save`. So I feel like I need to split them all out to avoid passing unexpected kwargs in the wrong spot ... something like this https://github.com/nkleinbaer/osmnx/commit/904ecda7a3532f4aaaeb25680844f50a776ca38e ... but that feels kinda hacky and difficult to maintain (not to mention I haven't tested out that passing any of these kwargs actually works as expected, that's just a draft). What are your thoughts?

gboeing:
@jabahm you expressed interest in animation in #1187. Do you have any comments/feedback for this PR?

osmnx/plot.py
gboeing: Ok, we'll defer this to the future then.

osmnx/plot.py
gboeing: @nkleinbaer perhaps each of the constant kwarg dicts you define in that commit at lines 404-460 could instead be individual function arguments of type dict. Then the user could pass individual kwarg dicts for each destination function.

gboeing:
@olivier-be thanks for your feedback. I agree we should make this useful for Jupyter users. We won't want to add an `ipython` dependency to the package to support it, but we could return the animation object to the caller, similar to how the static plotting function return `fig`, `ax` tuples to the caller.

gboeing:
Over time, it's become clear that this is going to be out of scope for the OSMnx project, given our increasing focus on maintainability and a tight API. So I'm going to close this. But, I'd like to thank everyone for contributing their ideas here! This functionality would be best suited as an external component or package building on OSMnx itself. That would give it the flexibility and thoroughness it needs to become a flexible, feature-rich animation tool.

Issue #1282, opened 2025-03-04
Problem with mutating socket.getaddrinfo to map hostname -> IP address


garvit-joshi:
### Contributing guidelines

- [x] I understand the contributing guidelines

### Documentation

- [x] My problem is not addressed by the documentation or examples

### Existing issues

- [x] My problem does not appear in an existing issue

### What operating system and Python version are you using?

Linux ######## 5.10.227-219.884.amzn2.aarch64 #1 SMP Tue Oct 22 16:38:25 UTC 2024 aarch64 aarch64 aarch64 GNU/Linux

### What OSMnx version are you using?

2.0.0

### Environment packages and versions

```shell
asn1crypto==1.5.1
async-timeout==4.0.3
beautifulsoup4==4.12.3
blinker==1.9.0
boto3==1.35.71
botocore==1.35.71
certifi==2024.7.4
charset-normalizer==3.3.2
click==8.1.7
Flask==3.1.0
geopandas==1.0.1
gunicorn==23.0.0
idna==3.8
itsdangerous==2.2.0
Jinja2==3.1.4
jmespath==1.0.1
joblib==1.4.2
lxml==5.3.0
MarkupSafe==2.1.5
networkx==3.4.2
numpy==2.1.3
osmnx==2.0.0
packaging==24.1
pandas==2.2.3
pyogrio==0.9.0
pyproj==3.6.1
python-dateutil==2.9.0.post0
python-louvain==0.16
pytz==2024.2
redis==5.2.0
redshift-connector==2.1.4
requests==2.32.3
s3transfer==0.10.2
scikit-learn==1.5.2
scipy==1.13.1
scramp==1.4.5
shapely==2.0.6
simplekml==1.3.6
six==1.16.0
soupsieve==2.6
threadpoolctl==3.5.0
tzdata==2024.1
tzlocal==5.2
urllib3==2.2.2
Werkzeug==3.1.3
xgboost==2.1.3
```

### How did you install OSMnx?

Pip

### Problem description

I'm facing an issue when using both osmnx and redshift-connector in the same Flask application. Specifically, when I make a request to an API that uses OSMnx, it overrides the socket.getaddrinfo function. Later, when a different API, which uses redshift-connector, is called, it results in an IndexError (tuple index out of range).

Steps to Reproduce:

1. Make an API request that uses OSMnx. This causes the following override of socket.getaddrinfo: /site-packages/osmnx/_http.py
```
[...]

PR #1286, opened 2025-03-15
#1282: Fix IndexError in _getaddrinfo by Safely Handling Missing Host…

garvit-joshi:
Reference to Related Issue(s):

Issue #1282


gboeing:
Would you update the changelog as well?

PR #1310, opened 2025-06-13
Added a function to give the nearest nodes connected to the largest connected component.

djs113:
When we try to route between 2 nodes, sometimes one or both of the target and source nodes will be an island node, that is a node that is not connected to the largest connected component in the graph, this causes an error while routing. This function aims to solve that problem by replacing the island node with the closes node that is connected to the largest connected component. If the node is not an island node, the same node is returned.

### DIFF
--- a/osmnx/distance.py
+++ b/osmnx/distance.py
@@ -538,3 +538,78 @@ def nearest_edges(
         return ne_array, dist_array
     # otherwise
     return ne_array
+
+
+def map_nodes_to_main_component(
+    G: nx.MultiDiGraph,
+    source_node: int,
+    target_node: int,
+) -> list[int]:
+    """
+    Map nodes to the largest strongly connected component in the graph.
+
+    If either the source or target node lies in an isolated subgraph
+    (i.e., outside the largest strongly connected component), this function
+    finds and returns the nearest node within the main component.
+
+    Parameters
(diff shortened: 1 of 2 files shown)


gboeing:
Thanks for using OSMnx! And thanks for proposing a new feature.

Per the [contributing guidelines](https://github.com/gboeing/osmnx/blob/main/CONTRIBUTING.md#if-you-have-a-feature-proposal):

 > The OSMnx project follows three principles when adding new functionality: 1) it is useful for a broad set of users, 2) it generalizes well, and 3) it is not trivially easy for users to implement themselves.

This functionality (finding nearest nodes in the largest connected component) already exists. For example, let's say you have a graph and some point coordinates:

```python
import osmnx as ox
G = ox.graph.graph_from_place("Piedmont, CA, USA", network_type="drive", retain_all=True)
points = ox.utils_geo.sample_points(G, 100)
```

You can find the nearest nodes in the largest connected component with a simple one-liner:

```python
ox.distance.nearest_nodes(ox.truncate.largest_component(G), points.x, points.y)
```

This existing functionality has the additional benefits of 1) seamlessly handling either projected or unprojected coordinates, and 2) calculating more-accurate haversine distances rather than Euclidean distances for unprojected coordinates.

PR #1316, opened 2025-06-18
streamline linkcheck

gboeing:
Builds on #1313


gboeing:
Nevermind, this actually does need other dev dependencies. See failure at https://github.com/gboeing/osmnx/actions/runs/15723859215/job/44309839807

PR #1338, opened 2025-08-14
Add pbf module to load graphs from OSM PBF files

gboeing:
This PR proposes a new module to read data from PBF files. It uses the `osmium` library to read PBF data and provides a familiar `graph_from_pbf` function for users. This is much faster than using the Overpass API and allows users to download data extracts locally for loading.

Initially I felt like we shouldn't provide any data filtering, and just load whatever data is in the user's desired PBF file. But we can maybe provide some simple built in filtering, with the expectation that the user who needs richer filtering should just pre-filter their PBF file as they wish.

Here's a simple usage example, using Geofabrik's [DC](https://download.geofabrik.de/north-america/us/district-of-columbia-latest.osm.pbf) extract:

```python
import osmnx as ox
filepath = "data/dc.osm.pbf"

# load PBF ways into graph with no filtering
G = ox.pbf.graph_from_pbf(filepath)
print(len(G))

# with simple filtering: only ways with highway=primary
tags = {"highway": ["primary"]}
G = ox.pbf.graph_from_pbf(filepath, tags)
print(len(G))
```

Or you can load the state of Oregon's highway network in ~8 seconds:

```python
filepath = "./data/oregon-latest.osm.pbf"
tags = {"highway": ["motorway", "motorway_link", "trunk", "trunk_link"]}
G = ox.pbf.graph_from_pbf(filepath, tags)
```

Or you can load all of Australia's highway network in ~45 seconds:

```python
filepath = "./data/australia-latest.osm.pbf"
tags = {"highway": ["motorway", "motorway_link", "trunk", "trunk_link"]}
G = ox.pbf.graph_from_pbf(filepath, tags)
```

More simple filtering examples:
```python
filepath = "data/dc.osm.pbf"
tags = ["highway"]
G = ox.pbf.graph_from_pbf(filepath, tags)
print(len(G))

tags = {"railway": ["subway"], "highway": ["primary"]}
G = ox.pbf.graph_from_pbf(filepath, tags)
print(len(G))

tags = {"railway": ["subway"], "highway": ["primary", "secondary"]}
G = ox.pbf.graph_from_pbf(filepath, tags)
print(len(G))

tags = {"railway": ["subway"]}
G = ox.pbf.graph_from_pbf(filepath, tags)
print(len(G))

tags = ["railway", "highway"]
G = ox.pbf.graph_from_pbf(filepath, tags)
print(len(G))

tags = {"highway": ["motorway", "motorway_link", "trunk", "trunk_link", "primary", "secondary"]}
G = ox.pbf.graph_from_pbf(filepath, tags)
print(len(G))

Gp = ox.projection.project_graph(G)
gdf_nodes, gdf_edges = ox.convert.graph_to_gdfs(Gp)
fig, ax = ox.plot.plot_graph(Gp, node_size=2)
```

I've tested projection, plotting, converting, etc and all seems to work fine with the graphs we've created from the PBF files.

You can also create a graph from Overpass with OSMnx, save to an OSM XML file, convert that to a PBF file, then load it back into OSMnx (as proof of concept only... this isn't a good workflow):

```python
from pathlib import Path
import osmium
fp_xml = Path("./data/graph.osm")
fp_pbf = Path("./data/graph.pbf")
ox.settings.all_oneway = True
G = ox.graph.graph_from_address("Piedmont, California, USA", dist=300, network_type="drive", simplify=False)
ox.io.save_graph_xml(G, fp_xml)
if fp_pbf.is_file():
    fp_pbf.unlink()
with osmium.SimpleWriter(fp_pbf) as writer:
    for obj in osmium.FileProcessor(fp_xml):
        writer.add(obj)
filepath = "./data/graph.pbf"
G = ox.pbf.graph_from_pbf(filepath)
len(G)
```

Any comments, feedback, or testing would be much appreciated.


shiqin-liu:
This is an impressive progress, thanks for working on this feature! 

I ran the feature locally, loading a sample Chicago metro osm .pbf, which takes about ~2.8 min. I also test for projection and routing, also works as expected! 

```Python
import osmnx as ox
import time
ox.__version__
'2.1.0.dev0'

start_time = time.time()
filepath='inputdata/osm_Chicago.pbf'
G = ox.pbf.graph_from_pbf(filepath, simplify=True)
print(f"Total processing time: {time.time() - start_time:.2f} seconds")

Total processing time: 170.85 seconds
```

Quick thoughts/questions on data filtering, the `global-indicators` software does rely on the OSMnx default walk type filter to retrieve the pedestrian network, e.g. https://github.com/healthysustainablecities/global-indicators/blob/1ad567bfe93c56a0c93f6512a1c215f11f773dc7/process/subprocesses/_03_create_network_resources.py#L112 . Many users might appreciate a default filter for convenience.  But with current setting, I am thinking of a scenario where we might want to get the OSMnx walk network from .pbf file, what would be the best way to handle it? does current tag parameter acts the same as a custom filter where we could pass a customized walk filter?

mvexel:
I stumbled upon this while I was looking into availability of the public Overpass server instances and reading #1377. 

Is loading from PBF still a path that you are pursuing? If yes, I may be able to contribute to the filtering logic. I have worked with Overpass in Python a fair amount and am also the maintainer of https://codeberg.org/mvexel/overpass-api-python-wrapper. 

As a first step, supporting the existing `osmnx` presets like `drive`, `walk`, `bike` could be useful?

gboeing:
@mvexel yes, this PR fell by the wayside a few months back, but it includes a simple working version of PBF loading functionality already.

The filtering logic presents an interesting problem... osmium makes it a bit tricky to do normal OSMnx-style `network_type` filters, but I'm open to clever ideas here.

However, I think the normal `custom_filter` function argument logic will be more or less impossible, because it would require parsing arbitrary Overpass filter code.

mvexel:
I was thinking in the direction of abstracting the filter definitions from the PBF / Overpass readers. There would be a `_filters.py` that would have the canonical declarations of the filter presets (highway k/v to exclude, etc.), and `_overpass.py` and `pbf.py` would transform those declarations into their respective syntax.

If this sounds like something you would consider, I can work on a minimal example to start.

gboeing:
Sounds like a potentially good approach. Sure I'd be happy to review a PR.

mvexel:
I had some free time to finish this up - sorry for the delay! https://github.com/gboeing/osmnx/pull/1381. I tried to keep it as small and contained as possible but the filter logic was a little bit more involved that I initially thought. Looking forward to discussing at #1381

b-a0:
This is great! I've installed it in my conda environment by adding the following to the environment.yml:

```yml
...
  - pip
  - pip:
    - osmnx @ git+https://github.com/gboeing/osmnx.git@pbf
    - osmium
```

I could then load the .pbf file [for the Netherlands from Geofabrik](https://download.geofabrik.de/europe/netherlands.html) (~1 GB) and extract and plot highways:

Loading took 88 seconds:
```py
tags = {"highway": ["motorway", "motorway_link", "trunk", "trunk_link", "primary", "secondary"]}
G = graph_from_pbf(filepath, tags)
```

Reprojecting and plotting took 8 seconds:

```py
Gp = ox.projection.project_graph(G)
gdf_nodes, gdf_edges = ox.convert.graph_to_gdfs(Gp)
fig, ax = ox.plot.plot_graph(Gp, node_size=2)
```
image

I hope this can be merged soon, such that it's easier to install.

If there is any specific functionality I can test, let me know, I'll do my best to help!

mszell:
Hi, just adding that I am also very much interested in this, as the overpass api is under heavy strain and fails too much, so this is really a good and needed approach. 
> If there is any specific functionality I can test, let me know, I'll do my best to help!

Same!

gboeing:
Thanks all for the feedback. Yes I'm happy to pick this back again in the coming days. The main issue is designing a good system for filtering. Overpass and PBF need very different filtering logic. Either they're used differently by users, or we need to do major work abstracting things.

gboeing:
Ok, the solution I have in mind is something like:
- Support the existing `network_type` presets by defining each preset once as structured tag predicates, then rendering those rules as Overpass QL or evaluating them directly against PBF way tags (aim to keep this simpler than all the abstraction proposed in #1381).
- Do not attempt to parse arbitrary Overpass `custom_filter` strings for PBF files. Instead, offer a Python `way_filter(tags) -> bool` callable for custom local filtering.
- Apply filters during the first PBF scan rather than creating a temporary filtered PBF; lazily import the optional `osmium` dependency and safely discard incomplete ways with a warning.

gboeing:
*Your feedback is requested!*

I've made some revisions to this PR. You can now either 1) use a built-in `network_type` or 2) pre-filter your PBF file with osmium to exactly what you want and then load that (akin to a `custom_filter` when querying Overpass).

The PR now:

- adds `ox.pbf.graph_from_pbf` for constructing graphs directly from local OSM PBF files
- supports all built-in OSMnx network types, like the other `graph_from_whatever` functions use
- uses native pyosmium filtering and a two-pass reader for (relatively) efficient way/node extraction
- supports `network_type=None` to load every way from an externally prefiltered PBF, including non-highway ways, so you can load any network data you want
- handles clipped PBFs by warning and removing ways whose referenced nodes are missing

Usage examples:

```python
import osmium
import osmnx as ox

# simply use a built-in network type
filepath = "data/dc.osm.pbf"
G = ox.pbf.graph_from_pbf(filepath, network_type="drive")
print(len(G))

# or, pre-filter with osmium to exactly what you want from a large extract
source = "data/oregon-latest.osm.pbf"
filtered = "data/oregon-major-roads.osm.pbf"
highways = ("motorway", "motorway_link", "trunk", "trunk_link")
tag_filter = osmium.filter.TagFilter(*(("highway", value) for value in highways))
with osmium.BackReferenceWriter(filtered, ref_src=source, overwrite=True) as writer:
    for way in osmium.FileProcessor(source, entities=osmium.osm.WAY).with_filter(tag_filter):
        writer.add_way(way)
G = ox.pbf.graph_from_pbf(filtered, network_type=None)
print(len(G))
```

carlhiggs:
Hi @gboeing - thanks for all your and others' work on this; it looks great.  I've been focused on a Fellowship application, so haven't completed my review, but tentatively, looks good (basically looks like we can replicate our existing GHSCI/`global-indicators` workflow using osmium for tag filtering from offline pbf).  I was distracted focusing on the Fellowship application primarily though, so I'm planning to check in more detail in coming days.  Just wanted to comment to let you know, I'm looking at this and will aim to get back soon.  Nice work!

mszell:
Very cool, thank you! I am happy to give it a spin next week.

gboeing:
FYI I also tested using pyrosm (out-of-core) as the back-end instead of osmium. There were trade-offs. pyrosm has a nicer API and could do the job in about ~150 lines of code instead of ~200 for osmium. For large region pbf files like Oregon or the Netherlands, pyrosm was 10-20% faster (runtime) than osmium. BUT, pyrosm also consumed 50-60% higher peak memory, has a 35x larger package installation footprint, and doesn't provide a linux ARM wheel. I ended up sticking with osmium under the assumption that there'll be users willing to spend ~30 extra seconds loading a graph to be able to get it to complete within their laptop's available RAM, since a key use case for pbf loading (instead of Overpass API) is very large regions. cc @HTenkanen if of interest.

mszell:
Oh nice, you went ahead, thx. Around when would you expect this feature to be part of a new conda osmnx release? No stress, just for simplicity I'll probably wait with my exploration + the implementation of pbf loading in my packages until then.

HTenkanen:
@gboeing : Thanks Geoff for the heads-up! 🙏🏻 I investigated the issue and now also Linux ARM is shipped. Package size of pyrosm was indeed also way larger than it needed to be on Linux as it included some unnecessary debug info along with the package. This is now fixed on the latest release (v0.13.1) and currently the wheel size is approx. 3.5 Mbs and the source is 5 Mbs (which is probably closer to the size of pyosmium): https://pypi.org/project/pyrosm/

gboeing:
@mszell I intend to include it in the next minor release. At this point I just need some feedback and test-driving from a few folks to make sure it works like people were expecting before I merge this PR into main. If anyone would like to volunteer -- please share your experience and thoughts here.

mszell:
Ah, I see, sorry I somehow mistook the main merging into pbf as pbf merging into main. Ok, I'll try to play around with it this week.

carlhiggs:
> _Your feedback is requested!_

Hi @gboeing --- I've run a test incorporating this in our GHSCI workflow; broadly it all checks out!  

I was using Claude Opus 5 to assist with this, and the one thing that was flagged as a possible upstream issue is that `graph_from_pbf` (like `graph_from_xml`, but unlike `graph_from_polygon`, which we currently use) doesn't explicitly set the `street_count` attribute.   This apparently has a cascading effect on how the `consolidate_intersections(dead_ends=False)` function which uses that attribute, and then impacts e.g. evaluation of intersection density (might end up 20% higher, unless we introduce new code to guard against this on our side --- which we can/should do, but might be worth handling in `osmnx`).  I just e-mailed the details; I didn't want to post here in case I was misunderstanding things -- and honestly, I don't have time to dig in depth myself right now (hence using the AI assistant to screen and run equivalence tests, so I'm not a blocker to your PR).   

Here are the suggested fixes (I like number 1; assuming this is an issue --- apologies I haven't first hand verified):

> 1. **Set it in `graph_from_pbf`** (mirrors `graph_from_polygon`):
>    ```python
>    spn = stats.count_streets_per_node(G)
>    nx.set_node_attributes(G, values=spn, name="street_count")
>    ```
>    Unlike the Overpass path there is no buffered graph available, so periphery counts will
>    under-report for ways clipped at the file boundary. That is inherent to a clipped extract and
>    still far better than the attribute being absent; it could be noted in the docstring.
> 
> 2. **Fail loudly instead of silently** (arguably the more robust fix, and it protects every entry
>    point including `graph_from_xml`): have `consolidate_intersections(dead_ends=False)` either
>    raise when `street_count` is missing, or fall back to computing counts on the graph it was
>    handed. Today an empty `spn` is indistinguishable from "no dead ends exist".
> 
> 3. Optionally apply (1) to `graph_from_xml` too, for consistency across the file-based
>    constructors.

mszell:
So, I implemented a draft for pbf imports at https://github.com/BikeNetKit/GrowBikeNet/tree/feature/import-from-pbf. Here my feedback:
- Importing the drive network works perfectly. I tested/explored manually (no unit tests yet) import from a city pbf import and from a country pbf import for Andorra la Vella, and the results seem exactly the same as for the graph_from_gdfs import, at least for my purposes. I didn't do any time benchmarking, but it would be interesting to know how slow the pbf import becomes if one uses a large pbf like germany or europe.
- All my existing dependencies are through conda-forge, so I would be very reluctant to add the osmium pip dependency. However, I then used the pyosmium library which exists on conda-forge, which you also mentioned, and which seems to provide the same functions as osmium (?) I assume this is the case, but I am unsure since you import osmium in your example. If I go ahead with implementing pbf imports, I would much prefer a new conda-forge dependency than a pip dependency.
- I didn't find the time to go down the "pre-filter with osmium" rabbit hole, but it would be important for me to apply a custom filter for importing non-trivially custom-filtered bike networks. In particular, I need to apply this filter: https://github.com/pyrosm/pyrosm/issues/341#issue-4742205953
I know that @HTenkanen recently implemented custom filters in pyrosm (I believe here: https://github.com/pyrosm/pyrosm/pull/357). I wonder if his code could be re-used here somehow, to have a function that translates a custom filter list into the proper osmium syntax.

In any case, this looks great, and thank you a lot for your efforts!

PR #1374, opened 2026-03-25
Tweaks to double-projection fix

gboeing:
Builds on #1373


gboeing:
@bridgesign in this PR I further streamlined #1373 and made it a bit more flexible with handling double re-projection. You can test it and see if it satisfies your needs.

bridgesign:
Thanks. No errors when projecting twice.

Issue #1008, opened 2023-06-23
Should the `geometries` module be renamed `features`?


gboeing:
Should the `geometries` module be renamed `features`, alongside the `geometries_from_X` functions --> `features_from_X`? As noted in #1007:

> Our documentation and docstrings have long had a disjointed way of referring to OSM map [features](https://wiki.openstreetmap.org/wiki/Map_features) as "geometries" or "entities" or "objects." This PR makes use of "features" terminology consistent.

This is true of our code base and API too. Over the years our code and API have referred to OSM map features as "geometries" or "entities" or "objects" or "footprints" or "pois" in different contexts and times.

I hate to make a major API breaking change at this point, but:
  1. this has been a point of confusion for users for a long time
  2. the proposed new `features` module and functions could be a drop-in replacement for an easy transition
  3. we could have a long deprecation time period with user warnings
  4. we could target the `geometries` module removal for a future v2.0 release

gboeing:
See WIP proposal in #1011.

PR #1376, opened 2026-04-08
Restore geometries_from_place as alias to features_from_place for bac…

html-Dashboard:
In OSMnx 2.x the function `geometries_from_place` was removed and replaced by `features_from_place`. Many tutorials, examples, and existing scripts still use the old name, which currently results in:

AttributeError: module 'osmnx' has no attribute 'geometries_from_place'

This change restores backward compatibility by adding a simple alias in `osmnx/__init__.py`:

from .features import features_from_place as geometries_from_place

Both calls

- ox.geometries_from_place("Weimar, Germany", tags={"building": True})
- ox.features_from_place("Weimar, Germany", tags={"building": True})

now return identical GeoDataFrames (`gdf_old.equals(gdf_new` is True).

This makes OSMnx 2.x compatible with older code and existing documentation again, with a minimal and safe change.

### DIFF
--- a/osmnx/__init__.py
+++ b/osmnx/__init__.py
@@ -35,3 +35,7 @@ from . import utils_geo as utils_geo
 
 # expose the old v1 API for backwards compatibility
 from ._api_v1 import *  # noqa: F403
+
+# backward compatibility alias for removed v1 function
+from .features import features_from_place as geometries_from_place
+


gboeing:
Following semantic versioning standards, major releases can include API-breaking changes. v2 was released a year and a half ago, and prior to that we provided about a year+ of runway of deprecation warnings for API changes, including this one: specifically, the `features` module replaced the `geometries` module about 3 years ago, with deprecation warnings (#1008 #1011). For the sake of a streamlined API and consistency, I don't think we should clutter and alias obsolete code.

> the function geometries_from_place was removed and replaced by features_from_place

It's not specific to one function. The entire `geometries` module was replaced by the `features` module.

PR #1402, opened 2026-07-27
Fix simplify_graph treating equally-missing edge_attrs_differ values as differing

harisbal:
\`_is_endpoint\`'s Rule 5 collects each incident edge's value for every \`edge_attrs_differ\` attribute into a set, and treats a node as a forced endpoint if that set has more than one element. When the attribute is missing (\`NaN\`) on multiple incident edges, this backfires: per IEEE 754 \`NaN != NaN\`, and each missing value is typically its own distinct \`float\` object rather than a shared singleton, so \`{nan, nan}\` doesn't collapse to one element the way \`{None, None}\` does. The node gets wrongly treated as an endpoint and simplification stops there even though every incident edge is equally missing the attribute.

Reproduced with a synthetic 3-node path graph where the middle node's \`lanes\` attribute is \`NaN\` on all four incident edges — \`simplify_graph(G, edge_attrs_differ=["lanes"])\` keeps the middle node instead of merging through it (verified the added test fails on main and passes with this fix).

Fix: canonicalize \`NaN\` to \`None\` before building the comparison sets, since \`None\` is a true singleton and compares correctly by identity.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - de-duplicate features when creating a features gdf (#1400)
 - validate tolerance when consolidating intersections (#1401)
+- fix simplify_graph treating equally-missing edge_attrs_differ values as differing (#TBD)
 
 ## 2.1.1 (2026-07-21)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - de-duplicate features when creating a features gdf (#1400)
 - validate tolerance when consolidating intersections (#1401)
-- fix simplify_graph treating equally-missing edge_attrs_differ values as differing (#TBD)
+- fix simplify_graph treating equally-missing edge_attrs_differ values as differing (#1402)
 
 ## 2.1.1 (2026-07-21)
 
--- a/osmnx/simplification.py
+++ b/osmnx/simplification.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import itertools
 import logging as lg
+import math
 from typing import TYPE_CHECKING
 from typing import Any
 
@@ -44,7 +45,7 @@ def _canonicalize_missing(value: Any) -> Any:  # noqa: ANN401
     value
         `None` if `value` is a NaN float, otherwise `value` unchanged.
     """
-    if isinstance(value, float) and pd.isna(value):
+    if isinstance(value, float) and math.isnan(value):
         return None
     return value
 
--- a/osmnx/simplification.py
+++ b/osmnx/simplification.py
@@ -26,24 +26,17 @@ if TYPE_CHECKING:
 
 def _canonicalize_missing(value: Any) -> Any:  # noqa: ANN401
     """
-    Normalize a missing (NaN) edge attribute value to `None`.
-
-    Per IEEE 754, NaN never equals itself, and each missing value is usually
-    its own distinct float object rather than a shared singleton. So a set of
-    otherwise-equally-missing values, such as `{nan, nan}`, does not
-    deduplicate to a single element the way `{None, None}` does. Canonicalizing
-    NaN to `None` before set-based comparisons avoids miscounting such values
-    as differing.
+    Convert a floating-point NaN to `None`.
 
     Parameters
     ----------
     value
-        The value to canonicalize.
+        The value to check.
 
     Returns
     -------
     value
-        `None` if `value` is a NaN float, otherwise `value` unchanged.
+        `None` for NaN; otherwise the original value.
     """
     if isinstance(value, float) and math.isnan(value):
         return None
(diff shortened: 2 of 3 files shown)


gboeing:
What specific python NaN data type are we talking about here?

harisbal:
The issue I faced involved float('nan'). However, retrieving np.nan values  from a DataFrame would also raise the same issue. Only pd.NA behaves like a singleton

**`np.nan`**
```python
>>> df = pd.DataFrame([np.nan, np.nan])
>>> a = df.loc[0, 0]
>>> b = df.loc[1, 0]
>>> a is b
False
>>> set([a, b])
{nan, nan}
```

**`float('nan')`** — same result, since pandas stores it as `float64` too
```python
>>> df = pd.DataFrame([float('nan'), float('nan')])
>>> a = df.loc[0, 0]
>>> b = df.loc[1, 0]
>>> a is b
False
>>> set([a, b])
{nan, nan}
```

**`pd.NA`** — collapses to one element, since `pd.NA` is a singleton
```python
>>> df = pd.DataFrame([pd.NA, pd.NA])
>>> a = df.loc[0, 0]
>>> b = df.loc[1, 0]
>>> a is b
True
>>> set([a, b])
{<NA>}
```

gboeing:
> The issue I faced involved float('nan').

Interesting. Could you share a MWE of how this happened in practice? Like what was the use case that produced non-singleton NaNs? One possible fix is what you proposed here. Another could be better NaN handling consistently across the package.

harisbal:
I dug a bit deeper and I found that osmnx deals with this case when creating a graph from gdfs:
https://github.com/gboeing/osmnx/blob/51619827e1f233d9139aed847b0c844c63c22a2d/osmnx/convert.py#L323

My case involved the use of pyrosm and the [to_graph](https://pyrosm.readthedocs.io/en/stable/api/pyrosm.OSM.to_graph.html) method which does not drop `nan` values in the edge attributes:
`to_graph(..., graph_type="networkx", osmnx_compatible=True)` 

However, any graph which has been touched externally before osmnx simplification can express this behavior:

```python
import networkx as nx
import osmnx as ox

# A 3-node path (1)-(2)-(3), both directions. Node 2 is a plain pass-through
# vertex (2 neighbors, degree 2) and should collapse away.
g = nx.MultiDiGraph(crs="epsg:4326")
coords = {1: (0.0, 0.0), 2: (1.0, 0.0), 3: (2.0, 0.0)}
for n, (x, y) in coords.items():
    g.add_node(n, x=x, y=y)

for u, v in [(1, 2), (2, 1), (2, 3), (3, 2)]:
    g.add_edge(u, v, osmid=1, highway="residential", length=1.0, lanes=float("nan"))

node = 2
a = g[1][node][0]["lanes"]
b = g[node][3][0]["lanes"]
print(f"a={a!r} b={b!r} a is b: {a is b} a == b: {a == b}")

simplified = ox.simplify_graph(g, edge_attrs_differ=["lanes"])
print("node wrongly kept as endpoint:", node in simplified.nodes)​
```

Output:
​```
a=nan b=nan a is b: False a == b: False
node wrongly kept as endpoint: True
​```

Both edges are equally missing `lanes`, but each carries its own distinct `float("nan")` object. `_is_endpoint`'s rule 5 unions the incident edges' values into a set to check whether they differ:

gboeing:
> any graph which has been touched externally before osmnx simplification can express this behavior

Part of the motivation for creating the `_validate` module was to manually check/enforce OSMnx expectations on demand. That is, if a user externally manipulates a graph, they can verify that it meets what OSMnx expects. Perhaps this should be a new validation check, rather than ad hoc NaN/None handling code in various parts of the package where nulls appear. That way, if a user externally adds non-singleton NaNs to their graph attributes, validation can flag it for them.

I'm more inclined to think that OSMnx should handle nulls correctly internally, but only provide a validation check for what users do externally. There are too many edge cases to over-engineer around if we try to catch everything that could happen outside the package. But I'm not positive that's the right path, and I'm open to debate.

gboeing:
I thought about this more, and I think I like your solution. I made a couple edits and pushed them. You can take a look and see if it's ready to go at this point.

harisbal:
The changes you suggested are reasonable, no objection by me.

On a more general note, I do agree that a more centralized way to handle NaNs would be more suitable, but I am unsure how that would look in practice, since OSMnx functions don't depend on a custom type (e.g. an ox.Graph) which could be used to enforce compatibility between nx and OSMnx. Validating the input graph before each function would be a solution, but it feels too troublesome/pervasive for what's otherwise a fairly lightweight, functional API.

Since arbitrary nx graphs are accepted as input, it feels natural to accept their behavior too and allow for arbitrary values in their attributes, rather than imposing a stricter contract after the fact. Maybe the _validate module you mentioned could be the right place for this though — an opt-in check that flags non-singleton NaNs, rather than OSMnx needing a custom graph type or validating on every call.

Separately, what I realized from this "bug" was that OSMnx drops attributes with value NaN. Initially that felt reasonable, but I realized a user may lose attributes and info silently without ever noticing. I think this deserves different treatment than the endpoint-detection issue — for example, setting the attribute to None (which is a singleton) instead of dropping it, so the fact that the attribute existed but was missing is preserved rather than erased.

gboeing:
I agree that a custom graph type or validation before every function would be excessive.

OSMnx uses sparse NetworkX attribute dictionaries: missing values are generally represented by absent keys. Accepting arbitrary NetworkX graphs does not imply lossless preservation of every attribute representation. For example, `simplify_graph` already transforms attributes by aggregating values, deduplicating them, and ignoring absent keys.

Changing that in OSMnx would affect user-facing behavior across many parts of the package including conversion, simplification, and file saving. That would require a broader schema-preservation policy, which isn't currently planned, but could be a topic of broad conversation someday if users end up needing this enough to change public behavior.

harisbal:
Osmnx is a great library as is, I don;t feel there is a need for major API changes at the moment. A custom graph sounds tempting but I am not sure it will be actually beneficial. Thank you for the discussion :)

PR #1406, opened 2026-09-12
Stop add_edge_bearings raising when every edge is a self-loop

DenisDrobyshev:
## Summary

`add_edge_bearings` ignores self-loop edges, as its docstring says, then builds the coordinate array from whatever is left and indexes it in two dimensions:

```python
uvk = [(u, v, k) for u, v, k in G.edges if u != v]
...
coords = np.array([(y[u], x[u], y[v], x[v]) for u, v, k in uvk])
bearings = calculate_bearing(coords[:, 0], coords[:, 1], coords[:, 2], coords[:, 3])
```

When nothing is left, `np.array([])` has shape `(0,)` and the two-dimensional index raises:

```python
import networkx as nx
import osmnx as ox

G = nx.MultiDiGraph(crs="epsg:4326")
G.add_node(0, x=0.0, y=0.0)
G.add_edge(0, 0, length=1.0)

ox.add_edge_bearings(G)
# IndexError: too many indices for array: array is 1-dimensional, but 2 were indexed
```

A graph with no edges at all fails the same way, and `orientation_entropy` inherits it because it needs the bearings first.

A graph of nothing but self-loops is not contrived: a closed way — a roundabout, a turning circle, a loop of track — becomes a self-loop once its interstitial nodes are simplified away, and truncating to a small bounding box can leave one of those on its own.

## Changes

* `osmnx/bearing.py` — return the graph unchanged when the self-loop filter leaves no edges, which is what "ignores self-loop edges" amounts to when they are all there is.
* `tests/test_osmnx.py` — `test_bearings_without_any_non_self_loop_edge`, covering both an edgeless graph and a self-loop-only one. It builds its graphs by hand, so it needs no network.
* `CHANGELOG.md` — entry under TBD.

## Verification

`pytest tests/test_osmnx.py -k without_any_non_self_loop`: **1 passed**. With the fix reverted it fails with the `IndexError` above, raised at `osmnx/bearing.py:124`. The test detects the defect rather than passing regardless.

The rest of the suite queries Overpass and Nominatim, so I did not run it locally.

`ruff check` and `ruff format --check` are clean on both files, apart from `CPY001`, which every file in the repository reports.


gboeing:
Can you provide a minimal example from the real world to reproduce the problem?

